PR #7, opened 2020-06-05
Historiser les CGUs d'un fournisseur

Ndpnt:
PR relative à l'issue #3

### DIFF
--- a/src/history/git.js
+++ b/src/history/git.js
@@ -8,6 +8,9 @@ const DEFAULT_GIT_OPTIONS = {
   dryRun: process.env.NODE_ENV === 'test'
 };
 
+const AUTHOR_NAME = 'AmbaNum Bot';
+const AUTHOR_EMAIL = 'ambanum.bot@disinfo.quaidorsay.fr';
+
 export async function add({ filepath }) {
   return await git.add({ ...DEFAULT_GIT_OPTIONS, filepath });
 }
@@ -17,8 +20,8 @@ export async function status({ filepath }) {
 }
 
 export async function commit(options) {
-  let name = 'AmbaNum Bot';
-  let email = 'ambanum.bot@disinfo.quaidorsay.fr';
+  let name = AUTHOR_NAME;
+  let email = AUTHOR_EMAIL;
 
   if (options.author) {
     name = options.author.name || name;
--- a/src/history/git.js
+++ b/src/history/git.js
@@ -2,10 +2,12 @@ import fs from 'fs';
 import path from 'path';
 import git from 'isomorphic-git';
 
+const __dirname = path.dirname(new URL(import.meta.url).pathname);
+
 const DEFAULT_GIT_OPTIONS = {
   fs,
-  dir: path.resolve(),
-  dryRun: process.env.NODE_ENV === 'test'
+  dir: path.resolve(__dirname, '../..'),
+  dryRun: process.env.NODE_ENV === 'test',
 };
 
 export async function add({ filepath }) {
--- a/src/history/git.js
+++ b/src/history/git.js
@@ -1,0 +1,38 @@
+import fs from 'fs';
+import path from 'path';
+import git from 'isomorphic-git';
+
+const DEFAULT_GIT_OPTIONS = {
+  fs,
+  dir: path.resolve(),
+  dryRun: process.env.NODE_ENV === 'test'
+};
+
+export async function add({ filepath }) {
+  return await git.add({ ...DEFAULT_GIT_OPTIONS, filepath });
+}
+
+export async function status({ filepath }) {
+  return await git.status({ ...DEFAULT_GIT_OPTIONS, filepath });
+}
+
+export async function commit(options) {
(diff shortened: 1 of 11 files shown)


src/history/index.js
```diff
@@ -0,0 +1,42 @@
...
+  const dir = `${RAW_DIRECTORY}/${serviceProviderId}`;
+
+  if (!fsApi.existsSync(dir)){
+    fsApi.mkdirSync(dir);
+  }
+
+  return fs.writeFile(`${dir}/${policyType}.html`, fileContent);
+}
+
+export async function commitRaw(serviceProviderId, policyType) {
```
MattiSG, line 25: We're exposing a more abstract function, let's make sure the name reflects the intent rather than the action 🙂 

```suggestion
export async function updateRaw(serviceProviderId, policyType) {
```

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+const SERVICE_PROVIDER_ID = 'test_service_provider';
+const POLICY_TYPE = 'terms_of_service';
+const FILE_CONTENT = 'ToS fixture data with UTF-8 çhãràčtęrs';
+const EXPECTED_FILE_PATH = `${RAW_DIRECTORY}/${SERVICE_PROVIDER_ID}/${POLICY_TYPE}.html`;
+
+
+describe('History', () => {
+  describe('#storeRaw', () => {
+    context("when service provider's directory already exist", () => {
```
MattiSG, line 16: Single quotes?

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+      after(() => {
+        fs.unlinkSync(EXPECTED_FILE_PATH);
```
MattiSG, line 18: > I believe Mocha supports async `before` and `after` hooks as long as we return the promise, so we could use `unlink` rather than `unlinkSync` and accelerate the tests 🙂

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+      });
+
+      it('creates a file for the given service provider', async () => {
+        await storeRaw(SERVICE_PROVIDER_ID, POLICY_TYPE, FILE_CONTENT);
+
+        expect(fs.readFileSync(EXPECTED_FILE_PATH, { encoding: 'utf8' })).to.equal(FILE_CONTENT);
+      });
+    });
+
+    context("when service provider's directory not already exist", () => {
```
MattiSG, line 28:
```suggestion
    context("when service provider's directory does not already exist", () => {
```

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+      const NOT_EXISTING_SERVICE_PROVIDER_ID = 'test_not_existing_service_provider';
```
MattiSG, line 29: > When I read `NOT_EXISTING_SERVICE_PROVIDER_ID` I expect the test to check for a failure condition. What about `NEW_SERVICE_PROVIDER_ID`? 🙂

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+        await storeRaw(NOT_EXISTING_SERVICE_PROVIDER_ID, POLICY_TYPE, FILE_CONTENT);
+
+        expect(fs.readFileSync(NOT_EXISTING_SERVICE_PROVIDER_EXPECTED_FILE_PATH, { encoding: 'utf8' })).to.equal(FILE_CONTENT);
+      });
+    });
+  });
+
+  describe('#commitRaw', () => {
+    before(async () => {
+      await storeRaw(SERVICE_PROVIDER_ID, POLICY_TYPE, FILE_CONTENT);
```
MattiSG, line 46: If we explicitly declare the hook as `async`, why not simply return the result of calling `storeRaw`?

src/history/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+  describe('#persistRaw', () => {
+    before(async () => {
+      await persistRaw(SERVICE_PROVIDER_ID, POLICY_TYPE, FILE_CONTENT);
+    });
+
+    after(() => {
+      fs.unlinkSync(EXPECTED_FILE_PATH);
+    });
+
+    it('creates a file for the given service provider', async () => {
```
MattiSG, line 68: Why is this `async`? 🤔

src/history/index.js
MattiSG: After discussion: this `export` is only for testing, it is not expected to be exposed. Maybe add a comment? 🙂

src/history/git.js
```diff
@@ -0,0 +1,41 @@
...
+import path from 'path';
+import git from 'isomorphic-git';
+
+const DEFAULT_GIT_OPTIONS = {
+  fs,
+  dir: path.resolve(),
+  dryRun: process.env.NODE_ENV === 'test'
+};
+
+const AUTHOR_NAME = 'AmbaNum Bot';
```
MattiSG, line 11:
```suggestion
const AUTHOR_NAME = 'AmbNum Bot';
```

PR #8, opened 2020-06-05
Mettre à jour les CGUs automatiquement

Ndpnt:
#5


schedule.js
```diff
@@ -0,0 +1,21 @@
+import schedule from 'node-schedule';
+import childProcess from 'child_process';
+
+const rule = new schedule.RecurrenceRule();
+rule.minute = 30; // every hour at 30 minutes after the hour
```
MattiSG, line 5: > Making the time at which the scraping is expected to occur visible in the logs could be useful to users 🙂

MattiSG, line 5: Isn't there a way to say “every hour starting from now”?

schedule.js
```diff
@@ -0,0 +1,21 @@
...
+
+console.log('The scheduler is running…');
+
+schedule.scheduleJob(rule, function(){
+  console.log('Start scheduled job:');
+
+  const forkedProcess = childProcess.fork('./index.js');
```
MattiSG, line 12: Why don't we just call the main function instead of spawning a new V8 instance with all the memory & CPU costs that entails? 🌱🙂

schedule.js
Ndpnt: Simply with `setInterval`, but I'm not sure that what we want. I think is better to have predictable runs.

PR #9, opened 2020-06-05
Nettoyer les CGUs

### LINKED ISSUE #6: Nettoyer les CGUs
Transformer le HTML brut en Markdown.

Ndpnt:
fixes #6


data/raw/README.md
```diff
@@ -0,0 +1,2 @@
+Ensures the current directory is tracked with git.
+Will contain raw service providers terms.
```
MattiSG, line 2:
```suggestion
Raw HTML pages of service providers terms are stored in this folder.
They are kept to ensure history is kept even if the sanitizing parsers that transform contents into more readable text fail or evolve.
```

data/sanitized/README.md
```diff
@@ -0,0 +1,2 @@
+Ensures the current directory is tracked with git.
+Will contain sanitized service providers terms.
```
MattiSG, line 2:
```suggestion
Tracked terms of services are stored in a pure text format in this folder.
This enables better diffs and readability.
```

src/sanitizer/index.js
```diff
@@ -0,0 +1,15 @@
+import TurndownService from 'turndown';
+import cheerio from 'cheerio';
```
MattiSG, line 2: >  “[Cheerio](https://www.npmjs.com/package/cheerio) removes all the DOM inconsistencies and browser cruft from the jQuery library, revealing its **truly gorgeous API**.”

😰 

It also says:

> Cheerio wraps around parse5 parser.

Considering the current use we have of `cheerio` is apparently only parsing (as opposed to DOM manipulation), could we consider using directly [`parse5`](https://github.com/inikulin/parse5) instead? 🙏

src/sanitizer/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default async function sanitize(content, selector) {
+  const turndownService = new TurndownService();
```
MattiSG, line 5: > Do we really need to instantiate a new service on each run? 🌱

src/sanitizer/index.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+const expect = chai.expect;
+
+describe('Sanitizer', () => {
+  let facebookTermsHTML;
+  let facebookTermsMarkdown;
+  
+  describe('#sanitize', () => {
+    before(() => {
+      facebookTermsHTML = fs.readFileSync(`${process.cwd()}/fixtures/facebook_terms_raw.html`, { encoding: 'utf8' });
```
MattiSG, line 14: It is fragile to rely on `cwd` rather than the module's path.

src/history/index.js
```diff
@@ -6,25 +6,46 @@ import * as git from './git.js';
...
+if (!fsApi.existsSync(directory)){
+    fsApi.mkdirSync(directory);
   }
 
-  return fs.writeFile(`${dir}/${policyType}.html`, fileContent);
+  return fs.writeFile(`${directory}/${policyType}.${isSanitized ? 'md' : 'html'}`, fileContent);
 }
 
-export async function commitRaw(serviceProviderId, policyType) {
+export async function commit({ serviceProviderId, policyType, isSanitized }) {
```
MattiSG, line 44: `isSanitized` has a default value of `undefined` which is falsy. I would have preferred to make that distinction super-clear through the function name (`storeRaw` / `storeSanitized`). If we want to go the way of passing a single options object, then I believe we should validate in some way that option.

Unless of course this is exported only for testing purposes…? Let's have a chat around this 🙂

src/history/index.js
```diff
@@ -6,25 +6,46 @@ import * as git from './git.js';
...
+    serviceProviderId, 
+    policyType, 
+    fileContent,
+    isSanitized: false
+  });
 }
 
-export async function storeRaw(serviceProviderId, policyType, fileContent) {
-  const dir = `${RAW_DIRECTORY}/${serviceProviderId}`;
+export async function persistSanitized(serviceProviderId, policyType, fileContent) {
```
MattiSG, line 20: Isn't this inconsistent with the new `persist({…})` API exposed in this changeset? 🤔

PR #22, opened 2020-06-16
Add related `raw` changes commit ID in `sanitized` commit messages

Ndpnt:


src/history/index.js
```diff
@@ -10,11 +10,12 @@ export async function persistRaw(serviceProviderId, policyType, fileContent) {
   });
 }
 
-export async function persistSanitized(serviceProviderId, policyType, fileContent) {
+export async function persistSanitized(serviceProviderId, policyType, fileContent, relatedRawCommit) {
```
MattiSG, line 13:
```suggestion
export async function persistSanitized(serviceProviderId, policyType, fileContent, relatedRawCommitSha) {
```

src/history/persistor.js
```diff
@@ -18,9 +18,16 @@ const ROOT_DIRECTORY = path.resolve(__dirname, `../../data`);
 export const RAW_DIRECTORY = `${ROOT_DIRECTORY}/raw`;
 export const SANITIZED_DIRECTORY = `${ROOT_DIRECTORY}/sanitized`;
 
-export async function persist({ serviceProviderId, policyType, fileContent, isSanitized }) {
+export async function persist({ serviceProviderId, policyType, fileContent, relatedRawCommit, isSanitized }) {
   const filePath = await save({ serviceProviderId, policyType, fileContent, isSanitized });
-  const message = `Update ${isSanitized ? 'sanitized' : 'raw'} ${serviceProviderId} ${DOCUMENTS_TYPES[policyType].name} document`;
+  let message = `Update ${isSanitized ? 'sanitized' : 'raw'} ${serviceProviderId} ${DOCUMENTS_TYPES[policyType].name} document`;
+
+  if (relatedRawCommit) {
```
MattiSG, line 25: Could we maybe replace `isSanitized` by `relatedRawCommitSha`? If we have a raw commit SHA then we're in a sanitized commit anyway…

Otherwise, maybe:
```suggestion
  if (isSanitized) {
```

src/history/persistor.js
```diff
@@ -18,9 +18,16 @@ const ROOT_DIRECTORY = path.resolve(__dirname, `../../data`);
...
+    message += `
+
+Find related raw file changes in this commit: https://github.com/ambanum/CGUs-data/tree/${relatedRawCommit}`;
```
MattiSG, line 28: I'd prefer to avoid hardcoding the storage backend URL in the entire history. If we're exploring on GitHub, GitHub will auto link in any case 🙂 

```suggestion
Find related raw file changes in commit ${relatedRawCommit};
```

src/history/persistor.js
Ndpnt: I'm not in favor of removing `isSanitized` because even if in theory we should always have a `relatedRawCommitSha` when we try to persist the sanitized version of the file, if for a reason or another the raw commit failed and we don't get the sha, we will call save function with parameters corresponding to the `raw` version and we could end up in inconsistent history with sanitized files in raw directory…

Ndpnt: And for the message I also would keep `if (relatedRawCommitSha) …`, to ensure adding message only when we effectively have a sha.

PR #24, opened 2020-06-18
Adding 39 new providers in JSON format

AdrienFines:


MattiSG:
Tests fail because the SSH key is not added for safety reasons on builds from forks.

@Ndpnt any way we could prevent that step from being tried in forks? 🙂

Ndpnt:
> Tests fail because the SSH key is not added for safety reasons on builds from forks.
> 
> @Ndpnt any way we could prevent that step from being tried in forks? 🙂

Note really, as it is necessary for cloning submodule.
But if we remove the submodule as we discussed, it will solve this issue.

ChessPolicy:
Hello,
I just saw your comments, thanks for your insights :)
I will try to follow and complete the README before next Tuesday (and adjusting the initial commit if necesary), I come back to you if I have any questions/problems !

Ndpnt:
Thanks for adding TikTok 👍. 
Compared to the list that LV shared with us, it still miss some of the most popular online content-sharing services.
Can you look at this list and add the major ones?

AdrienFines:
I will add them as soon as possible.
I've also been trying to test providers following the Readme, but encountered some problems, I will open an issue soon if it is ok with you ! :)

AdrienFines:
Hello, 
After a small talk with Matti we concluded that it would be better if I add clean providers (i.e. with correct sanitized datas, without artefacts) progressively, which is why I close this pull request.

PR #25, opened 2020-06-18
Ensure being on branch master before pushing

Ndpnt:


src/history/git.js
```diff
@@ -29,6 +29,7 @@ export async function commit(filepath, message) {
 }
 
 export async function pushChanges() {
+  await git.checkout('master');
   return git.push('origin', 'master');
```
MattiSG, line 33: Assuming that all git parameters are passed directly by this low-level library, this should ensure the `master` branch is pushed without needing the added line, and thus without having any side-effect on the target repository.

```suggestion
  return git.push('origin', 'master:master');
```

Ndpnt:
With #27 it's no longer relevant.

PR #26, opened 2020-06-18
Add notifications

Ndpnt:
Notify on scraping errors
Notify on sanitized document update

_There is no tests for now but I still open this PR to allow to review during my days off. 
As testing interaction with SendInBlue seems to be a heavy work, it would be useful for me to discuss what deserve to be tested._

PR #29, opened 2020-06-30
Refactor

Ndpnt:
Extract environment dependent variables in config files.


Ndpnt:
No longer relevant as #26 is closed.

PR #30, opened 2020-06-30
Notify on errors and updates

Ndpnt:
Notify on scraping errors
Notify on sanitized document update

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -1,0 +1,9 @@
+{
+  "history": {
+    "author": {
+      "name": "AmbNum Bot Dev",
+      "email": "ambnum.bot.dev@disinfo.quaidorsay.fr"
+    }
+  }
+}
+
--- a/config/development.json
+++ b/config/development.json
@@ -1,9 +1,0 @@
-{
-  "history": {
-    "author": {
-      "name": "AmbNum Bot Dev",
-      "email": "ambnum.bot.dev@disinfo.quaidorsay.fr"
-    }
-  }
-}
-
--- a/index.js
+++ b/index.js
@@ -3,8 +3,6 @@ import CGUs from './src/index.js';
 (async () => {
   try {
     const app = new CGUs();
-    await app.init();
-
     await app.updateTerms();
   } catch (e) {
     console.log(e);
--- a/index.js
+++ b/index.js
@@ -1,8 +1,18 @@
-import updateTerms from './src/index.js';
+import * as notifier from './src/notifier/index.js';
+import CGUs from './src/index.js';
 
 (async () => {
   try {
-    await updateTerms();
+    const app = new CGUs();
+    await app.init();
+
+    notifier.init(app.serviceProviders, app.documentsTypes);
+
+    app.on('sanitizedDocumentChange', notifier.onSanitizedDocumentChange);
+    app.on('documentScrapingError', notifier.onDocumentScrapingError);
+    app.on('applicationError', notifier.onApplicationError);
+
+    await app.updateTerms();
   } catch (e) {
     console.log(e);
   }
--- a/index.js
+++ b/index.js
@@ -1,16 +1,9 @@
-import * as notifier from './src/notifier/index.js';
 import CGUs from './src/index.js';
 
 (async () => {
   try {
     const app = new CGUs();
     await app.init();
-
-    notifier.init(app.serviceProviders, app.documentsTypes);
-
-    app.on('sanitizedDocumentChange', notifier.onSanitizedDocumentChange);
-    app.on('documentScrapingError', notifier.onDocumentScrapingError);
-    app.on('applicationError', notifier.onApplicationError);
 
     await app.updateTerms();
   } catch (e) {
--- a/src/history/git.js
+++ b/src/history/git.js
@@ -1,7 +1,7 @@
 import path from 'path';
 
-import dotenv from 'dotenv';
 import config from 'config';
+import dotenv from 'dotenv';
 import simpleGit from 'simple-git';
 
 dotenv.config();
--- a/src/history/git.js
+++ b/src/history/git.js
@@ -1,9 +1,10 @@
 import path from 'path';
 
 import dotenv from 'dotenv';
-dotenv.config();
+import config from 'config';
 import simpleGit from 'simple-git';
 
+dotenv.config();
 const __dirname = path.dirname(new URL(import.meta.url).pathname);
 
 const DATA_PATH = (process.env.CI ? '' : '../') + '../../cgus-data';
@@ -11,8 +12,8 @@ const DATA_PATH = (process.env.CI ? '' : '../') + '../../cgus-data';
 export const git = simpleGit(path.resolve(__dirname, DATA_PATH));
 
 if (process.env.CI || process.env.NODE_ENV === 'production') {
-  git.addConfig('user.name', process.env.AUTHOR_NAME)
-     .addConfig('user.email', process.env.AUTHOR_EMAIL);
+  git.addConfig('user.name', config.get('history.author.name'))
+     .addConfig('user.email', config.get('history.author.email'));
 }
 
 export async function add(filepath) {
@@ -24,7 +25,7 @@ export async function status() {
 }
 
 export async function commit(filepath, message) {
-  const summary = await git.commit(message, relativePath(filepath), { '--author': `${process.env.AUTHOR_NAME} <${process.env.AUTHOR_EMAIL}>` });
+  const summary = await git.commit(message, relativePath(filepath), { '--author': `${config.get('history.author.name')} <${config.get('history.author.email')}>` });
   return summary.commit.replace('HEAD ', '');
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -62,7 +62,7 @@ export default class CGUs extends events.EventEmitter {
       console.log('Pushed changes to the repository');
       console.log('______________________________');
     }
-  };
+  }
 
   async updateServiceProviderDocument({ serviceProviderId, serviceProviderName, document }) {
     const { documentType, url, contentSelector, sanitizationPipeline } = document;
--- a/src/index.js
+++ b/src/index.js
@@ -14,9 +14,15 @@ import { DOCUMENTS_TYPES } from './documents_types.js';
 export default class CGUs extends events.EventEmitter {
   constructor() {
     super();
-    this.initialized;
-    this.serviceProvidersManifests;
     this.documentTypes = DOCUMENTS_TYPES;
+  }
+
+  get serviceProviders() {
+    return this.serviceProvidersManifests;
+  }
+
+  get documentsTypes() {
+    return this.documentTypes;
   }
 
   async init() {
@@ -96,12 +102,5 @@ export default class CGUs extends events.EventEmitter {
       console.error(`${logPrefix} Error:`, error);
       this.emit('applicationError', serviceProviderId, documentType, error);
     }
-  }
-
-  get serviceProviders() {
-    return this.serviceProvidersManifests;
-  }
-  get documentsTypes() {
-    return this.documentTypes;
   }
 }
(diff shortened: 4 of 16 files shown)


config/development.json
```diff
@@ -0,0 +1,9 @@
+{
+  "history": {
+    "author": {
+      "name": "AmbNum Bot Dev",
+      "email": "ambnum.bot.dev@disinfo.quaidorsay.fr"
```
MattiSG, line 5: > I'm not sure I see the point of having a different email address in dev 🙂

index.js
```diff
@@ -1,8 +1,18 @@
...
+import * as notifier from './src/notifier/index.js';
+import CGUs from './src/index.js';
 
 (async () => {
   try {
-    await updateTerms();
+    const app = new CGUs();
+    await app.init();
+
+    notifier.init(app.serviceProviders, app.documentsTypes);
```
MattiSG, line 9: > It might be simpler, for API consistency, to always declare `init` functions as `async`.

index.js
```diff
@@ -1,8 +1,18 @@
...
+
+    app.on('sanitizedDocumentChange', notifier.onSanitizedDocumentChange);
+    app.on('documentScrapingError', notifier.onDocumentScrapingError);
+    app.on('applicationError', notifier.onApplicationError);
```
MattiSG, line 13: Do we really want to do that in the `index.js` (as opposed to `scheduler`)? If I execute locally, I probably don't want to send notifications.

src/history/git.js
```diff
@@ -1,18 +1,19 @@
 import path from 'path';
 
 import dotenv from 'dotenv';
-dotenv.config();
+import config from 'config';
 import simpleGit from 'simple-git';
```
MattiSG, line 5: > Alphabetical order for dependencies groups?

src/index.js
```diff
@@ -9,60 +11,97 @@ import { persistRaw, persistSanitized, pushChanges } from './history/index.js';
...
-  if (rawSha) {
-    console.log(`${logPrefix} Save raw file to '${rawFilePath}'.`);
-    console.log(`${logPrefix} Commit raw file in '${rawSha}'.`);
-  } else {
-    console.log(`${logPrefix} No raw changes, didn't commit.`);
+export default class CGUs extends events.EventEmitter {
+  constructor() {
+    super();
+    this.initialized;
+    this.serviceProvidersManifests;
```
MattiSG, line 18: I don't understand the effect of these statements.

src/index.js
```diff
@@ -9,60 +11,97 @@ import { persistRaw, persistSanitized, pushChanges } from './history/index.js';
...
+    this.documentTypes = DOCUMENTS_TYPES;
   }
 
-  const sanitizers = getSanitizers(serviceProviderId);
-  const sanitizedContent = await sanitize(content, contentSelector, sanitizationPipeline, sanitizers);
+  async init() {
+    if (!this.initialized) {
+      this.serviceProvidersManifests = await getServiceProviders();
```
MattiSG, line 24: The whole aim of a constructor is to initialize. Since we switch to a class syntax, I believe initialization should be started at instantiation, and a promise exposed on which to chain, so that we expose something equivalent to `new CGUs().isReady.then(…)`.

src/index.js
```diff
@@ -9,60 +11,97 @@ import { persistRaw, persistSanitized, pushChanges } from './history/index.js';
...
+      this.emit('applicationError', serviceProviderId, documentType, error);
+    }
+  }
+
+  get serviceProviders() {
+    return this.serviceProvidersManifests;
+  }
+  get documentsTypes() {
+    return this.documentTypes;
   }
```
MattiSG, line 106: Getters / Setters are by convention better declared before instance methods (and right after the constructor, IIRC).

src/notifier/README.md
```diff
@@ -0,0 +1 @@
+This module is intended to be considered as a simple consumer of the CGU application's events API and therefore to be extracted from this repository. It is still in this repository to facilitate quick iterations but should be extracted as soon as the CGUs app is stabilized.
```
MattiSG, line 1:
```suggestion
This module is intended to be considered as a simple consumer of the CGU application's events API and therefore to be extracted from this repository. It is still in this repository to facilitate quick iterations but is expected to be extracted as soon as the CGUs app is stabilized.
```

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+authentication.apiKey = process.env.SENDINBLUE_API_KEY;
+
+const apiInstance = new sendInBlue.SMTPApi();
+const contactsInstance = new sendInBlue.ContactsApi();
+
+const ADMINISTRATORS_LIST_ID = config.get('notifier.sendInBlue.administratorsListId');
+const UPDATES_LIST_ID = config.get('notifier.sendInBlue.updatesListId');
+const UPDATE_TEMPLATE_ID = config.get('notifier.sendInBlue.updateTemplateId');
+const ERROR_TEMPLATE_ID = config.get('notifier.sendInBlue.errorTemplateId');
+const BASE_URL = config.get('notifier.sendInBlue.baseUrl');
```
MattiSG, line 18: > I don't see the added value of the constants indirection. I wonder if the code would not be simpler if config values were simply obtained directly in the lines that use them.

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+const serviceProvidersMailingLists = {};
+let serviceProviders;
+let documentTypes;
+
+export function init(passedServiceProviders, passedDocumentTypes) {
+  serviceProviders = passedServiceProviders;
+  documentTypes = passedDocumentTypes;
+}
```
MattiSG, line 27: If we're introducing class usage in this app, I believe this clearly calls for using it here as well 🙂

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+export async function onDocumentScrapingError(serviceProviderId, documentTypeId, error) {
+  const sendParams = {
+    templateId: ERROR_TEMPLATE_ID,
+    params: {
+      SERVICE_PROVIDER_NAME: serviceProviders[serviceProviderId].serviceProviderName,
+      DOCUMENT_TYPE: documentTypes[documentTypeId].name,
+      ERROR_TEXT: `An error has occured when trying to scrape the document:
```
MattiSG, line 35:
```suggestion
      ERROR_TEXT: `An error occured when trying to scrape the document:
```

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+  return send([ADMINISTRATORS_LIST_ID, UPDATES_LIST_ID], sendParams);
+};
+
+export async function onApplicationError(serviceProviderId, documentTypeId, error) {
+  const sendParams = {
+    templateId: ERROR_TEMPLATE_ID,
+    params: {
+      SERVICE_PROVIDER_NAME: serviceProviders[serviceProviderId].serviceProviderName,
+      DOCUMENT_TYPE: documentTypes[documentTypeId].name,
+      ERROR_TEXT: `An error has occured when trying to update the document:
```
MattiSG, line 62:
```suggestion
      ERROR_TEXT: `An error occured when trying to update the document:
```

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+async function send(lists, sendParams) {
+  let contacts = [];
+
+  for (let listId of lists) {
+    const listContacts = await getListContacts(listId);
+    contacts = contacts.concat(...listContacts);
+  }
+
+  const uniqueContacts = contacts.reduce((acc, current) => acc.find(contact => contact.id === current.id) ? acc : acc.concat([current]), []);
```
MattiSG, line 78: Wouldn't it be simpler to use a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)? 🙂

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+  const sendPromises = [];
+  uniqueContacts.forEach(contact => {
```
MattiSG, line 81: A [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) would be more appropriate considering the operation type 🙂

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+async function getListContacts(listId) {
+  let offset = 0;
+  let limit = 50;
+  const list = await contactsInstance.getList(listId);
+
+  return recursivePaginationAccumulator('getContactsFromList', 'contacts', listId, limit, offset, [], list.totalSubscribers);
+}
+
+async function recursivePaginationAccumulator(apiFunctionName, resultKey, resourceId, limit, offset, accumulator, count) {
```
MattiSG, line 99: Use a verb for the function name?
```suggestion
async function getAllPaginatedEntries(functionName, resourceIdParameter, resultKey, accumulator, expectedResultCount, offset = 0, paginationSize = 50) {
```
(⚠️ this suggestion switches the 2nd and 3rd parameters, switches the 4th and 5th and moves them to the end)

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+      to: [{ email: contact.email }]
+    }));
+  });
+
+  return Promise.all(sendPromises);
+}
+
+async function getListContacts(listId) {
+  let offset = 0;
+  let limit = 50;
```
MattiSG, line 93: I don't understand why this is passed as a parameter to the accumulator, it feels like it is an internal optimisation of the pagination size that has no point in being changed from the outside.

src/notifier/index.js
```diff
@@ -0,0 +1,107 @@
...
+      ...sendParams,
+      to: [{ email: contact.email }]
+    }));
+  });
+
+  return Promise.all(sendPromises);
+}
+
+async function getListContacts(listId) {
+  let offset = 0;
```
MattiSG, line 92: I feel that this parameter should be optional for the accumulator and default to `0`.

config/development.json
Ndpnt: I agree, not useful at all! (dev's fantasy… :D)

index.js
Ndpnt: 👍

index.js
Ndpnt: Yep I agree.

src/index.js
Ndpnt: Useless. I converted from simple `let` declarations to a class too quickly

src/index.js
Ndpnt: To be discussed

src/notifier/index.js
Ndpnt: I hesitated, but as we have to use Set on an array it gives that:
`const uniqueAddresses = Array.from(new Set(contacts.map(contact => contact.id))).map(id => contacts.find(contact => contact.id === id))`
Do you find it simpler?

src/index.js
Ndpnt: It seems to be a bad practice to execute async code in constructor (see [here](https://stackoverflow.com/questions/43431550/async-await-class-constructor) and [here](https://stackoverflow.com/questions/24398699/is-it-bad-practice-to-have-a-constructor-function-return-a-promise)) as the constructor should be a pure function.
So to expose something really simple to use, I suggest to do something like that:
````
export default class CGUs extends events.EventEmitter {
  constructor() {
    super();
    this.documentTypes = DOCUMENTS_TYPES;
  }

  async init() {
    if (!this.initialized) {
      this.serviceProvidersManifests = await getServiceProviders();
      this.initialized = Promise.resolve();
    }

    return this.initialized;
  }

  async updateTerms() {
    await this.init();

    console.log('Start scraping and saving terms of service…');
    
    […]
  }

  […]
}
```

src/notifier/index.js
MattiSG: After oral discussion: we'd need to use the `id` or `email` as a key, and we will probably want to use other fields (such as `name`) in the future, so the `Set` is not a proper implementation.

PR #39, opened 2020-07-03
Add 21 ready-to-use providers

AdrienFines:
Addition of several providers :
AskFM
Badoo
CafeMom
deviantART
Foursquare
Goodreads
hi5
LINE
LastFM
LibraryThing
LinguaLeo
LiveJournal
Match
Meetup
Mix
MouthShut
Ning
Tagged
Viber
WeHeartIt
WhatsApp

All of them don't need more sanitizing.


Ndpnt:
Good job 👍!
In order to have a meaningful history, it's better to avoid having all your issues and their fixes as independent commits (like commits related to tumblr, for example). 
And in this specific case, the whole intention is to add providers. So, can you squash all your commit in one please?

_PS: Do not hesitate to ask me if you don't know how to do :)._

AdrienFines:
I've messed up with rebasing, so I will open another cleaner PR with squashed commits.

PR #41, opened 2020-07-07
Let sanitizer download images locally

LucasVerneyDGE:
Hi,

Currently, the HTML to markdown transformation keeps the actual src of images. This can be seen in https://github.com/ambanum/CGUs-data/blob/master/sanitized/Facebook/developer_policy.md for instance, where rendering the markdown document makes external calls to Facebook servers. This means the images are not archived.

Images in this policy are not very important but they can be required to archive for some policies. E.g. https://developer.android.com/distribute/marketing-tools/brand-guidelines.

This PR should add support for locally downloading the images and embedding them as data URIs in the resulting markdown file.

Best,

EDIT: [This issue](https://github.com/domchristie/turndown/issues/296) at `turndown` makes me think this code should not work, but it seems to actually work locally for me.


src/sanitizer/index.js
```diff
@@ -1,10 +1,44 @@
...
-export default async function sanitize(content, selector, sanitizationPipeline, providerSpecificSanitizers) {
+    let localSrc = src;
+    if (!src.startsWith('data:')) {  // Not a data-URI, download content
+      let queryUrl = src;
+      if (!src.startsWith('http:') && !src.startsWith('https:')) {  // Relative URL
+        queryUrl = new URL(src, url).href;
+      }
+      const blob = await fetchUrl(queryUrl);
+      const arrayBuffer = await blob.arrayBuffer();
+      localSrc = dataUriParser.format('.png', arrayBuffer);
```
LucasVerneyDGE, line 35: Should not be hard coded.

MattiSG:
As discussed this morning: let's rather use a filter to transform `` into data-uris, so that we minimise the number of cases where images are downloaded 🙂
If we end up seeing such a filter used repeatedly, we'll find a way to expose filters globally.

LucasVerneyDGE:
Here is a refactor turning the "download images" pass into a filter (including a basic implementation of what global filters could be).

Let me know what you think about it!

services/Facebook.filters.js
```diff
@@ -1,4 +1,4 @@
-export function removeHelpButtons(document) {
+export async function removeHelpButtons(document, url) {
```
LucasVerneyDGE, line 1: Note: I had to convert filters into `async` functions to be able to execute fetch etc into them.

services/GooglePlayStore.filters.js
```diff
@@ -0,0 +1 @@
+export { convertImagesToDataURI } from '../src/filter/common.js';
```
LucasVerneyDGE, line 1: This is a basic implementation of what global filters could be.

src/fetcher/index.js
```diff
@@ -22,5 +22,9 @@ export default async function fetch(url) {
 
   const blob = await response.blob();
 
-  return blob.text();
+  if (asText) {
```
LucasVerneyDGE, line 25: This could probably be refactored into a MIME-Type switch, cf #52

MattiSG:
Superseded by #101 🙂

LucasVerneyDGE:
Thanks for taking this over!

PR #47, opened 2020-07-15
Removed noise from Facebook

AdrienFines:
The removeHelpButton function should now remove all links and images.

### DIFF
--- a/services/Facebook.filters.js
+++ b/services/Facebook.filters.js
@@ -1,9 +1,10 @@
 
 export function removeHelpButtons(document) {
   const imgs = document.querySelectorAll('img[src*="https://scontent"]');
-  imgs.forEach(img => {
-    if(img.parentNode.tagName == 'A' && !img.parentNode.text){
-      img.parentNode.remove();
+    imgs.forEach(img => {
+    const parent = img.parentNode;
+    if (parent.tagName == 'A' && !parent.text) {
+      parent.remove();
     }
   });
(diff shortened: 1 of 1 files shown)


services/Facebook.filters.js
```diff
@@ -1,10 +1,9 @@
 
 export function removeHelpButtons(document) {
-  const links = document.querySelectorAll('a[href="#"][id*="-markdown-internal-id"]');
-  links.forEach(link => {
-    // Helps buttons do not have text content
-    if (!link.text) {
-      link.remove();
+  const imgs = document.querySelectorAll('img[src*="https://scontent"]');
+  imgs.forEach(img => {
+    if(img.parentNode.tagName == 'A' && !img.parentNode.text){
```
Ndpnt, line 5: You may use an intermediate variable to avoir repeating `img.parentNode` like this:
```js
    const parent = img.parentNode;
    if (parent.tagName == 'A' && !parent.text) {
      parent.remove();
    }
```

Ndpnt: Also, it misses some spaces before and after parenthesis :).

Ndpnt:
Good job, smart way to fix the issue! 👍 
I hope we will get less email with this. 😄

services/Facebook.filters.js
AdrienFines: Done thanks !

PR #51, opened 2020-07-16
Split snapshot history from version history

Ndpnt:
Before merging:
- Create a tag in the repo `CGUs-data` with current data to have a backup
- Clean the history on `CGUs-data` master, reset hard `master` branch with `clean-history` branch


config/production.json
```diff
@@ -1,6 +1,8 @@
 {
   "history": {
     "dataPath": "../cgus-data",
```
MattiSG, line 3: Why keep this?

config/production.json
```diff
@@ -1,6 +1,8 @@
...
+    "snapshotsPath": "../cgus-data-snapshots",
+    "versionsPath": "../cgus-data",
```
MattiSG, line 5: > Why not use this opportunity to unify production with defaults? 🙂

scripts/setup.js
```diff
@@ -8,14 +8,13 @@ import simpleGit from 'simple-git';
...
 const NEW_HISTORY_MESSAGE = 'Start a new local empty history';
-const DOWNLOAD_HISTORY_MESSAGE = `Download the entire history of terms of services from ${DATA_REPOSITORY_URL}`;
+const DOWNLOAD_HISTORY_MESSAGE = `Download the entire history of terms of services from ${config.get('history.publicSnapshotsRepository')} and ${config.get('history.publicVersionsRepository')}`;
 
 (async () => {
   try {
-    if (fsApi.existsSync(config.get('history.dataPath'))) {
-      return console.log(`It seems that the database is already initialized as the "${config.get('history.dataPath')}" directory already exists. Erase that folder first if you’d like to set up a new database.`);
+    if (fsApi.existsSync(config.get('history.versionsPath')) || fsApi.existsSync(config.get('history.snapshotsPath'))) {
+      return console.log(`It seems that the database is already initialized as the "${config.get('history.versionsPath')}" directory already exists. Erase that folder first if you’d like to set up a new database.`);
```
MattiSG, line 17: This message does not reflect the conditions defined in the code.
```suggestion
      return console.log(`It seems that the database is already initialized as the "${config.get('history.versionsPath')}" or the "${config.get('history.snapshotsPath')}" directories already exist. Erase these folders first (or change the source path in the config) if you’d like to set up a new database.`);
```

scripts/setup.js
```diff
@@ -29,12 +28,16 @@ const DOWNLOAD_HISTORY_MESSAGE = `Download the entire history of terms of servic
     }]);
 
     if (answer.history === NEW_HISTORY_MESSAGE) {
-      await fs.mkdir(config.get('history.dataPath'), { recursive: true });
-      const git = simpleGit(path.resolve(__dirname, '../', config.get('history.dataPath')));
+      await fs.mkdir(config.get('history.snapshotsPath'), { recursive: true });
+      await fs.mkdir(config.get('history.versionsPath'), { recursive: true });
+      let git = simpleGit(path.resolve(__dirname, '../', config.get('history.snapshotsPath')));
+      await git.init();
+      git = simpleGit(path.resolve(__dirname, '../', config.get('history.versionsPath')));
```
MattiSG, line 35: > It would be more efficient not to wait each time, but to instead chain promises so that initializing the first repo doesn't wait upon the creation of the first directory.

scripts/setup.js
```diff
@@ -29,12 +28,16 @@ const DOWNLOAD_HISTORY_MESSAGE = `Download the entire history of terms of servic
...
       await git.init();
     } else {
       const git = simpleGit();
-      await git.clone(DATA_REPOSITORY_URL, config.get('history.dataPath'));
+      await git.clone(config.get('history.publicSnapshotsRepository'), config.get('history.snapshotsPath'));
+      await git.clone(config.get('history.publicVersionsRepository'), config.get('history.versionsPath'));
```
MattiSG, line 40: > Maybe start with the fastest one so it's less frightening for the user when the second download starts?

src/history/index.js
```diff
@@ -14,10 +24,23 @@ export async function recordVersion(serviceId, documentType, content, snapshotId
     throw new Error(`A snapshot ID is required to ensure data consistency for ${serviceId}'s ${documentType}`);
   }
 
-  return record({
+  let id = '';
+  if (config.get('history.publish')) {
+    id += config.get('history.publicSnapshotsRepository')
+      .replace('git@github.com:', 'https://github.com/')
+      .replace('.git', '/commit/');
```
MattiSG, line 31: This makes the assumption that the repository is published on GitHub (and is fragile as it expects the syntax to be SSH and end with `.git`). Let's avoid vendor lock-in and allow the prefix to be defined in config 🙂

src/history/index.js
```diff
@@ -14,10 +24,23 @@ export async function recordVersion(serviceId, documentType, content, snapshotId
...
+  }
+  id += snapshotId;
+
+  return versionRecorder.record({
     serviceId,
     documentType,
     content,
-    snapshotId
+    messageToAppend: `This version was recorded after filtering snapshot ${id}`
```
MattiSG, line 39: > I read this more as the “body” of the commit message, as opposed to the “title” or “subject” of the commit message.
> If we want to stay away from the storage mechanism, maybe “details” would be appropriate?

```suggestion
    messageBody: `This version was recorded after filtering snapshot ${id}`
```

src/history/index.js
```diff
@@ -14,10 +24,23 @@ export async function recordVersion(serviceId, documentType, content, snapshotId
...
   });
 }
+
+export async function publish() {
+  snapshotRecorder.publish();
+  return versionRecorder.publish();
```
MattiSG, line 45: If this is all async, then we should return the sum of both promises. Currently, if the snapshot publish fails, no one will know.

src/history/recorder.js
```diff
@@ -1,73 +1,76 @@
...
-  });
-}
+  async _commit({ filePath, message, resolve }) {
+    await this.git.add(filePath);
+    resolve(await this.git.commit(filePath, message));
+  }
+
+  _commitQueueErrorHandler(error, { filePath, message, reject }) {
+    reject(new Error(`Could not commit ${filePath} with message "${message}" due to error: ${error}`));
+  }
```
MattiSG, line 69: > This could be static or even a helper function as it does not rely on any instance property.

src/history/recorder.test.js
```diff
@@ -1,23 +1,34 @@
...
 
+const SNAPSHOTS_PATH = path.resolve(__dirname, '../..', config.get('history.versionsPath'));
 const SERVICE_PROVIDER_ID = 'test_service';
 const TYPE = 'tos';
 const FILE_CONTENT = 'ToS fixture data with UTF-8 çhãràčtęrs';
-const EXPECTED_FILE_PATH = `${SNAPSHOTS_DIRECTORY}/${SERVICE_PROVIDER_ID}/${TYPES[TYPE].fileName}.html`;
+const EXPECTED_FILE_PATH = `${SNAPSHOTS_PATH}/${SERVICE_PROVIDER_ID}/${TYPES[TYPE].fileName}.html`;
 
 describe('Recorder', () => {
+  let recorder;
```
MattiSG, line 21: > I like to unify my SUT instances varnames 🙂 

```suggestion
  let subject;
```

test/helper.js
```diff
@@ -4,10 +4,13 @@ const fs = fsApi.promises;
 
 import config from 'config';
 
-import { git } from '../src/history/git.js';
+import Git from '../src/history/git.js';
 
 const __dirname = path.dirname(new URL(import.meta.url).pathname);
 
+export const SNAPSHOTS_PATH = path.resolve(__dirname, '../', config.get('history.snapshotsPath'));
+export const VERSIONS_PATH = path.resolve(__dirname, '../', config.get('history.versionsPath'));
```
MattiSG, line 12: Why aren't these imported from the `export` in `../src/history/index.js`?

test/helper.js
```diff
@@ -17,24 +20,30 @@ export async function resetGitRepository() {
 }
 
 async function initRepo() {
-  await git.init();
-  git.addConfig('user.name', config.get('history.author').name)
-     .addConfig('user.email', config.get('history.author').email);
+  for (const repoPath of [VERSIONS_PATH, SNAPSHOTS_PATH]) {
```
MattiSG, line 23:
```suggestion
  for (const repoPath of [ VERSIONS_PATH, SNAPSHOTS_PATH ]) {
```

src/history/git.js
```diff
@@ -2,45 +2,52 @@ import path from 'path';
...
+  }
+
+  async isNew(filepath) {
+    const status = await this.git.status();
+    return (status.created.indexOf(this.relativePath(filepath)) > -1) ||
+           (status.not_added.indexOf(this.relativePath(filepath)) > -1);
+  }
+
+  addConfig() {
+    return this.git.addConfig();
```
MattiSG, line 46: Where are the parameters?

test/helper.js
```diff
@@ -17,24 +20,30 @@ export async function resetGitRepository() {
...
-    } else if (file.name !== 'README.md') {
-      return fs.unlink(filePath);
-    }
-  });
+  const promises = [];
+
+  for (const repoPath of [VERSIONS_PATH, SNAPSHOTS_PATH]) {
+    const files = await fs.readdir(repoPath, { withFileTypes: true });
+
+    promises.push(...files.map(file => {
```
MattiSG, line 37:
```suggestion
    promises.concat(files.map(file => {
```

ops/app.yml
```diff
@@ -4,4 +4,5 @@
 
   roles:
     - cgus-data
```
MattiSG, line 6: Why not call this one `cgus-data-versions` if a new `cgus-data-snapshots` is created?

test/helper.js
Ndpnt: Not exactly :). It should be:
```
promises = promises.concat(files.map(file => {
```
so, instead of 
```
promises.push(...files.map(file => {
```
I'm not sure which one is better…? Do you have any preference?

ops/app.yml
Ndpnt: Because I prefer considering that `cgus-data` is the real database and the snapshot database is only a backup to recover from filtering errors.

test/helper.js
MattiSG: Ugh… Thanks for the correction and sorry for the noise.

Both look as complex now. I probably have a small preference for `.push` as it is clearer that it is an in-place operation extending the array rather than a full change of its content.

ops/app.yml
MattiSG: What about we name both `cgus-versions` and `cgus-snapshots`? 🙂

Ndpnt: okay 👍

src/history/index.js
Ndpnt: Ok but if remove the `replace`s I have to add a config key for the link in body commit message for the link to the snapshot commit. Are you ok with that? What about `snapshotsBaseUrl`?

config/production.json
Ndpnt: In production I prefer having the three repositories at the same level but I have not impactful arguments…

MattiSG: No worries, I'm happy with the current option if it feels better to you, I just like to have as little difference as possible between dev and production :wink:

src/history/index.js
MattiSG: Yes, adding a config key was what I had in mind! `snapshotsBaseUrl` sounds good 👍

test/helper.js
Ndpnt: I prefer having no spaces between array brackets and other tokens.
Is it important to you and if it is the case can you explain me why please? :)

MattiSG: I feel the inner spaces strongly improve readability:

1. It makes it easier to visually distinguish `{`, `[`, `(`.
2. It makes it easier to visually distinguish function calls and boolean groupings from data structures.
3. It unifies the distance between each member of a data structure (otherwise only the innermost members are separated by spaces, not the first and last).

I can live without, but it does really help me to parse the code.

Ndpnt: What is the English expression to say "Vendu!"? =)

MattiSG: “Deal!” 😉

Ndpnt: Thanks! :) 
So, deal!

PR #53, opened 2020-07-20
Multi selectoring

AdrienFines:
This PR adds the new service Skype, and improves slightly the filtering process by allowing CSS multi-selectoring (for instance ".pageIntro, .div_content" selects two elements). 
Skype has two terms of services : one official, and another inside Microsoft's TOS. The official is under the "tos" document name, and I put the second under the "secondary-tos" name. The "secondary-tos" name isn't supported yet, if you want to test it I suggest replacing it with "devAgreement" instead.

### DIFF
--- a/services/Skype.filters.js
+++ b/services/Skype.filters.js
@@ -1,4 +1,4 @@
-export function removeIrrelevantModules(document) {
+export function removeIrrelevantModulesPrivacyPolicy(document) {
     const moduleDescriptions = document.querySelectorAll('.divModuleDescription');
     const toDelete = [
         "mainenterprisedeveloperproductsmodule",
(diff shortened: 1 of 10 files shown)


AdrienFines:
I don't know why two tests are not passing, apparently the EOL character and \n are different ?

MattiSG:
No, both fail because `AssertionError: expected '' to equal ' First provider TOS\n\nTerms of […]`. We're returning the empty string. I suspect this is a consequence of always enforcing a selector. It is likely that these tests do not pass a selector and expect to receive the full body.

services/Skype.filters.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export function removeIrrelevantModulesParentTos(document) {
+    const moduleDescriptions = document.querySelectorAll('.divModuleDescription');
+    moduleDescriptions.forEach(moduleDescription => {
+        var name = moduleDescription.querySelector('.FileName');
+        if (name) {
+            name = name.textContent;
+        }
+        if (name.substr(0,2)=="14" && name.substr(0,3)!="14e" && name.substr(0,3)!="14_") {
```
MattiSG, line 63: This would be more readable with a RegExp.

I'll handle it 🙂

services/Skype.filters.js
```diff
@@ -0,0 +1,95 @@
...
+            moduleDescription.remove();
+        }
+        if (name=="serviceslist") {
```
MattiSG, line 66: Spaces before and after comparison operators.

 I'll handle it 🙂

services/Skype.filters.js
```diff
@@ -0,0 +1,95 @@
...
+    document.querySelectorAll('#moduleName, .FileName').forEach(element => element.remove());
+}
+
+export function removeInvisibleText(document) {
+    const notDisplayed = document.querySelectorAll('.displayNone');
+    const summaries = document.querySelectorAll('.printsummary');
+    const fulltexts = document.querySelectorAll('.printDetailedContent');
+    notDisplayed.forEach(element => element.remove());
+    summaries.forEach(summary =>  {
+        if (summary.textContent=="Summary") {
```
MattiSG, line 86: This would be more elegant and readable with an [`Array.filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).

I'll handle it 🙂

services/Skype.filters.js
```diff
@@ -0,0 +1,95 @@
...
+            summary.remove();
+        }
+    });
+    fulltexts.forEach(fulltext => {
+        if (fulltext.textContent=="Full text") {
```
MattiSG, line 91: Why do we add these checks, both for Summary and Full text? When I check, I see no case where the class is set and the text content is different 🤔

services/Skype.json
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "Skype",
+    "documents": {
+        "tos": {
+            "location": "https://www.skype.com/en/legal/ios/tos/",
+            "contentSelector": "div.row:nth-child(2) > div:nth-child(1)"
```
MattiSG, line 6: Selecting the first child is precisely the standard behavior of CSS selectors 🙂 

```suggestion
            "contentSelector": "div.row:nth-child(2) > div"
```

src/filter/index.js
```diff
@@ -15,16 +15,14 @@ export default async function filter(content, selector, filterNames, filterFunct
...
 
-  if (selector) {
-    const selectedContent = webPageDOM.querySelector(selector);
+  let selectedContent = [ webPageDOM ];
 
-    if (selectedContent) {
-      contentToFilter = selectedContent;
-    } else {
-      console.warn(`The provided selector "${selector}" has no match in the web page.`)
+  if (selector) {
```
MattiSG, line 20: We agreed that, since this is validated in the schema, a `selector` will always be present. This test is thus irrelevant.

I'll take care of that change.

src/loader/index.test.js
```diff
@@ -15,7 +15,7 @@ const expected = {
     documents: {
       tos: {
         location: 'https://www.servicea.example/tos',
-        contentSelector: 'main'
+        contentSelector: 'body'
```
MattiSG, line 18: Good catch! :+1:

services/Skype.filters.js
AdrienFines: I thought it would be more robust, just in case, but tell me if this is irrelevant :)

MattiSG: Thanks for clarifying! It is impossible to predict what “robustness” looks like until we've seen which evolutions are usual. Given the current lack of information, this condition makes the reader assume that there was good reason for checking the content, and this currently useless code will stay in place and possibly be maintained for no reason. Always strive for the simplest piece of code that can achieve the result 🙂

I've refactored the code already, and I'll let you remove this condition just so you can get a feel of how the map/reduce approach improves the maintainability of the code compared to the imperative approach.

AdrienFines: Done thanks for your useful advices ! :)

PR #54, opened 2020-07-21
More noise removing

AdrienFines:
Removal of several invisible elements as well as solving some minor problems.
Especially, it turns out that ":nth-child(1)" is necessary in Skype.json to avoid duplication of texts (since now we have querySelectorAll and not only querySelector I think).

### DIFF
--- a/services/Reddit.filters.js
+++ b/services/Reddit.filters.js
@@ -1,0 +1,3 @@
+export function removeOptionsList(document) {
+  document.querySelectorAll('.filter-holder').forEach(element => element.remove());
+}
--- a/services/Reddit.json
+++ b/services/Reddit.json
@@ -3,11 +3,17 @@
     "documents": {
         "tos": {
             "location": "https://www.redditinc.com/policies/user-agreement",
-            "contentSelector": "#content"
+            "contentSelector": "#content",
+            "filters": [
+                "removeOptionsList"
+            ]
         },
         "privacy": {
             "location": "https://www.redditinc.com/policies/privacy-policy",
-            "contentSelector": "div#content"
+            "contentSelector": "div#content",
+            "filters": [
+                "removeOptionsList"
+            ]
         }
     }
 }
--- a/services/Reddit.json
+++ b/services/Reddit.json
@@ -10,7 +10,7 @@
         },
         "privacy": {
             "location": "https://www.redditinc.com/policies/privacy-policy",
-            "contentSelector": "div#content",
+            "contentSelector": "#content",
             "filters": [
                 "removeOptionsList"
             ]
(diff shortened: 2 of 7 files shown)


MattiSG:
> it turns out that `:nth-child(1)` is necessary in Skype.json

I'm sorry, I gave you improper explanations when we paired 😞
The case where the `:nth-child(1)` pseudo-selector was not necessary was an edge case related to the use of the adjacent selector `+`, which always targets a single element.

In any case, it is more idiomatic to use the `:first-child` pseudo-selector rather than `:nth-child(1)` 🙂

services/Reddit.filters.js
```diff
@@ -0,0 +1,3 @@
+export function removeOptionsList(document) {
+  document.querySelectorAll('.filter-holder').forEach(element => element.remove());
```
MattiSG, line 2: > I see only one `.filter-holder` in the document. Using `document.querySelector('.filter-holder').remove()` would be simpler 🙂

services/Reddit.json
```diff
@@ -3,11 +3,17 @@
...
-            "contentSelector": "#content"
+            "contentSelector": "#content",
+            "filters": [
+                "removeOptionsList"
+            ]
         },
         "privacy": {
             "location": "https://www.redditinc.com/policies/privacy-policy",
-            "contentSelector": "div#content"
+            "contentSelector": "div#content",
```
MattiSG, line 13: When you target with an ID, the tag is (in the vast majority of cases, and theoretically always) not needed, since an ID is expected to be unique in the page.

```suggestion
            "contentSelector": "#content",
```

services/Skype.json
```diff
@@ -7,7 +7,7 @@
         },
         "privacy": {
             "location": "https://privacy.microsoft.com/en-gb/privacystatement",
-            "contentSelector": "h1, #psp_last_updated, span.Description, .div_content",
+            "contentSelector": "h1, #psp_last_updated, span.Description:nth-child(1), .div_content",
```
MattiSG, line 10: More idiomatic 🙂 
```suggestion
            "contentSelector": "h1, #psp_last_updated, span.Description:first-child, .div_content",
```

services/Twitter.json
```diff
@@ -10,7 +10,10 @@
     },
     "privacy": {
       "location": "https://twitter.com/privacy",
-      "contentSelector": "main"
+      "contentSelector": "main",
+      "filters": [
+        "removeNotDisplayedElements"
```
MattiSG, line 15: Yup, we could have done that earlier 😅

services/Skype.json
MattiSG: Actually after looking into this selector and the structure of the page more in depth, it turns out what we really want is to include the `.Description` only for the `.header_content`, which turns out to also be the only one to be a `:first-child` of its parent, but it is a coincidence 🙂

So what we want is:
```suggestion
            "contentSelector": "h1, #psp_last_updated, .header_content .Description, .div_content",
```

PR #56, opened 2020-07-24
Convert HTML tables to Markdown

MattiSG:
The following services contain at least one `<table>`. They have a checkbox associated with them that should be checked when the result of running the filters on them has been double-checked.

- [x] AskFM
- [x] Badoo
- [x] deviantART
- [x] MouthShut
- [x] StackOverflow
- [x] WeChat
- [x] Yelp

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2303,6 +2303,11 @@
         "jsdom": "^16.2.0"
       }
     },
+    "turndown-plugin-gfm": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/turndown-plugin-gfm/-/turndown-plugin-gfm-1.0.2.tgz",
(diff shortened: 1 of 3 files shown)


MattiSG:
@AdrienFines could you please double-check the impact on the last two services? 🙂 I already checked MouthShut's Privacy Policy, no change.

AdrienFines:
I checked each service and everything seems fine. There is only one little flaw : when no column headers are found, one empty line is added at the beginning of the table (for example in StackOverflow Privacy Policy). Yet I don't think it will be that annoying.

MattiSG:
> when no column headers are found, one empty line is added

Yup, saw that too. It is a workaround added in https://github.com/domchristie/turndown-plugin-gfm/pull/10.

Thanks for the review! 😃

PR #62, opened 2020-07-28
Select EU jurisdiction for LibraryThing, Snapchat and TikTok

AdrienFines:


services/LibraryThing.filters.js
```diff
@@ -0,0 +1,38 @@
+export function selectOnlyTOS(document) {
+	const h2Titles = document.querySelectorAll('h2');
+	const titlePP = h2Titles[0];
+	const titleTOS = h2Titles[2];
```
Ndpnt, line 4: I think it's more robust to select section title by using text content, for example:
```js
const sectionTitles = Array.from(document.querySelectorAll('h2'));
const TOSTitle = sectionTitles.filter(el => el.textContent === "Terms of Service");
```

services/Snapchat.filters.js
```diff
@@ -0,0 +1,15 @@
+export function selectOnlyEuropeanTOS(document) {
+  //Select titles for juridictions
```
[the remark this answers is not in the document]
AdrienFines: Thanks it's done ! 
What do you prefer for spaces and tabs ? Only spaces or only tabs ? I've just configured my editor so I should'nt be encoutering this probleme again ! :)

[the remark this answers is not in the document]
Ndpnt: Cool ;). 
We use spaces with length 2.

Ndpnt:
It works but I think it will be better to use [Range](https://developer.mozilla.org/en-US/docs/Web/API/Range/Range) instead of walking through each node element. Can you do the changes everywhere it is appropriate?

AdrienFines:
Done :) TikTok.filters.js is up to date with master (meaning using newer CSS selector).

services/TikTok.filters.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const range2 = document.createRange();
+  range2.setStartBefore(titleIndia);
+  range2.setEndBefore(lastElement);
+  range2.deleteContents();
+}
+export function selectOnlyEuropeanPrivacyPolicy(document) {
+  const firstElement = document.querySelector('article > *:first-child')
+  const titleEuropean = document.querySelector('#privacy-eea');
+  const titleIndia = document.querySelector('#privacy-in')
```
Ndpnt, line 20: There are no India title on this page.

services/TikTok.filters.js
```diff
@@ -0,0 +1,32 @@
...
+  const lastElement = document.querySelector('article > *:last-child')
+
+  const range1 = document.createRange();
+  range1.setStartBefore(firstElement);
+  range1.setEndBefore(europeanSectionTitle);
```
Ndpnt, line 25: `europeanSectionTitle` is not defined.

services/TikTok.filters.js
```diff
@@ -0,0 +1,32 @@
+export function selectOnlyEuropeanTOS(document) {
+  const firstElement = document.querySelector('article > *:first-child')
+  const titleEuropean = document.querySelector('#terms-eea');
+  const titleIndia = document.querySelector('#terms-in')
+  const lastElement = document.querySelector('article > *:last-child')
+
+  const range1 = document.createRange();
+  range1.setStartBefore(firstElement);
+  range1.setEndBefore(europeanSectionTitle);
```
Ndpnt, line 9: `europeanSectionTitle` is not defined.

services/TikTok.filters.js
```diff
@@ -0,0 +1,32 @@
+export function selectOnlyEuropeanTOS(document) {
+  const firstElement = document.querySelector('article > *:first-child')
+  const titleEuropean = document.querySelector('#terms-eea');
+  const titleIndia = document.querySelector('#terms-in')
+  const lastElement = document.querySelector('article > *:last-child')
+
+  const range1 = document.createRange();
```
Ndpnt, line 7: var name like `range1` is not very explicit. Try to explicit the meaning of the variable, for example `otherJuridictionsContent` or something like that.

services/TikTok.filters.js
AdrienFines: Oops sorry a bad copy-paste modification... It was meant to be "#privacy-row".

PR #64, opened 2020-07-29
Number MySpace lists correctly

AdrienFines:
In MySpace documents, lists encapsulated in <ol> tags were badly numbered because invisible tag which were not <li> tags were disrupting numbering. This PR solve the issue.


AdrienFines:
Done !

Ndpnt:
I rewrote the history because there were some issues :)

PR #65, opened 2020-07-29
Fix email sending on document fetch error

Ndpnt:


MattiSG:
Aaah, so that's the reason why we weren't getting any error email… When reviewing #51 I realized TikTok has an empty selector and MouthShut yields a 503, but we weren't notified.

PR #66, opened 2020-07-29
Add a wiki about noise types

AdrienFines:
Create a wiki about noise types and examples of how to solve them.

I don't know whether this will be in the official "Wiki" page of the repo, but as I didn't find a convenient way of asking for a PR through official Wiki, I've asked it the classic way :)

Don't hesitate to comment !

### DIFF
--- a/usual-noise-wiki.md
+++ b/usual-noise-wiki.md
@@ -1,0 +1,156 @@
+# Usual noise
+
+Generally speaking, noises are unwanted content in versions. At the moment we found three types of noise.
+
+### Irrelevant content
+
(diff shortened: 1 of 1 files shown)


MattiSG:
Great, thanks a lot @AdrienFines, that's good documentation! :smiley:

Please use the [wiki](https://github.com/ambanum/CGUs/wiki) for this content. You're right, there is no review process in the wiki: it is simply edit-and-forget. It's fine for this kind of “recipes”: we want to give examples and decrease the friction to contribution.

I'm closing this PR as we won't integrate this documentation in the repo itself  :)

PR #67, opened 2020-07-30
Convert relative URLs to absolute URLs

AdrienFines:
This is achieved using npm packages "is-relative-url" and "url-toolkit".

I have modified tests accordingly but I am unsure if I have done it properly.

### DIFF
--- a/src/filter/index.js
+++ b/src/filter/index.js
@@ -1,13 +1,15 @@
 import TurndownService from 'turndown';
 import turndownPluginGithubFlavouredMarkdown from 'joplin-turndown-plugin-gfm';
 import jsdom from 'jsdom';
+import urlToolkit from 'url-toolkit'
+import isRelativeUrl from 'is-relative-url'
 
 const { JSDOM } = jsdom;
 const turndownService = new TurndownService();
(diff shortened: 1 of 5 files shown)


src/filter/index.js
```diff
@@ -1,13 +1,15 @@
 import TurndownService from 'turndown';
 import turndownPluginGithubFlavouredMarkdown from 'joplin-turndown-plugin-gfm';
 import jsdom from 'jsdom';
+import urlToolkit from 'url-toolkit'
+import isRelativeUrl from 'is-relative-url'
```
Ndpnt: it seems that this package is not useful, `url-toolkit` do nothing if the url is not relative.

src/filter/index.js
```diff
@@ -17,6 +19,13 @@ export default async function filter(content, selector, filterNames, filterFunct
     });
   }
 
+  Array.from(webPageDOM.querySelectorAll('a')).map(link => {
```
Ndpnt: Improve selector instead of checking if the href is defined, like this `'a:not([href=""])'`.

Ndpnt: Extract this in a separate function

PR #69, opened 2020-07-31
Handling terms in PDF

LucasVerneyDGE:
Hi,

Here is a first PoC of what handling PDF files might look like. Please note that this PR is built upon https://github.com/ambanum/CGUs/pull/41 for easier getting started / merging.

It is currently using https://www.npmjs.com/package/pdf2html (which relies on Apache Tika underneath, only extra requirement is to have Java JRE available). I gave a try with `poppler-utils`'s `pdftohtml` as well, but the result was worse.

Result is still far from perfect, but this gives a rough idea of what it could look like I think.

Best,


Ndpnt:
Superseded by #109 🙂

LucasVerneyDGE:
Thanks for the work on this!

MattiSG:
Thanks for your suggestion and implementation demonstration! We're glad to have first-level support for PDF files 😃

PR #79, opened 2020-08-05
Publish data before sending notifications

Ndpnt:


schedule.js
```diff
@@ -13,12 +13,24 @@ import CGUs from './src/index.js';
...
-    app.on('error', notifier.onApplicationError.bind(notifier));
+    const delayedVersionNotitificationsParams = [];
+
+    app.on('versionRecorded', (serviceId, type, versionId) => {
+      delayedVersionNotitificationsParams.push({ serviceId, type, versionId });
+    });
+
+    app.on('changesPublished', () => {
+      delayedVersionNotitificationsParams.forEach(({ serviceId, type, versionId }) => {
+        notifier.onVersionRecorded(serviceId, type, versionId);
```
MattiSG, line 26: > Rename the function, since it is not really a handler anymore?

schedule.js
```diff
@@ -13,12 +13,24 @@ import CGUs from './src/index.js';
 
     const app = new CGUs();
     await app.init();
-
     const notifier = new Notifier(app.serviceDeclarations, app.documentTypes);
 
-    app.on('versionRecorded', notifier.onVersionRecorded.bind(notifier));
-    app.on('fetchingError', notifier.onDocumentFetchError.bind(notifier));
-    app.on('error', notifier.onApplicationError.bind(notifier));
+    const delayedVersionNotitificationsParams = [];
```
MattiSG, line 18:
```suggestion
    const delayedVersionNotificationsParams = [];
```

src/index.js
```diff
@@ -40,32 +40,43 @@ export default class CGUs extends events.EventEmitter {
...
+      await Promise.all(documentTrackingPromises);
+
+      if (config.get('history.publish')) {
+        try {
+          await publish();
+          this.emit('changesPublished');
+          console.log('Changes published');
+        } catch (error) {
+          console.error(`Error when trying to publish changes: ${error}`);
+          this.emit('publicationError', error);
```
MattiSG, line 74: > Unify the order of logging & emitting events 😰

schedule.js
Ndpnt: ok, in this case, I also rename the other exposed functions for consistency: 
`onVersionRecorded` -> `notifyVersionRecorded` or `versionRecorded`
`onDocumentFetchError` -> `documentFetchError` or …
`onDocumentUpdateError` -> `documentUpdateError` or …
`onApplicationError` -> `applicationError` or …
?

MattiSG: Never mind, I didn't properly take into account the file… I somehow thought we were in the notifier itself, not in the caller. Sorry, this comment is not appropriate 😞

PR #86, opened 2020-08-06
Add WeChat documents

AdrienFines:
One document isn't followed because of JS but is documented in the wiki.

### DIFF
--- a/services/WeChatOpenPlatform.json
+++ b/services/WeChatOpenPlatform.json
@@ -1,0 +1,20 @@
+{
+    "name": "WeChat Open Platform",
+    "documents": {
+        "tosParent": {
+            "fetch": "https://www.tencent.com/en-us/service-agreement.html",
+            "select": ".page-content"
+        },
+        "devAgreement": {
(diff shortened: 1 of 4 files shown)


services/WeChatOpenPlatform.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "WeChat",
```
MattiSG, line 2:
```suggestion
    "name": "WeChat Open Platform",
```

services/WeChatOpenPlatform.json
```diff
@@ -0,0 +1,20 @@
...
+            "contentSelector": ".page-content"
+        },
+        "devAgreement": {
+            "location": "https://open.weixin.qq.com/cgi-bin/frame?t=news/protocol_developer_tmpl",
+            "contentSelector": "#en_US_tpl",
+            "filters": [
+                "convertScriptToDiv"
+            ]
+        },
+        "softwareLicenseAgreement": {
```
MattiSG, line 15: Shouldn't this be `tosParent` instead? 🤔

AdrienFines:
Should be ok now :)

PR #90, opened 2020-08-10
Refactor document types

Ndpnt:


CONTRIBUTING.md
```diff
@@ -22,7 +22,28 @@ In the folder `services`, create a JSON file with the name of the service you wa
 
 ### Document type
 
-For the `<document type>` key, you will have to use one of those listed in `/src/types.js` (or create a new one there if it is not already referenced).
+For the `<document type>` key, you will have to use one of those listed in `/src/types.js`.
```
MattiSG, line 25:
```suggestion
For the `<document type>` key, you will have to use one of those listed in `/src/types.json`.
```

CONTRIBUTING.md
```diff
@@ -22,7 +22,28 @@ In the folder `services`, create a JSON file with the name of the service you wa
...
+Each document type is characterised along three dimensions of the contractual relationship, the `writer` of the contract, the targeted `audience` and the `object` of commitment.
```
MattiSG, line 26:
```suggestion
Each document type is characterised along three dimensions of the contractual relationship, the `writer` of the contract, the targeted `audience` and the `object` of the commitment.
```

CONTRIBUTING.md
```diff
@@ -22,7 +22,28 @@ In the folder `services`, create a JSON file with the name of the service you wa
...
+
+
+Document type declaration look like:
```
MattiSG, line 29:
```suggestion
Document type declarations look like:
```

package.json
```diff
@@ -8,17 +8,17 @@
     "node": ">=12.17"
   },
   "scripts": {
-    "start": "node index.js",
-    "start:prod": "NODE_ENV=production node index.js",
-    "start:scheduler": "node schedule.js",
-    "start:scheduler:prod": "NODE_ENV=production node schedule.js",
-    "setup": "node scripts/setup.js",
+    "start": "node --experimental-json-modules index.js",
```
MattiSG, line 11: I would prefer that we use [`createRequire`](http://nodejs.org/api/modules.html#modules_module_createrequire_filename) when needed, [as recommended](https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff#30).

src/types.json
```diff
@@ -0,0 +1,79 @@
+{
+  "Terms of service": {
```
MattiSG, line 2: Let's use title case everywhere 🙂

```suggestion
  "Terms of Service": {
```

src/types.json
```diff
@@ -0,0 +1,79 @@
...
+    }
+  },
+  "Developer Agreement": {
+    "commitment": {
+      "writer": "service provider",
+      "audience": "developer",
+      "object": "API and Content usage"
+    }
+  },
+  "Developer Policy": {
```
MattiSG, line 23: What's the difference with `Developer Agreement`?

Ndpnt:
> Congrats for the tryptic descriptions, I personally find them very clear and useful!

Great :)

> I love how simpler this all becomes after unifying file name, ID and name! 😃

Totally! 😃

src/types.json
Ndpnt: I didn't find one. But Twitter define both.
If you can explain the difference:
https://developer.twitter.com/fr/developer-terms/agreement
https://developer.twitter.com/fr/developer-terms/policy

package.json
Ndpnt: I do not see where it's recommended. 
They recommend to use createRequire instead of `--experimental-modules` but here I use `--experimental-json-modules` and about this one, they said:

> A separate flag --experimental-json-modules enables experimental support for import of JSON files. There is work in progress for standardizing this feature with browsers, and Node.js hopes to align our support with the eventual standard.

Do you still prefer recreating the require function?

CONTRIBUTING.md
Ndpnt: We may add a link to the file

src/types.json
MattiSG: Thanks.
I feel that the “agreement” is a full, legally binding contract, whereas the “policy” describes what Twitter expects and how it can be expected to behave.

I find this separation very interesting, but a bit uncommon as well.

I see two options, in order of preference:

1. Rewrite the `object` of the Developer Policy as “expected behavior of and towards third-party services”.
2. Stop tracking the Developer Policy for Twitter as of now, since we'll need to cover the specific case of Twitter and its multitude of documents in some special way anyway.

I'll let you decide.

package.json
Ndpnt: Moreover, it's only required for the `npm run validate` command.

MattiSG: > I do not see where it's recommended.

You're right, I'm sorry, the second paragraph you mentioned does add details to the initial statement.

> Do you still prefer recreating the require function?

Yes, considering the stage of advancement of the standardisation for JSON modules (see https://github.com/whatwg/html/issues/4315), I would feel much safer relying on a stable API rather than a flag whose behavior is less specified.

src/types.json
Ndpnt: If we keep both, which one will you use for theses documents?
- https://developers.facebook.com/policy
- https://help.instagram.com/325135857663734
- https://play.google.com/intl/ALL_en/about/developer-distribution-agreement.html

package.json
Ndpnt: Okay :)

src/types.json
MattiSG: Okay, it's a mess 🙃

In favor of option 2 then.

Ndpnt: Ok, if we keep one, which one do you prefer, "Developer Terms", "Developer Policy", "Developer Agreement"?

MattiSG: “Policy” seems more common, isn't it? Otherwise I'd say “Developer Terms”, by similarity with “Terms of Service“.

Ndpnt: Yep, I also found that "Policy" seems more common but as you explained for Twitter where “agreement” is a full, legally binding contract, whereas the “policy” describes what Twitter expects and how it can be expected to behave, "agreement" seems to be more in line with what we want. 
Another point, if we keep "Policy", for Twitter the url is not the one which target Twitter Policy but this one "https://developer.twitter.com/developer-terms/agreement" which target the agreement.
So, even if "Developer Terms" seems less common, I think it avoids confusion while still being very understandable (and similar with “Terms of Service“ as you said), so I vote for this one.

Ndpnt:
@michielbdejong What do you think about this document types refactor?

michielbdejong:
Great! So far we haven't used document type strictly in tosback.org, nor in tosdr.org, so I'll probably keep a crawler running somewhere that also crawls "other" documents that don't fit in this more structured list of documents by type by service. But I'll try to assign one of the new document types to each one! Maybe that can also be part of the user interface of edit.tosdr.org, so volunteers can do this work.

PR #94, opened 2020-08-12
Do not notify in case of changes due to a change in filtering

Ndpnt:


src/index.js
```diff
@@ -106,8 +215,23 @@ export default class CGUs extends events.EventEmitter {
...
-      this.emit('documentUpdateError', serviceId, type, error);
+      console.error(`${logPrefix} RecordVersionError:`, error.message);
+      this.emit('recordVersionError', serviceId, type, error);
+    }
+  }
+
+  async publish() {
+    if (!config.get('history.publish')) {
+      return console.log('Publication disabled for this environment.');
+    }
```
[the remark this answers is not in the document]
Ndpnt: Do you prefer that I remove the log?

src/history/recorder.js
```diff
@@ -41,34 +40,48 @@ export default class Recorder {
...
+    if (commits.length) {
+      result.id = commits[0].hash;
+    }
+
+    return result;
   }
 
-  async publish() {
-    return this.git.pushChanges();
+  _filePath(serviceId, documentType) {
```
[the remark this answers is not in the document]
Ndpnt: I get a `SyntaxError` when I try to use a `#`

.eslintrc.yaml
```diff
@@ -36,6 +36,7 @@ rules:
   - error
   - 2
   max-len: 0
+  class-methods-use-this: 0
```
MattiSG, line 39: > I find this warning relevant.

src/history/git.js
```diff
@@ -45,6 +45,19 @@ export default class Git {
            || (status.not_added.indexOf(this.relativePath(filepath)) > -1);
   }
 
+  async log(options = {}) {
+    try {
+      options.file = options.file ? this.relativePath(options.file) : options.file;
```
MattiSG, line 50:
```suggestion
      options.file = options.file && this.relativePath(options.file);
```

src/history/index.js
```diff
@@ -12,24 +12,55 @@ const snapshotRecorder = new Recorder({ path: SNAPSHOTS_PATH, fileExtension: 'ht
...
+
+async function _recordVersion({ serviceId, documentType, content, snapshotId, isRefiltering }) {
   if (!snapshotId) {
     throw new Error(`A snapshot ID is required to ensure data consistency for ${serviceId}'s ${documentType}`);
   }
 
-  return versionRecorder.record({
+  const isFirstRecord = !await versionRecorder.isTracked(serviceId, documentType);
+  let prefix = isFirstRecord ? 'Start tracking' : 'Update';
+  prefix = isRefiltering ? 'Refilter' : prefix;
```
MattiSG, line 47: I don't think `isRefiltering` should take precedence over `isFirstRecord`: if we re-apply filters on a snapshot that results in the first version being recorded, aren't we actually starting to track rather than refiltering? 🙂

src/history/index.js
```diff
@@ -12,24 +12,55 @@ const snapshotRecorder = new Recorder({ path: SNAPSHOTS_PATH, fileExtension: 'ht
...
+
+  let changelog = `${prefix} ${serviceId} ${documentType}`;
+
+  changelog += `\n\nThis version was recorded after filtering snapshot ${config.get('history.publish') ? config.get('history.snapshotsBaseUrl') : ''}${snapshotId}`;
```
MattiSG, line 51:
```suggestion
  const changelog = `${prefix} ${serviceId} ${documentType}

This version was recorded after filtering snapshot ${config.get('history.publish') ? config.get('history.snapshotsBaseUrl') : ''}${snapshotId}`;
```

src/history/index.js
```diff
@@ -12,24 +12,55 @@ const snapshotRecorder = new Recorder({ path: SNAPSHOTS_PATH, fileExtension: 'ht
...
     serviceId,
     documentType,
     content,
-    details: `This version was recorded after filtering snapshot ${config.get('history.publish') ? config.get('history.snapshotsBaseUrl') : ''}${snapshotId}`
+    changelog
   });
+
+  return {
+    ...recordResult,
+    isFirstRecord
```
MattiSG, line 62: > Note: the only reason `isFirstRecord` is exposed to the caller at this time is for logging in `src/index.js`. This should be refactored later.

src/history/index.js
```diff
@@ -38,3 +69,7 @@ export async function publish() {
     versionRecorder.publish()
   ]);
 }
+
+export function getLastSnapshot(serviceId, documentType) {
```
MattiSG, line 73: Let's use consistent adjectives 🙂 

```suggestion
export function getLatestSnapshot(serviceId, documentType) {
```

src/history/index.test.js
```diff
@@ -1,66 +1,316 @@
...
+        before(async () => {
+          const { id: snapshotId, path: snapshotPath } = await recordSnapshot(SERVICE_ID, TYPE, MODIFIED_FILE_CONTENT);
+          id = snapshotId;
+          path = snapshotPath;
+          const commits = await gitSnapshot().log();
+          [commit] = commits;
+        });
+
+        it('does not commit', () => {
+          expect(id).to.be.undefined;
```
MattiSG, line 96: > We are not precisely testing that there was no commit: we are testing that the return value of `recordSnapshot` is `undefined`. I would find it more accurate to test the log result.

src/history/index.test.js
```diff
@@ -1,66 +1,316 @@
...
+    it('returns the id of the commit', async () => {
+      expect(commit.hash).to.includes(id);
+    });
+
+    it('properly saves the commit message', async () => {
+      expect(commit.message).to.equal(FIRST_COMMIT_MESSAGE);
+    });
+
+    it('properly adds snapshot’s id in the commit body', async () => {
+      expect(commit.body).to.includes(SNAPSHOTS_ID);
```
MattiSG, line 144: (repeated in several occurrences)

```suggestion
      expect(commit.body).to.include(SNAPSHOTS_ID);
```

src/history/index.test.js
```diff
@@ -1,66 +1,316 @@
...
+        before(async () => {
+          const { id: versionId, path: versionPath } = await recordVersion(SERVICE_ID, TYPE, MODIFIED_FILE_CONTENT, SNAPSHOTS_ID);
+          id = versionId;
+          path = versionPath;
+          const commits = await gitVersion().log();
+          [commit] = commits;
+        });
+
+        it('does not commit', () => {
+          expect(id).to.be.undefined;
```
MattiSG, line 194: > We are not precisely testing that there was no commit: we are testing that the return value of `recordSnapshot` is `undefined`. I would find it more accurate to test the log result.

src/history/index.test.js
```diff
@@ -1,66 +1,316 @@
...
+    let path;
+    let isFirstRecord;
+
+    before(async () => {
+      const { id: versionId, path: versionPath, isFirstRecord: isFirstVersionRecord } = await recordRefilter(SERVICE_ID, TYPE, FILE_CONTENT, SNAPSHOTS_ID);
+      id = versionId;
+      path = versionPath;
+      isFirstRecord = isFirstVersionRecord;
+      const commits = await gitVersion().log();
+      [commit] = commits;
```
MattiSG, line 229: > Couldn't this repeated assignment, as well as the variable declarations, be factorized in the parent `describe`? 🙂

src/history/recorder.test.js
```diff
@@ -58,66 +56,157 @@ describe('Recorder', () => {
...
+    after(resetGitRepository);
+
+    it('returns all commits related to the given file path', async () => {
+      const filePath = subject.getPathFor(SERVICE_PROVIDER_ID, TYPE);
+      const commits = await subject._getCommits(filePath);
+      expect(commits).to.have.length(2);
+      expect(commits[0].hash).to.include(secondRecordId);
+      expect(commits[1].hash).to.include(firstRecordId);
+    });
+  });
```
MattiSG, line 181: We should record at least another entry for another document, otherwise `_getCommits` might just as well be returning the entire history and we wouldn't know 😅

src/history/recorder.test.js
```diff
@@ -58,66 +56,157 @@ describe('Recorder', () => {
...
+        documentType: TYPE,
+        content: FILE_CONTENT,
+      });
+      const { id: recordId } = await subject.record({
+        serviceId: SERVICE_PROVIDER_ID,
+        documentType: TYPE,
+        content: `${FILE_CONTENT} (with additional content to trigger a record)`,
+      });
+      lastSnapshotId = recordId;
+      lastestRecord = await subject.getLatestRecord(SERVICE_PROVIDER_ID, TYPE);
```
MattiSG, line 199: Either `latest` (= _plus récent_) or `last` (= _dernier_), but `lastest` doesn't exist 😉 
```suggestion
      latestRecord = await subject.getLatestRecord(SERVICE_PROVIDER_ID, TYPE);
```

test/helper.js
```diff
@@ -15,10 +17,17 @@ export async function resetGitRepository() {
...
+  _gitSnapshot = new Git(SNAPSHOTS_PATH);
+  return Promise.all([ _gitVersion.init(), _gitSnapshot.init() ]);
+}
+
+export function gitVersion() {
+  return _gitVersion;
+}
+
+export function gitSnapshot() {
+  return _gitSnapshot;
```
MattiSG, line 30: > Why not simply `export gitVersion`? 🤔

src/index.js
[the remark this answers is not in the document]
MattiSG: Unless you have identified a use case for it, yes please 🙂

src/index.js
```diff
@@ -74,40 +66,150 @@ export default class CGUs extends events.EventEmitter {
...
-        console.log(message);
-        this.emit(isFirstRecord ? 'documentAdded' : 'versionRecorded', serviceId, type, versionId);
-      } else {
-        console.log(`${logPrefix} No changes after filtering, did not record version.`);
+    let snapshotContent;
+    try {
+      snapshotContent = await fs.readFile(snapshotPath, { encoding: 'utf8' });
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        return console.error(`The snapshot file ${snapshotPath} does not exists.`);
```
MattiSG, line 143: If the _do_ takes the _s_, there is no more left for the verb 🙂 
```suggestion
        return console.error(`The snapshot file ${snapshotPath} does not exist.`);
```

src/history/recorder.js
```diff
@@ -41,34 +28,59 @@ export default class Recorder {
...
+  async getLatestRecord(serviceId, documentType) {
+    const filePath = this.getPathFor(serviceId, documentType);
+    const commits = await this._getCommits(filePath);
+    const result = { id: null, path: filePath };
+
+    if (commits.length) {
+      result.id = commits[0].hash;
+    }
+
+    return result;
```
MattiSG, line 64:
```suggestion
    return {
      id: commits[0] && commits[0].hash,
      path: filePath,
    };
```

src/history/recorder.js
```diff
@@ -41,34 +28,59 @@ export default class Recorder {
...
+
+  async isTracked(serviceId, documentType) {
+    const filePath = this.getPathFor(serviceId, documentType);
+
+    if (!await fileExists(filePath)) {
+      return false;
+    }
+
+    const isNew = await this.git.isNew(filePath);
+    return !isNew;
```
MattiSG, line 79: This method is called on each record. It is immensely more likely that the file is already tracked than it is new, and so that the initial `fileExists` call is useless while still incurring a filesystem access. We could thus cut the cost of this call by two in the immense majority of cases by inverting the tests: can we call `git.isNew` with the expectation that the file exists? 🙂

src/index.test.js
```diff
@@ -61,4 +57,64 @@ describe('CGUs', () => {
       expect(resultingTerms).to.be.equal(SERVICE_B_TOS_VERSION);
     });
   });
+
+  describe('#refilterAndRecord', () => {
+    context('When a service’s filter declaration change', () => {
```
MattiSG, line 62:
```suggestion
    context('When a service’s filter declaration changes', () => {
```

src/index.test.js
```diff
@@ -61,4 +57,64 @@ describe('CGUs', () => {
...
+      let refilterVersionId;
+      let refilterVersionMessageBody;
+      let serviceBVersionId;
+
+      before(async () => {
+        nock('https://www.servicea.example').get('/tos').reply(200, SERVICE_A_TOS_SNAPSHOT);
+        nock('https://www.serviceb.example').get('/tos').reply(200, SERVICE_B_TOS_SNAPSHOT);
+        const app = new CGUs();
+        await app.init();
+        await app.trackChanges();
```
MattiSG, line 74: > Could maybe this initial setup be factorized in the parent `describe`? 🙂

src/index.test.js
```diff
@@ -61,4 +57,64 @@ describe('CGUs', () => {
...
+
+        const [originalSnapshotCommit] = await gitSnapshot().log({ file: SERVICE_A_EXPECTED_SNAPSHOT_FILE_PATH });
+        originalSnapshotId = originalSnapshotCommit.hash;
+
+        const [firstVersionCommit] = await gitVersion().log({ file: SERVICE_A_EXPECTED_VERSION_FILE_PATH });
```
MattiSG, line 79: I thought ESlint was supposed to catch this 🤔 
```suggestion
        const [ firstVersionCommit ] = await gitVersion().log({ file: SERVICE_A_EXPECTED_VERSION_FILE_PATH });
```

src/index.test.js
```diff
@@ -61,4 +57,64 @@ describe('CGUs', () => {
...
+      it('refilters the content and saves the file', async () => {
+        const serviceAContent = await fs.readFile(path.resolve(__dirname, SERVICE_A_EXPECTED_VERSION_FILE_PATH), { encoding: 'utf8' })
+        expect(serviceAContent).to.be.equal('Terms of service\n================');
+      });
+
+      it('generates a new version id', async () => {
+        expect(refilterVersionId).to.not.equal(firstVersionId);
+      });
+
+      it('refers the snapshot id in the message', async () => {
```
MattiSG, line 105:
```suggestion
      it('mentions the snapshot id in the changelog', async () => {
```

src/index.test.js
```diff
@@ -61,4 +57,64 @@ describe('CGUs', () => {
...
+
+      it('does not change other services', async () => {
+        const serviceBVersion = await fs.readFile(path.resolve(__dirname, SERVICE_B_EXPECTED_VERSION_FILE_PATH), { encoding: 'utf8' });
+        expect(serviceBVersion).to.be.equal(SERVICE_B_TOS_VERSION);
+      });
+
+      it('does not generate a new id for other services', async () => {
+        const [serviceBCommit] = await gitVersion().log({ file: SERVICE_B_EXPECTED_VERSION_FILE_PATH });
+        expect(serviceBCommit.hash).to.be.equal(serviceBVersionId);
+      });
```
MattiSG, line 117: > I'm not sure these no-side-effects tests are very relevant.

src/history/index.js
Ndpnt: It is also used for event emission, thus to not notify when it is a first record.

src/history/index.test.js
Ndpnt: I dont think so

src/history/git.js
```diff
@@ -43,6 +43,18 @@ export default class Git {
...
 
+  async log(options) {
+    try {
+      const logSummary = await this.git.log(options);
+      return logSummary.all;
+    } catch (error) {
+      if (!error.message.includes('unknown revision or path not in the working tree')) {
+        throw (error);
+      }
+      return [];
```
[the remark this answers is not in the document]
Ndpnt: Yep, If the file does not exists I want to receive an empty array

test/helper.js
Ndpnt: Because it has to be initialized before, and this is done in a the global `before` of this helper.

Ndpnt:
> More than side-effect checks, shouldn't we test that no `newVersion` events were emitted when refiltering?

Yep, but it is the whole events system that has to be tested. I suggest to do it in another PR.

CONTRIBUTING.md
```diff
@@ -122,3 +122,9 @@ Since this operation fetches documents and could be long, you can also validate
 ```
 npm run validate:schema $service_id
 ```
+
+If your change filters or selectors and want to refilter documents from snapshots and regenerate versions only (`$service_id` is optional):
```
MattiSG, line 126:
```suggestion
If you change filters or selectors and want to refilter documents from snapshots and regenerate versions only:
```

CONTRIBUTING.md
```diff
@@ -122,3 +122,9 @@ Since this operation fetches documents and could be long, you can also validate
...
+
+```
+npm run refilter $service_id
```
MattiSG, line 129:
```suggestion
npm run refilter [$service_id]
```

src/history/git.js
```diff
@@ -45,6 +45,24 @@ export default class Git {
...
+    }
+  }
+
+  async isTracked(filepath) {
+    const result = await this.git.raw('ls-files', this.relativePath(filepath));
```
MattiSG, line 62: > If the file has been `git add`-ed, without having been committed, it will still be returned by `ls-files`. I hope that's expected 🙂

src/history/recorder.js
```diff
@@ -54,14 +54,12 @@ export default class Recorder {
 
   async getLatestRecord(serviceId, documentType) {
     const filePath = this.getPathFor(serviceId, documentType);
-    const commits = await this._getCommits(filePath);
-    const result = { id: null, path: filePath };
+    const [ latestCommit ] = await this._getCommits(filePath);
```
MattiSG, line 57: 👍

src/history/git.js
MattiSG: Solution: use `git ls-files --modified`.

Ndpnt: Do not work for already committed file that are not modified

PR #96, opened 2020-08-12
Start to follow Google Ads (removed GoogleAdMob), Google AdSense and Google Analytics

AdrienFines:
I documented one missing document for multi-pages reason in the wiki (AdImplementationPolicies : https://support.google.com/adsense/answer/1346295?hl=en&ref_topic=1271508).

### DIFF
--- a/services/Google.json
+++ b/services/Google.json
@@ -2,14 +2,15 @@
   "name": "Google",
   "documents": {
     "Privacy Policy": {
-      "fetch": "https://policies.google.com/privacy?hl=en-US",
+      "fetch": "https://policies.google.com/privacy?hl=en-GB",
       "filter": [
         "removeUTMfromUrls"
       ],
-      "select": "div[role=article]"
+      "select": "div[role=article]",
+      "remove": ".KMMDve, .TL2alc, img[src$='.svg']"
     },
     "Terms of Service": {
-      "fetch": "https://policies.google.com/terms?hl=en-US",
+      "fetch": "https://policies.google.com/terms?hl=en-GB",
       "filter": [
         "removeUTMfromUrls"
       ],
--- a/services/GoogleAdSense.json
+++ b/services/GoogleAdSense.json
@@ -1,0 +1,51 @@
+{
+  "name": "Google AdSense",
+  "documents": {
+    "Terms of Service": {
+      "fetch": "https://www.google.com/adsense/tc/2018/UnitedKingdom.html",
+      "select": "body"
+    },
+    "Program Policies": {
+      "fetch": "https://support.google.com/adsense/answer/48182",
+      "select": ".article-container",
+      "remove": ".action-button, .zippy"
+    },
+    "Parent Organization Privacy Policy": {
+      "fetch": "https://policies.google.com/privacy?hl=en-GB",
+      "filter": [
+        "removeUTMfromUrls"
+      ],
+      "select": "div[role=article]",
+      "remove": ".KMMDve, .TL2alc, img[src$='.svg']"
+    },
+    "Brand Guidelines": {
+      "fetch": "https://support.google.com/adsense/answer/6370057",
+      "select": ".article-container",
+      "remove": ".no-bullets"
+    },
+    "Quality Guidelines": {
+      "fetch": "https://support.google.com/webmasters/answer/35769",
+      "select": ".article-container"
+    },
+    "Acceptable Use Policy": {
+      "fetch": "https://support.google.com/adsense/answer/9335564?hl=en&ref_topic=1250104",
+      "select": ".article-container"
+    },
+    "Restricted Use Policy": {
+      "fetch": "https://support.google.com/adsense/answer/9335567?hl=en&ref_topic=1250104",
+      "select": ".article-container"
+    },
+    "Data Controller Agreement": {
+      "fetch": "https://privacy.google.com/businesses/controllerterms/",
+      "select": ".ads-controller"
+    },
+    "Trackers Policy": {
+      "fetch": "https://policies.google.com/technologies/ads",
+      "select": "div[role=article]"
+    },
+    "User Consent Policy": {
+      "fetch": "https://www.google.com/about/company/user-consent-policy/",
+      "select": ".page-section"
+    }
+  }
+}
--- a/services/GoogleAds.json
+++ b/services/GoogleAds.json
@@ -1,9 +1,9 @@
 {
-  "name": "Google Analytics",
+  "name": "Google Ads",
   "documents": {
     "Terms of Service": {
-      "fetch": "https://marketingplatform.google.com/about/analytics/terms/gb/",
-      "select": ".terms-section"
+      "fetch": "https://payments.google.com/payments/apis-secure/get_legal_document?ldi=31472",
+      "select": "body"
     },
     "Parent Organization Privacy Policy": {
       "fetch": "https://policies.google.com/privacy?hl=en-GB",
@@ -12,6 +12,14 @@
       ],
       "select": "div[role=article]",
       "remove": ".KMMDve, .TL2alc, img[src$='.svg']"
+    },
+    "Program Policies": {
+      "fetch": "https://support.google.com/adspolicy/answer/6008942",
+      "select": ".article-container"
+    },
+    "Data Controller Agreement": {
+      "fetch": "https://privacy.google.com/businesses/controllerterms/",
+      "select": ".ads-controller"
     },
     "Data Processor Agreement": {
       "fetch": "https://privacy.google.com/businesses/processorterms/",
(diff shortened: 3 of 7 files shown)


services/GoogleAdSense.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "Google AdSense",
+  "documents": {
+    "Terms of Service": {
+      "fetch": "https://www.google.com/adsense/tc/2018/UnitedKingdom.html",
```
Ndpnt, line 5: Why this url and not this one? https://www.google.com/adsense/new/localized-terms?gsessionid=sCco6eXpRU7UA1V1qemcHUPldG-ggJZCHpsm76Y19tI

services/GoogleAdSense.json
```diff
@@ -0,0 +1,53 @@
...
+      "select": "body"
+    },
+    "Program Policies": {
+      "fetch": "https://support.google.com/adsense/answer/48182",
+      "select": ".article-container",
+      "remove": ".action-button, .zippy"
+    },
+    "Parent Organization Privacy Policy": {
+      "fetch": "https://policies.google.com/privacy?hl=en-US",
```
Ndpnt, line 14: There still are noises on this document, at least SVG images and Table of content.

services/GoogleAdSense.json
```diff
@@ -0,0 +1,53 @@
...
+      "filter": [
+        "removeUTMfromUrls"
+      ],
+      "select": "div[role=article]"
+    },
+    "Brand Guidelines": {
+      "fetch": "https://support.google.com/adsense/answer/6370057",
```
Ndpnt, line 21: As internal links does not work for use, maybe remove the table of content

services/GoogleAdSense.json
```diff
@@ -0,0 +1,53 @@
...
+      "select": ".article-container"
+    },
+    "Quality Guidelines": {
+      "fetch": "https://support.google.com/webmasters/answer/35769",
```
Ndpnt, line 25: @MattiSG In this document the version generates an empty link to the intro video (`[](https://www.youtube.com/watch?v=yFxNda5Z4eE)`). Should we keep this or remove it, what do you think?

services/GoogleAdSense.json
AdrienFines: They are the same document : this url is the static adress for English speaking european countries, since I could'nt retrieve the document from the second url.

services/GoogleAdSense.json
AdrienFines: I copy-pasted the Google document and consequently didn't check it. Don't we want to keep images ? 
Will take care of it in a separate PR to clean Google if you don't mind :)

Ndpnt: We only want to keep images that have a meaning in the terms, not illustrations.

AdrienFines: Finally solved, I didn't open another PR.

services/GoogleAdSense.json
AdrienFines: Solved !

services/GoogleAdSense.json
MattiSG: As discussed: I don't have an answer for that as the provider of _CGUs_. I believe it is up to the contributor to decide if the intro video is meaningful (in which case it might be worth adding a filter to add a textual content to the link), useless (in which case it might be worth adding a filter to remove the link), or if they don't care 🙂

services/GoogleAdSense.json
MattiSG: The UK being out from the EU, I'm not sure it is very safe to rely on the UK version 😕

AdrienFines: I agree, yet when searching for documents of English-speaking countries such as Ireland or even Iceland, I always came across this specific document which seems to be not restrictive to UK.

services/GoogleAdSense.json
AdrienFines: Having quickly watched videos, I don't think they are very meaningful, but I'm not sure if we gain anything by removing them so I'm neutral about both decisions.

services/GoogleAdSense.json
```diff
@@ -0,0 +1,55 @@
...
+      "select": "div[role=article]",
+      "remove": ".KMMDve, .TL2alc, img[src$='.svg']"
+    },
+    "Brand Guidelines": {
+      "fetch": "https://support.google.com/adsense/answer/6370057",
+      "select": ".article-container",
+      "remove": ".no-bullets"
+    },
+    "Quality Guidelines": {
+      "fetch": "https://support.google.com/webmasters/answer/35769",
```
Ndpnt, line 27: Why adding this document to AdSense service? It seems to be global Webmaster Guidelines and independent from AdSense, isn't it?

src/types.json
```diff
@@ -27,6 +27,20 @@
       "object": "products and services usage"
     }
   },
+  "Optional Service Terms": {
```
Ndpnt, line 30: I don't understand the meaning of this document type. Can you explain please? :)

services/GoogleAdSense.json
```diff
@@ -0,0 +1,55 @@
...
+      "select": ".article-container"
+    },
+    "Acceptable Use Policy": {
+      "fetch": "https://support.google.com/adsense/answer/9335564?hl=en&ref_topic=1250104",
+      "select": ".article-container"
+    },
+    "Restricted Use Policy": {
```
Ndpnt, line 34: Shouldn't be `Restriction Policy` instead?

services/GoogleAdSense.json
AdrienFines: I based myself upon the following sentence of Adsense TOS : "the AdSense Program Policies, which include but are not limited to the Content Policies, the **Webmaster Quality Guidelines**, the Ad Implementation Policies, and the EU User Consent policy", which refers to this [page](https://support.google.com/adsense/answer/1348737?hl=en&ref_topic=1261918), referering itself to the global Webmaster Guidelines :)

src/types.json
AdrienFines: This document stands for optional services/subservices potentially useful when using the main service, such as those quoted in AdSense TOS : 
```
16. Service-Specific Terms

If you choose to implement any of the following Services on a Property, you also agree to the additional terms identified below:

AdMob: the AdMob Publisher Guidelines and Policies.

Custom Search Engine: the Custom Search Engine Terms of Service.
```
I wasn't sure of the name though

services/GoogleAdSense.json
AdrienFines: I liked the parallel between "Restricted Use Policy" and "Acceptable Use Policy" the latter meaning what uses are accepted and forbidden, the former meaning what uses are allowed but restricted 
But if you find it unclear I can change that :)

services/GoogleAdSense.json
Ndpnt: okay, 👍

services/GoogleAdSense.json
Ndpnt: Ok I understand

src/types.json
Ndpnt: I think this document type is too non-specific and I fear that it becomes a catch-all type. So I'm in favor of removing this it.

AdrienFines: Do you want me to find another document type or do we put aside those documents for the time being ?

Ndpnt: Put aside those documents for now.

AdrienFines: Done, I rebased onto master :)

services/GoogleAdSense.json
```diff
@@ -0,0 +1,51 @@
...
+      "fetch": "https://support.google.com/adsense/answer/9335567?hl=en&ref_topic=1250104",
+      "select": ".article-container"
+    },
+    "Data Controller Agreement": {
+      "fetch": "https://privacy.google.com/businesses/controllerterms/",
+      "select": ".ads-controller"
+    },
+    "Cookies Policy": {
```
Ndpnt, line 42: Since this morning `Cookies Policy` is renamed to `Trackers Policy`

services/GoogleAds.json
```diff
@@ -0,0 +1,42 @@
...
+    },
+    "Data Controller Agreement": {
+      "fetch": "https://privacy.google.com/businesses/controllerterms/",
+      "select": ".ads-controller"
+    },
+    "Data Processor Agreement": {
+      "fetch": "https://privacy.google.com/businesses/processorterms/",
+      "select": ".ads-controller"
+    },
+    "Cookies Policy": {
```
Ndpnt, line 28: Since this morning `Cookies Policy` is renamed to `Trackers Policy`

AdrienFines: Yes :) I've just refactored a few seconds ago !

services/Google.json
```diff
@@ -2,11 +2,12 @@
...
+      "fetch": "https://policies.google.com/privacy?hl=en-GB",
       "filter": [
         "removeUTMfromUrls"
       ],
-      "select": "div[role=article]"
+      "select": "div[role=article]",
+      "remove": ".KMMDve, .TL2alc, img[src$='.svg']"
     },
     "Terms of Service": {
       "fetch": "https://policies.google.com/terms?hl=en-US",
```
Ndpnt, line 13: `US` --> `GB`?

PR #71, opened 2020-08-03
Tosback rules import

michielbdejong:

PR #100, opened 2020-08-17
Reuse validation code in scripts/validate/validate.js

michielbdejong:


michielbdejong:
Moved these changes out of #71 because it's not certain we want to keep using Mocha in `npm run validate`.

michielbdejong:
This will be needed in the future when we integrate this validation code in https://edit.tosdr.org
But for now we decided to focus on integrating with https://tosback.org using  https://github.com/ambanum/TOSBack-CGUs-bridge and not keep that code in sync with the validation code here.

PR #101, opened 2020-08-17
Allow tracking images as datauri

Ndpnt:
Re-implementation of #41

### DIFF
--- a/services/GooglePlayStore.filters.js
+++ b/services/GooglePlayStore.filters.js
@@ -8,7 +8,7 @@ export async function downloadImages(document, { fetch: baseUrl, select: selecto
 
   return Promise.all(images.map(async ({ src }, index) => {
     if (src.startsWith('data:')) {
-      return Promise.resolve(); // Already a data-URI, skip.
+      return; // Already a data-URI, skip
     }
 
     const imageUrl = new URL(src, baseUrl).href; // Ensure url is absolute
--- a/services/GooglePlayStore.filters.js
+++ b/services/GooglePlayStore.filters.js
@@ -1,0 +1,23 @@
+import DatauriParser from 'datauri/parser.js';
+import fetch from '../src/fetcher/index.js';
+
+const dataUriParser = new DatauriParser();
+
+export async function convertImagesToDataURI(document, { fetch: baseUrl, select: selector }) {
+  const images = Array.from(document.querySelectorAll(`${selector} img`));
+
+  return Promise.all(images.map(async ({ src }, index) => {
+    if (src.startsWith('data:')) {
+      return Promise.resolve(); // Already a data-URI, skip.
+    }
+
+    const imageUrl = new URL(src, baseUrl).href; // Ensure url is absolute
+
+    const blob = await fetch(imageUrl, { returnBlob: true });
+
+    const extension = `.${blob.type.split('/')[1].split('+')[0]}`;
+    const arrayBuffer = await blob.arrayBuffer();
+
+    images[index].src = dataUriParser.format(extension, arrayBuffer).content;
+  }));
+}
--- a/services/GooglePlayStore.filters.js
+++ b/services/GooglePlayStore.filters.js
@@ -1,9 +1,9 @@
-import DatauriParser from 'datauri/parser.js';
+import DataURIParser from 'datauri/parser.js';
 import fetch from '../src/fetcher/index.js';
 
-const dataUriParser = new DatauriParser();
+const dataURI = new DataURIParser();
 
-export async function convertImagesToDataURI(document, { fetch: baseUrl, select: selector }) {
+export async function downloadImages(document, { fetch: baseUrl, select: selector }) {
   const images = Array.from(document.querySelectorAll(`${selector} img`));
 
   return Promise.all(images.map(async ({ src }, index) => {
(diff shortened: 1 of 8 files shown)


services/GooglePlayStore.filters.js
```diff
@@ -0,0 +1,23 @@
+import DatauriParser from 'datauri/parser.js';
```
MattiSG, line 1:
```suggestion
import DataURIParser from 'datauri/parser.js';
```

services/GooglePlayStore.filters.js
```diff
@@ -0,0 +1,23 @@
...
+import fetch from '../src/fetcher/index.js';
+
+const dataUriParser = new DatauriParser();
```
MattiSG, line 4:
```suggestion
const dataURI = new DataURIParser();
```

services/GooglePlayStore.filters.js
```diff
@@ -0,0 +1,23 @@
...
+
+    const imageUrl = new URL(src, baseUrl).href; // Ensure url is absolute
+
+    const blob = await fetch(imageUrl, { returnBlob: true });
+
+    const extension = `.${blob.type.split('/')[1].split('+')[0]}`;
+    const arrayBuffer = await blob.arrayBuffer();
+
+    images[index].src = dataUriParser.format(extension, arrayBuffer).content;
+  }));
```
MattiSG, line 22: No promise is returned 🤔

services/GooglePlayStore.filters.js
```diff
@@ -0,0 +1,23 @@
...
+  return Promise.all(images.map(async ({ src }, index) => {
+    if (src.startsWith('data:')) {
+      return Promise.resolve(); // Already a data-URI, skip.
+    }
+
+    const imageUrl = new URL(src, baseUrl).href; // Ensure url is absolute
+
+    const blob = await fetch(imageUrl, { returnBlob: true });
+
+    const extension = `.${blob.type.split('/')[1].split('+')[0]}`;
```
MattiSG, line 18: > I would feel more comfortable relying on a MIME-type handling library such as [`mime`](https://www.npmjs.com/package/mime) 🙂

services/GooglePlayStore.json
```diff
@@ -11,6 +11,9 @@
     },
     "Brand Guidelines": {
       "fetch": "https://developer.android.com/distribute/marketing-tools/brand-guidelines",
+      "filter": [
+        "convertImagesToDataURI"
```
MattiSG, line 15:
```suggestion
        "storeImagesInFile"
```
or maybe
```suggestion
        "downloadImages"
```

src/fetcher/index.js
```diff
@@ -4,7 +4,7 @@ import HttpsProxyAgent from 'https-proxy-agent';
 
 const LANGUAGE = 'en';
 
-export default async function fetch(url) {
+export default async function fetch(url, { returnBlob = false } = {}) {
```
MattiSG, line 7:
```suggestion
export default async function fetch(url, { asRawData = false } = {}) {
```

src/fetcher/index.test.js
```diff
@@ -10,22 +10,32 @@ const __dirname = path.dirname(new URL(import.meta.url).pathname);
...
   describe('#fetch', () => {
     it('returns the web page content of the given URL', async () => {
       const result = await fetch('https://www.facebook.com/terms.php');
       expect(result).to.be.equal(facebookTermsHTML);
     });
 
+    context('With returns as blob option', () => {
+      it('returns the blob of the given URL', async () => {
+        const result = await fetch('https://www.facebook.com/terms.php', { returnBlob: true });
+        expect(result.constructor.name).to.equal('Blob');
```
MattiSG, line 35:
```suggestion
        expect(result).to.be.an.instanceOf(Blob);
```

src/filter/index.js
```diff
@@ -10,12 +10,26 @@ turndownService.use(turndownPluginGithubFlavouredMarkdown.gfm);
...
+    filter: serviceSpecificFilters = [],
+  } = documentDeclaration;
+
   const { document: webPageDOM } = new JSDOM(content).window;
 
-  serviceSpecificFilters.forEach(filterName => {
-    filterFunctions[filterName](webPageDOM); // filters work in place
-  });
+  for (const filterName of serviceSpecificFilters) {
+    const isAsyncFilter = filterFunctions[filterName].constructor.name === 'AsyncFunction';
```
MattiSG, line 24: > Extract a function `isAsync` that checks if its parameter is an `instanceof` a cached `(async () => {}).constructor` 🙂  🌱

MattiSG (changes requested):
Data URI is generated and stored! 👍
…but the resulting Markdown doesn't show any image 🙁

services/GooglePlayStore.filters.js
Ndpnt: Not necessary. As function is declared `async`, the return value is automatically wrapped in a promise.

src/fetcher/index.test.js
Ndpnt: `Blob` is not exposed in Node.js

services/GooglePlayStore.filters.js
MattiSG: Will be done in #109.

MattiSG (approved):
Tested locally, works!

# :clap: 🚀

PR #105, opened 2020-08-18
Add topsites script

michielbdejong:
This is mainly just an experiment, to see how far we would get with this


michielbdejong:
Not a priority for ambanum currently, I can continue this work in the future, but then it will be sponsored by https://tosdr.org/ instead.

PR #110, opened 2020-08-20
Validate added and modified services on CI

Ndpnt:


scripts/validation/validate.js
```diff
@@ -143,3 +154,11 @@ function assertValid(schema, subject) {
     throw new Error(errorMessage);
   }
 }
+
+async function getModifiedServices() {
+  const git = simpleGit(path.resolve(__dirname, '../..'), { maxConcurrentProcesses: 1 });
+  const modifiedFiles = await (await git.diff([ '--name-only', 'master...' ])).split('\n');
```
MattiSG, line 160: Why use triple-dot notation? From [what I understand](https://stackoverflow.com/questions/462974/what-are-the-differences-between-double-dot-and-triple-dot-in-git-com#24186641), this means we will include in the diff all services that have changed in `master` since the branch we're on has diverged, which is not needed.
```suggestion
  const modifiedFiles = await (await git.diff([ '--name-only', 'master' ])).split('\n');
```

scripts/validation/validate.js
```diff
@@ -143,3 +154,11 @@ function assertValid(schema, subject) {
...
+  const modifiedServicePaths = modifiedFiles.filter(filepath => filepath.startsWith('services/') && !filepath.includes('.filters.js'));
```
MattiSG, line 161: Why not simply rely on the `path` argument to `git diff`? 🙂
Above, add the following parameters so that the diff command is `git diff --name-only master services/*.json`.

scripts/validation/validate.js
```diff
@@ -143,3 +154,11 @@ function assertValid(schema, subject) {
...
+  const modifiedServiceNames = modifiedServicePaths.map(filepath => path.basename(filepath).replace('.json', ''));
```
MattiSG, line 162:
```suggestion
  const modifiedServiceNames = modifiedServicePaths.map(filepath => path.basename(filepath, '.json');
```

scripts/validation/validate.js
```diff
@@ -18,16 +19,26 @@ const rootPath = path.join(__dirname, '../..');
 const MIN_DOC_LENGTH = 100;
 
 let serviceDeclarations;
+let modifiedServices;
 (async () => {
   try {
     serviceDeclarations = await loadServiceDeclarations(path.join(rootPath, config.get('serviceDeclarationsPath')));
+    modifiedServices = await getModifiedServices();
```
MattiSG, line 26: Why do that even when it's not needed? Only compute modified services if it is requested! 🌱

scripts/validation/validate.js
```diff
@@ -18,16 +19,26 @@ const rootPath = path.join(__dirname, '../..');
 const MIN_DOC_LENGTH = 100;
 
 let serviceDeclarations;
+let modifiedServices;
```
MattiSG, line 22: Why declare this here?

MattiSG (changes requested):
I find it needlessly complex to do everything in JS. I'll open a separate PR with a different implementation suggestion 🙂

scripts/validation/validate.js
Ndpnt: From my tests, it's exactly the opposite, that's why I added it :)

scripts/validation/validate.js
```diff
@@ -17,24 +18,41 @@ const { expect } = chai;
 const rootPath = path.join(__dirname, '../..');
 const MIN_DOC_LENGTH = 100;
 
-let serviceDeclarations;
+let args = process.argv.slice(6);
```
MattiSG, line 21: Add a comment to explain the origin of this `6`.

scripts/validation/validate.js
```diff
@@ -17,24 +18,41 @@ const { expect } = chai;
...
+if (args.includes('--schema-only')) {
+  args.splice(args.indexOf('--schema-only'), 1);
+  schemaOnly = true;
+}
+
+let modifiedOnly = false;
+if (args.includes('--modified-only')) {
+  args.splice(args.indexOf('--modified-only'), 1);
+  modifiedOnly = true;
+}
```
MattiSG, line 33: > Simplify, support extension and use `const` for options.

```suggestion
const schemaOnly = args.includes('--schema-only');
const modifiedOnly = args.includes('--modified-only');
let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
```

scripts/validation/validate.js
```diff
@@ -17,24 +18,41 @@ const { expect } = chai;
...
+    if (modifiedOnly) {
+      args = await getModifiedServices();
+      if (!args) {
+        console.log('No services modified');
+        return;
+      }
+    }
 
-      const servicesToValidate = serviceId ? [ serviceId ] : serviceIds;
+    const servicesToValidate = args.length ? args : Object.keys(serviceDeclarations);
```
MattiSG, line 47: Suggestion:

```js
    if (modifiedOnly) {
      servicesToValidate = await getModifiedServices();
    } else if (!servicesToValidate.length) {
      servicesToValidate = Object.keys(serviceDeclarations);
    }
```

scripts/validation/validate.js
```diff
@@ -48,18 +66,15 @@ let serviceDeclarations;
...
                   const { fetch: location } = service.documents[type];
-                  content = await fetch(location);
-
-                  if (content.constructor.name == 'Blob' && content.type == 'application/pdf') {
-                    isPDF = true;
-                    content = Buffer.from(await content.arrayBuffer());
-                  }
+                  const { content: tmpContent, mimeType: tmpMimeType } = await fetch(location);
+                  content = tmpContent;
+                  mimeType = tmpMimeType;
```
MattiSG, line 77: See suggestions from #121.

scripts/validation/validate.js
```diff
@@ -165,3 +176,15 @@ function assertValid(schema, subject) {
     throw new Error(errorMessage);
   }
 }
+
+async function getModifiedServices() {
+  const git = simpleGit(path.resolve(__dirname, '../..'), { maxConcurrentProcesses: 1 });
```
MattiSG, line 181:
```suggestion
  const git = simpleGit(rootPath, { maxConcurrentProcesses: 1 });
```

scripts/validation/validate.js
```diff
@@ -165,3 +176,15 @@ function assertValid(schema, subject) {
...
+  const modifiedFilesPathString = await git.diff([ '--name-only', 'master...', 'services/*.json' ]);
```
MattiSG, line 182: Continued from [previous comment](https://github.com/ambanum/CGUs/pull/110#discussion_r479269055).

Which tests do you run? 🙂 Here are mine:

```
$ git mv services/App\ Store.json services/App_Store.json
$ npm run validate:modified
No services modified
# Apply suggested changes (remove the `...` in tree selection)
$ npm run validate:modified
 Services validation
    App_Store
…
```

```suggestion
  const modifiedFilesPathString = await git.diff([ '--name-only', 'master', 'services/*.json' ]);
```

MattiSG (changes requested):
Not blocking, but when I run `npm run validate <service>`, I get the following warnings from Mocha:

```
Warning: Cannot find any files matching pattern "Apple"
Warning: Cannot find any files matching pattern "App Store"
…
```

I've consolidated suggested changes in branch `refactor-validate`.

scripts/validation/validate.js
Ndpnt: Be careful, `git log` and `git diff` behave differently with `...` and the link you provided is for `git log` but we are using `git diff` here. See https://stackoverflow.com/a/7256391

MattiSG: I provided in https://github.com/ambanum/CGUs/pull/110#discussion_r484880144 the way I tested expected behaviour.

PR #112, opened 2020-08-21
Use events for logs

Ndpnt:
- Modify folder structure to represent which module should be considered outside of the app.
- Remove `refilter` and `schedule` scripts, integrate their feature inside the main `index` script.
- Create logger based on events.
- Improve logs readability.


src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+import filter from './filter/index.js';
+import loadServiceDeclarations from './loader/index.js';
+
+const fs = fsApi.promises;
+const __dirname = path.dirname(new URL(import.meta.url).pathname);
+const SERVICE_DECLARATIONS_PATH = path.resolve(__dirname, '../../', config.get('serviceDeclarationsPath'));
+
+export const AVAILABLE_EVENTS = [
+  'startTrackingChanges',
+  'endTrackingChanges',
```
MattiSG, line 18: > “to end” entails we could have been interrupted, “to finish” means we are done.
```suggestion
  'finishTrackingChanges',
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  async init() {
+    if (!this.initialized) {
+      this._serviceDeclarations = await loadServiceDeclarations(SERVICE_DECLARATIONS_PATH);
+      this.initialized = Promise.resolve();
+    }
+
+    return this.initialized;
+  }
```
MattiSG, line 46:
```suggestion
  async init() {
    if (!this.initialized) {
      this.initialized = loadServiceDeclarations(SERVICE_DECLARATIONS_PATH);
    }
    return this.initialized;
  }
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+
+  attach(listener) {
+    AVAILABLE_EVENTS.forEach(event => {
+      const handlerName = `on${event[0].toUpperCase()}${event.substr(1)}`;
+
+      if (Object.prototype.hasOwnProperty.call(listener, handlerName)
```
MattiSG, line 52: > Why do we want to gracefully support the cases where `listener` is undefined, a primitive or has an empty prototype?
```suggestion
      if (listener.hasOwnProperty(handlerName)
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+      || Object.prototype.hasOwnProperty.call(listener.constructor.prototype, handlerName)) {
```
MattiSG, line 53: I don't understand why we manually check `hasOwnProperty` on the parameter and its constructor instead of relying on the prototype chain navigation 🤔

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+      this.emit('startTrackingChanges');
+
+      const services = serviceToTrack ? { [serviceToTrack]: this._serviceDeclarations[serviceToTrack] } : this._serviceDeclarations;
+
+      const documentTrackingPromises = [];
+
+      Object.keys(services).forEach(serviceId => {
+        const { documents, name: serviceName } = this._serviceDeclarations[serviceId];
+        Object.keys(documents).forEach(type => {
+          documentTrackingPromises.push(this.trackDocumentChanges({
```
MattiSG, line 70:
```suggestion
      const documentTrackingPromises = Object.keys(services).flatMap(serviceId => {
        const { documents, name: serviceName } = this._serviceDeclarations[serviceId];
        return Object.keys(documents).map(type => {
          return this.trackDocumentChanges({
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+            serviceName,
+            document: {
+              type,
+              ...documents[type]
+            }
+          }));
+        });
+      });
+
+      await Promise.all(documentTrackingPromises);
```
MattiSG, line 81: We probably don't want to publish as soon as one document fails 🙊

> [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) rejects with the reason of the first promise that rejects.

```suggestion
      await Promise.allSettled(documentTrackingPromises);
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+
+      await this.publish();
+      this.emit('endTrackingChanges');
+    } catch (error) {
+      this.emit('applicationError', error);
```
MattiSG, line 86: I'm not sure it is appropriate to wrap the entire call in a try / catch, which questions the `applicationError` event type. Let's chat 🙂 🗣

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+          serviceName,
+          document: {
+            type,
+            ...documents[type]
+          }
+        }));
+      });
+    });
+
+    await Promise.all(refilterAndRecordDocumentPromises);
```
MattiSG, line 146: Unless we abort…
```suggestion
    await Promise.allSettled(refilterAndRecordDocumentPromises);
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+        throw new Error(`The snapshot file ${snapshotPath} does not exist.`);
+      }
+    }
+
+    return this.recordRefilter({
+      snapshotContent,
+      snapshotId,
+      serviceId,
+      documentDeclaration,
+      logPrefix,
```
MattiSG, line 173: Unused variable.
```suggestion
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+      });
+    });
+
+    await Promise.all(refilterAndRecordDocumentPromises);
+    this.emit('endRefiltering');
+  }
+
+  async refilterAndRecordDocument({ serviceId, serviceName, document: documentDeclaration }) {
+    const { type } = documentDeclaration;
+    const logPrefix = `[${serviceName}-${type}]`;
```
MattiSG, line 152: Unused variable.
```suggestion
```

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+      return;
+    }
+
+    let snapshotContent;
+    try {
+      snapshotContent = await fs.readFile(snapshotPath, { encoding: 'utf8' });
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        throw new Error(`The snapshot file ${snapshotPath} does not exist.`);
+      }
```
MattiSG, line 165: Do we really want not to fail in all other cases where `readFile` threw?

src/app/index.js
```diff
@@ -0,0 +1,234 @@
...
+  'firstSnapshotRecorded',
+  'noSnapshotChanges',
+  'versionRecorded',
+  'firstVersionRecorded',
+  'noVersionChanges',
+  'changesPublished',
+  'applicationError',
+  'documentUpdateError',
+  'documentFetchError',
+  'recordRefilterError',
```
MattiSG, line 31: I feel we should unify times and names. Let's discuss this 🙂 🗣

src/app/index.test.js
```diff
@@ -0,0 +1,682 @@
...
+        after(() => {
+          resetSpiesHistory();
+          return resetGitRepository();
+        });
+
+        it('emits "onFirstSnapshotRecorded" event', async () => {
+          expect(spies.onFirstSnapshotRecorded).to.have.been.calledWith(SERVICE_A_ID, SERVICE_A_TYPE);
+        });
+
+        it('emits no other events', async () => {
```
MattiSG, line 161: Why is this `async`? 🤔

src/app/index.test.js
```diff
@@ -0,0 +1,682 @@
...
+          before(async () => {
+            await app.recordSnapshot({ content: 'document content', serviceId: SERVICE_A_ID, type: SERVICE_A_TYPE });
+            resetSpiesHistory();
+            await app.recordSnapshot({ content: 'document content modified', serviceId: SERVICE_A_ID, type: SERVICE_A_TYPE });
+          });
+
+          after(() => {
+            resetSpiesHistory();
+            return resetGitRepository();
+          });
```
MattiSG, line 190: Couldn't this be factorized?

src/app/index.test.js
```diff
@@ -0,0 +1,682 @@
...
+          expect(spies.onSnapshotRecorded).to.have.not.been.called;
+          expect(spies.onNoSnapshotChanges).to.have.not.been.called;
+          expect(spies.onFirstVersionRecorded).to.have.not.been.called;
+          expect(spies.onVersionRecorded).to.have.not.been.called;
+          expect(spies.onNoVersionChanges).to.have.not.been.called;
+          expect(spies.onChangesPublished).to.have.not.been.called;
+          expect(spies.onApplicationError).to.have.not.been.called;
+          expect(spies.onRecordRefilterError).to.have.not.been.called;
+          expect(spies.onDocumentUpdateError).to.have.not.been.called;
+          expect(spies.onDocumentFetchError).to.have.not.been.called;
```
MattiSG, line 484: All these calls look very unmaintainable to me: every time we'll update an event name, create or remove an event, we'll have to update hundreds of lines 😕

I am not 100% sure of the added value of testing that all other events have not been called, but if you want to test it, what about creating a helper test function that takes as a parameter the name of the only event that is expected to have been called, checks that is has been called, and iterates on all the others to check that they haven't? 🙂

src/app/loader/index.test.js
```diff
@@ -22,7 +22,7 @@ const expected = {
   service_B: {
     name: 'Service B',
     documents: {
-      'Terms of Service': {
+      'Privacy Policy': {
```
MattiSG, line 25: > I'm quite curious why this changed 😮

src/app/loader/index.test.js
```diff
@@ -22,7 +22,7 @@ const expected = {
...
         fetch: 'https://www.serviceb.example/tos',
```
MattiSG, line 26: This is inconsistent with the document type above. I know it's only a test, but it'll make debugging much harder if it ever comes to logging what happens 😉

src/index.js
```diff
@@ -1,217 +1,41 @@
...
+import scheduler from 'node-schedule';
 
-import config from 'config';
+import CGUs from './app/index.js';
+import * as logger from './logger/index.js';
+import Notifier from './notifier/index.js';
 
-import consoleStamp from 'console-stamp';
+(async () => {
+  try {
```
MattiSG, line 8: Could we please prefer to wrap with `try` only the parts where we actually expect an exception? This way we don't `catch` stuff that we don't expect 🙂

src/index.js
```diff
@@ -1,217 +1,41 @@
...
+    const serviceId = process.argv.slice(process.argv.indexOf('--serviceId'))[1];
+    const refilterOnly = process.argv.indexOf('--refilter-only') != -1;
+    const schedule = process.argv.indexOf('--schedule') != -1;
```
MattiSG, line 11: I see no reason for this to be `async`.

src/index.js
```diff
@@ -1,217 +1,41 @@
...
 
-import * as history from './history/index.js';
-import fetch from './fetcher/index.js';
-import filter from './filter/index.js';
-import loadServiceDeclarations from './loader/index.js';
+    const app = new CGUs();
+    await app.init();
+    await app.attach(logger);
```
MattiSG, line 15: Why is attaching an event listener asynchronous? 🤔

src/index.js
```diff
@@ -1,217 +1,41 @@
...
-      const message = isFirstRecord
-        ? `${logPrefix} First version recorded in ${documentPath} with id ${versionId}.`
-        : `${logPrefix} Recorded version in ${documentPath} with id ${versionId}.`;
-      console.log(message);
-      this.emit(isFirstRecord ? 'documentAdded' : 'versionRecorded', serviceId, type, versionId);
+      console.log('The scheduler is running…');
+      console.log('Documents will be tracked at minute 30 past every hour.');
+      scheduler.scheduleJob(rule, () => {
+        app.trackChanges(serviceId);
+      });
```
MattiSG, line 34:
```suggestion
      scheduler.scheduleJob(rule, app.trackChanges.bind(app, serviceId));
```

src/index.js
```diff
@@ -1,217 +1,41 @@
...
     } else {
-      console.log(`${logPrefix} No changes after filtering, did not record version.`);
+      await app.trackChanges(serviceId);
```
MattiSG, line 36: > Invert the condition and transform this into an early exit to improve readability 🙂

src/index.js
```diff
@@ -1,217 +1,41 @@
...
     }
+  } catch (error) {
+    console.error(error);
```
MattiSG, line 39: Why don't we want to raise an exception and exit with a proper failure exit code?

src/logger/index.js
```diff
@@ -0,0 +1,84 @@
...
+  }));
+
+const logger = winston.createLogger({
+  format: alignedWithColorsAndTime,
+  transports: [
+    new winston.transports.Console({
+      handleExceptions: true
+    }),
+  ],
+  exitOnError: false
```
MattiSG, line 23: Why?
If set to `false`, then why `catch` everything at top level?
Let's discuss this🗣

test/services/service_B.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Service B",
   "documents": {
-    "Terms of Service": {
+    "Privacy Policy": {
       "fetch": "https://www.serviceb.example/tos",
```
MattiSG, line 5: Inconsistent as above 😉

src/index.js
```diff
@@ -1,217 +1,41 @@
...
 
-import config from 'config';
+import CGUs from './app/index.js';
+import * as logger from './logger/index.js';
+import Notifier from './notifier/index.js';
 
-import consoleStamp from 'console-stamp';
+(async () => {
+  try {
+    const serviceId = process.argv.slice(process.argv.indexOf('--serviceId'))[1];
```
MattiSG, line 9: > This means we support only one service at a time.

MattiSG (changes requested):
Quite a few points raise the need for a discussion 🙂

src/app/index.js
MattiSG: Not part of this PR, just moved code.

src/app/index.js
MattiSG: [live discussion] We'll use `[]` property existence check instead.

src/app/index.js
Ndpnt: Yep, we have to do the same for refiltering

src/app/index.test.js
MattiSG: “Because it's annoying to change only when needed”

src/app/loader/index.test.js
MattiSG: “It made tests more readable” — @Ndpnt

src/index.js
Ndpnt: Move before init

src/index.js
MattiSG: Out of scope.

src/app/index.js
```diff
@@ -9,27 +8,20 @@ import fetch from './fetcher/index.js';
...
+export const AVAILABLE_EVENTS = {
+  snapshotRecorded: 'snapshotRecorded',
+  firstSnapshotRecorded: 'firstSnapshotRecorded',
+  snapshotNotChanged: 'snapshotNotChanged',
+  versionRecorded: 'versionRecorded',
+  firstVersionRecorded: 'firstVersionRecorded',
+  versionNotChanged: 'versionNotChanged',
+  recordsPublished: 'recordsPublished',
+  inaccessibleContentError: 'inaccessibleContentError',
+  refilteringError: 'refilteringError',
```
MattiSG, line 23: I'm not sure I understand the point of this mapping 😅

src/logger/index.js
```diff
@@ -59,26 +63,30 @@ export function onVersionRecorded(serviceId, type, versionId) {
...
-export function onDocumentUpdateError(serviceId, type, error) {
-  logger.error({ message: `Could not update document: ${error}`, serviceId, type });
+export function error(options) {
+  logger.error(options);
 }
 
-export function onDocumentFetchError(serviceId, type, error) {
-  logger.error({ message: `Could not fetch document: ${error}`, serviceId, type });
+export function warn(options) {
+  logger.warn(options);
```
MattiSG: Couldn't this work?
```js
export {
  logger.info as info,
  logger.warn as warn,
  logger.error as error,
}
```

src/logger/index.js
```diff
@@ -13,31 +17,31 @@ const alignedWithColorsAndTime = combine(colorize(),
...
 
-export function onEndTrackingChanges() {
-  logger.info({ message: 'Tracking changes done.' });
-}
+if (config.get('logger.sendMailOnError')) {
+  const mailErrorTransport = new winston.transports.Mail({
+    level: 'error',
+    to: config.get('logger.sendMailOnError.to'),
+    from: config.get('logger.sendMailOnError.from'),
+    subject: 'CGUs - Error Report',
```
MattiSG, line 29: More common in email subjects.

```suggestion
    subject: '[CGUs] Error Report',
```

src/logger/index.js
```diff
@@ -13,31 +17,31 @@ const alignedWithColorsAndTime = combine(colorize(),
...
+    host: process.env.SMTP_HOST,
+    username: process.env.SMTP_USERNAME,
+    password: process.env.SMTP_PASSWORD,
+    ssl: true,
+    formatter: args => args[Object.getOwnPropertySymbols(args)[1]]
```
MattiSG, line 34: Beg your pardon? 😳

src/index.js
```diff
@@ -4,38 +4,39 @@ import CGUs from './app/index.js';
 import * as logger from './logger/index.js';
 import Notifier from './notifier/index.js';
 
+const serviceId = process.argv.slice(process.argv.indexOf('--serviceId'))[1];
+const refilterOnly = process.argv.indexOf('--refilter-only') != -1;
+const schedule = process.argv.indexOf('--schedule') != -1;
```
MattiSG, line 9:
```suggestion
const refilterOnly = process.argv.includes('--refilter-only');
const schedule = process.argv.includes('--schedule');
```

MattiSG, line 9:
```suggestion
const refilterOnly = process.argv.includes('--refilter-only');
const schedule = process.argv.includes('--schedule');
```

src/index.js
```diff
@@ -4,38 +4,39 @@ import CGUs from './app/index.js';
 import * as logger from './logger/index.js';
 import Notifier from './notifier/index.js';
 
+const serviceId = process.argv.slice(process.argv.indexOf('--serviceId'))[1];
```
MattiSG, line 7: See the way this could be implemented in #122 (c8437bbf9ba77226f6582540e01c81414f1a923a). In any case, unify with arguments parsing code in #110 or #122 🙂

MattiSG (changes requested):
I did not review the entirety of `src/app/index.js` and `src/app/index.test.js` because the amount of noise due to upstream merge is massive, and based on the provided feedback it is likely that upcoming changes —if accepted— will be large too.

src/logger/index.js
Ndpnt: Nope…

ops/roles/cgus/files/.env
```diff
@@ -1,10 +1,18 @@
 $ANSIBLE_VAULT;1.1;AES256
-66313936623434663337386433636138346433373062313134353564393961663565616638303438
```
Ndpnt, line 2: Add example in `.env.exmaple` for added variables

MattiSG (approved):
## Tested locally: this works 👍 

### Usage improvements

- When referencing a service that does not exist (potential typo), the UI is not very clear:
Capture d’écran 2020-09-10 à 14 51 12

- I don't need the stack trace when the selector does not match.
Capture d’écran 2020-09-10 à 14 50 51

- I am not sure that contentNotAcessible should really be an error, as they are somehow expected states.

> - I would appreciate the capability to (at least visually) separate “high level” events from more specific ones: I'd like to have information on each document on a `debug` or equivalent level, to make it it easy to spot big steps (or even silence them altogether). This would also remove the need for an artificial newline after steps like “refiltering done”.
Capture d’écran 2020-09-10 à 14 52 42

src/logger/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+dotenv.config();
+const { combine, timestamp, printf, colorize } = winston.format;
+
+const alignedWithColorsAndTime = combine(colorize(),
+  timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
+  printf(({ level, message, timestamp, serviceId, type }) => {
+    let prefix = '';
+    if (serviceId && type) {
+      prefix = `${serviceId} — ${type}`;
```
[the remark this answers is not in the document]
Ndpnt: It should never happened

src/logger/index.js
```diff
@@ -0,0 +1,92 @@
...
+  logger.info({ message: `Recorded first version with id ${versionId}.`, serviceId, type });
+}
+
+export function onVersionRecorded(serviceId, type, versionId) {
+  logger.info({ message: `Recorded version with id ${versionId}.`, serviceId, type });
+}
+
+export function onVersionNotChanged(serviceId, type) {
+  logger.info({ message: 'No changes after filtering, did not record version.', serviceId, type });
+}
```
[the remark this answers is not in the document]
Ndpnt: I'm not sure to understand, it seems already aligned for me. 
Moreover `serviceId` and `type` are passed time so I prefer having the message, which is different from one to another, as the first option.

src/index.js
```diff
@@ -1,239 +1,43 @@
...
-        logPrefix,
-      });
-    } catch (error) {
-      console.error(`${logPrefix} Error:`, error);
-      this.emit('documentUpdateError', serviceId, type, error);
-    }
+  if (!schedule) {
+    logger.info('Start tracking changes…');
+    await app.trackChanges(services);
+    return logger.info('Tracking changes done.');
```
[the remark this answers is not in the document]
Ndpnt: Not really, because if we pass no arguments it will display 0 whereas we will treat all declared documents

Issue #106, opened 2020-08-18
Document common practice for choosing a service name


michielbdejong:
In order to go from 50 to 5,000 services, we need to have and understand our system for picking a canonical service name.

If you look at the filenames in https://github.com/ambanum/CGUs/tree/master/services most of the names are spelt and capitalized:
1. the same way as on wikipedia
2. Pascal case, but abbreviations in all-caps
3. preceded by parent service if applicable
4. just the commonly known name, no suffixes like ".com" or "Inc."
5. with spaces and dots removed

Examples:
ask.fm (1) Ask.fm (2) Ask.FM (5) AskFM
https://www.apple.com/ios/app-store/ (1) App Store (3) Apple App Store (5) AppleAppStore
Facebook Payments Inc. (4) Facebook Payments (5) FacebookPayments
Foursquare City Guide (4) Foursquare
reddit.com (1) reddit (2) Reddit

Exceptions (fix these?):
* [ ] GooglePlayStore instead of GooglePlay
* [ ] LastFm instead of LastFM (inconsistent because we also have AskFM)
* [ ] deviantART instead of DeviantArt
* [ ] hi5 instead of Hi5

michielbdejong:
We should also avoid special characters, e.g.
EUC Lillebælt -> EUC Lillebaelt
fógraí.ie -> Fograi
Pepper&Carrot -> PepperCarrot
Yahoo! -> Yahoo
RTÉ -> RTE
Historielærer.dk -> Historielaerer

MattiSG:
Totally right, thanks for putting this up @michielbdejong!

Here is an RFC, what do you think of it? 🙂

### We need

1. A name that reflects the common name used by the provider itself, to be exposed in a GUI. This name is currently exposed as the `name` property in the service declaration.
2. An ID of sorts that can be represented in the filesystem. This ID is currently exposed as the filename of the service declaration, without the `.json` extension.

### Use cases

- The service name is presented to end users. It should reflect as closely as possible the official service name, so that it can be identified easily.
- The ID is used internally and exposed for analysis. It should be easy to handle with scripts and other tools.

### Constraints for the ID

As long as this ID is stored in the filesystem:

- No `/` for UNIX.
- No `\` for Windows.
- No `:` for APFS and HFS.
- No case-sensitive duplicates to support case-insensitive filesystems.
- No more than 255 characters to support transfer over [FAT32](https://en.wikipedia.org/wiki/File_Allocation_Table#FAT32).

[...]

PR #113, opened 2020-08-28
An example rule to add

michielbdejong:
In tosback2 we have for instance "Alexa Terms of Service" under amazon.com.xml.
I translated those to "Terms of Service" of "AmazonAlexa.json".

Here are a few examples of how my script creates them now. Note that they contain the selectors translated from xpath, but when creating the imported versions, i'm using 'body' as the selector.

### DIFF
--- a/services/AmazonAlexa.json
+++ b/services/AmazonAlexa.json
@@ -1,0 +1,10 @@
+{
+  "name": "AmazonAlexa",
+  "importedFrom": "https://github.com/tosdr/tosback2/blob/b10b738ce7f5fc18b90c513c4db360f728348058/rules/amazon.com.xml",
+  "documents": {
+    "Terms of Service": {
+      "fetch": "https://www.amazon.com/gp/help/customer/display.html?nodeId=201809740",
+      "select": "div.help-content"
+    }
+  }
+}


Ndpnt:
Great!
Last week, with @MattiSG , we wrote a [first version of what could be documentation for common practice](https://github.com/ambanum/CGUs/issues/106#issuecomment-681870747) for choosing a service name. If you agree with it and you think it covers all the cases you might encounter, can you use it to name both the file and the service in the declaration?

MattiSG:
This looks good, taking into accounts @Ndpnt's comment on naming 👍

I see only Alexa for now while you use the plural in comments, is it expected or is there a glitch? 🙂

michielbdejong:
Ah yes sorry, instead of " Here are a few examples" I should have said "Here is one example" :)

MattiSG:
No worries, just wanted to make sure I had not missed something  :)

services/AmazonAlexa.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "AmazonAlexa",
```
MattiSG, line 2: According to #106:

```suggestion
  "name": "Alexa",
```

Or, if ambiguous:

```suggestion
  "name": "Amazon Alexa",
```

MattiSG (changes requested):

PR #115, opened 2020-08-31
Improve filters documentation

Ndpnt:
Improve filters documentation to reflect changes from https://github.com/ambanum/CGUs/pull/101

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,14 +64,20 @@ Some documents will need additional filtering beyond simple element deletion and
 
 These filters are declared as JavaScript functions that modify the downloaded web page through the [DOM API](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model).
 
-Filters take the document as parameter and are:
+Filters take the document and the document declaration as parameters and are:
 
 - **in-place**: they modify the document structure and content directly;
 - **idempotent**: they should return the same document structure and content even if run repeatedly on their own result.
 
 Filters are loaded automatically from files named after the service they operate on. For example, filters for the Meetup service, which is declared in `services/Meetup.json`, are loaded from `services/Meetup.filters.js`.
 
-Each filter is exposed as a named function export, which could be `async`, that takes a `document` parameter and behaves like the `document` object in a browser DOM.
+The generic function signature for a filter is:
+
+```js
+export [async] function filterName(document [, documentDeclaration])
+```
+
+Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM. These functions can be `async`, but they will still run sequentially.
 
 > It is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
 
@@ -87,15 +93,9 @@ export function removeImages(document, { select: selector }) {
 
 ```
 
-So the generic function signature is:
-
-```js
-export [async] function filterName(document [, documentDeclaration])
-```
-
-Learn more about [usual noise](https://github.com/ambanum/CGUs/wiki/Usual-noise).
-
 You can find examples of filters in [`/services/*.filters.json`](./services) files.
+
+You can also learn more about [usual noise](https://github.com/ambanum/CGUs/wiki/Usual-noise).
 
 ### Selectors
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,9 +79,9 @@ export [async] function filterName(document [, documentDeclaration])
 
 Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM. These functions can be `async`, but they will still run sequentially.
 
-> It is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
+> The `document` parameter is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
 
 Filters are loaded automatically from files named after the service they operate on. For example, filters for the Meetup service, which is declared in `services/Meetup.json`, are loaded from `services/Meetup.filters.js`.
 
-Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM.
+Each filter is exposed as a named function export, which could be `async`, that takes a `document` parameter and behaves like the `document` object in a browser DOM.
```
MattiSG, line 74: If I understand correctly, I feel this is a useful precision:
```suggestion
Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM. These functions can be `async`, but they will still run sequentially.
```

CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
...
 
 > It is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
 
+There is also an optional parameter containing the whole document declaration. It can be useful if you need to access the defined document url or selector inside your filter.
+
+For example, you can scope your the selection in your filter by the `selector` defined in the document declaration:
```
MattiSG, line 80:
```suggestion
For example, you can scope the selection in your filter with the `selector` defined in the document declaration:
```

CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
...
+
+```js
+export function removeImages(document, { select: selector }) {
+  const images = document.querySelectorAll(`${selector} img`);
+  images.forEach(el => el.remove());
```
MattiSG, line 85: > This documentation is very important as it sets the standard for what we expect 🙂
> Do we want to encourage this or `document.querySelectorAll(`${selector} img`).forEach(el => el.remove());`?
> 
> It's a real question, I'm not sure I have an opinion on this.

CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
...
+}
+
+```
+
+So the generic function signature is:
+
+```js
+export [async] function filterName(document [, documentDeclaration])
+```
```
MattiSG, line 94: Maybe move this to the top of the description, right before “each filter is exposed as a named function export”? This way, if the contributor is experimented she doesn't feel obligated to read the entire following paragraphs  🙂

CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
...
+
 Learn more about [usual noise](https://github.com/ambanum/CGUs/wiki/Usual-noise).
```
MattiSG, line 96: This feels out of place now, maybe move it up or after the examples paragraph?

CONTRIBUTING.md
```diff
@@ -71,12 +71,32 @@ Filters take the document as parameter and are:
...
 
+You can find example of filters in [`/services/*.filters.json`](./services) files.
```
MattiSG, line 98:
```suggestion
You can find examples of filters in [`/services/*.filters.json`](./services) files.
```

MattiSG (changes requested):

MattiSG (approved):
Nothing blocking anymore, but please make sure to handle the open (non-quoted) comments before merging 🙂

Congrats for thinking about updating this documentation and for the clarity of the explanations! 👍

CONTRIBUTING.md
Ndpnt: I do not have a strong opinion on this neither, so I suppose it is no so important :).

CONTRIBUTING.md
```diff
@@ -64,18 +64,38 @@ Some documents will need additional filtering beyond simple element deletion and
...
-Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM.
+The generic function signature for a filter is:
+
+```js
+export [async] function filterName(document [, documentDeclaration])
+```
+
+Each filter is exposed as a named function export that takes a `document` parameter and behaves like the `document` object in a browser DOM. These functions can be `async`, but they will still run sequentially.
 
 > It is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
```
MattiSG, line 82:
```suggestion
> The `document` parameter is actually a [JSDOM](https://github.com/jsdom/jsdom) document instance.
```

CONTRIBUTING.md
```diff
@@ -64,18 +64,38 @@ Some documents will need additional filtering beyond simple element deletion and
...
 
-Learn more about [usual noise](https://github.com/ambanum/CGUs/wiki/Usual-noise).
+There is also an optional parameter containing the whole document declaration. It can be useful if you need to access the defined document url or selector inside your filter.
```
MattiSG, line 84: > All parameters are “optional” in the sense that they are provided to the function, you may or may not use them.

```suggestion
The second parameter contains the whole document declaration. It can be useful if you need to access the defined document URL or selector inside your filter.
```

MattiSG (approved):
👏

PR #121, opened 2020-09-07
Fix validate command

Ndpnt:


scripts/validation/validate.js
```diff
@@ -48,18 +48,15 @@ let serviceDeclarations;
...
                   const { fetch: location } = service.documents[type];
-                  content = await fetch(location);
-
-                  if (content.constructor.name == 'Blob' && content.type == 'application/pdf') {
-                    isPDF = true;
-                    content = Buffer.from(await content.arrayBuffer());
-                  }
+                  const { content: tmpContent, mimeType: tmpMimeType } = await fetch(location);
+                  content = tmpContent;
+                  mimeType = tmpMimeType;
```
MattiSG, line 59:
```suggestion
                  const document = await fetch(location);
                  content = document.content;
                  mimeType = document.mimeType;
```

MattiSG (changes requested):
Tested locally, works!

MattiSG (approved):
Tested locally, works!

PR #122, opened 2020-09-08
Refactor validate

MattiSG:
Contains most suggestions from #110 review, consolidated.

### DIFF
--- a/scripts/validation/validate.js
+++ b/scripts/validation/validate.js
@@ -20,31 +20,19 @@ const MIN_DOC_LENGTH = 100;
 
 let args = process.argv.slice(6);
 
-let schemaOnly = false;
-if (args.includes('--schema-only')) {
-  args.splice(args.indexOf('--schema-only'), 1);
-  schemaOnly = true;
-}
-
-let modifiedOnly = false;
-if (args.includes('--modified-only')) {
-  args.splice(args.indexOf('--modified-only'), 1);
-  modifiedOnly = true;
-}
+const schemaOnly = args.includes('--schema-only');
+const modifiedOnly = args.includes('--modified-only');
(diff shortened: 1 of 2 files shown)


scripts/validation/validate.js
```diff
@@ -178,13 +166,9 @@ function assertValid(schema, subject) {
...
 
 async function getModifiedServices() {
-  const git = simpleGit(path.resolve(__dirname, '../..'), { maxConcurrentProcesses: 1 });
-  const modifiedFilesPathString = await git.diff([ '--name-only', 'master...', 'services/*.json' ]);
-
-  if (!modifiedFilesPathString) {
-    return;
-  }
+  const git = simpleGit(rootPath, { maxConcurrentProcesses: 1 });
+  const modifiedFilesPathString = await git.diff([ '--name-only', 'master', 'services/*.json' ]);
```
Ndpnt, line 170: From my tests, if your remove `...` it will try to validate services which are modified on master but not necessary present on the current branch and it will fails in these cases.
So it is necessary. See https://stackoverflow.com/a/7256391

scripts/validation/validate.js
```diff
@@ -20,32 +20,20 @@ const MIN_DOC_LENGTH = 100;
...
-}
-
-let modifiedOnly = false;
-if (args.includes('--modified-only')) {
-  args.splice(args.indexOf('--modified-only'), 1);
-  modifiedOnly = true;
-}
+const schemaOnly = args.includes('--schema-only');
+const modifiedOnly = args.includes('--modified-only');
+let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
```
Ndpnt, line 25: 👍

Ndpnt (changes requested):

Ndpnt (approved):

PR #124, opened 2020-09-10
Improve refiltering performance

Ndpnt:
Improve perf x3
Allow refiltering process to log output continuously

### DIFF
--- a/src/app/history/git.js
+++ b/src/app/history/git.js
@@ -17,10 +17,6 @@ export default class Git {
     return this.git.add(this.relativePath(filepath));
   }
 
-  async status() {
-    return this.git.status();
-  }
-
   async commit(filepath, message) {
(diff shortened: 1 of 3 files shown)


src/app/history/git.js
```diff
@@ -59,16 +49,30 @@ export default class Git {
...
+    const [ latestCommit ] = await this.log([ '-n', '1', '--stat=4096', glob ]);
+
+    if (!latestCommit) {
+      return {};
+    }
+
+    const filePaths = latestCommit.diff.files.map(file => file.file);
+
+    if (filePaths.length > 1) {
+      throw new Error(`Only one document should have been recorded in ${latestCommit.hash}, but all these documens have been recorded: ${filePaths}`);
```
MattiSG, line 67: I thought we had discovered this can happen without being a problem, for example when we manually rename files 🤔

MattiSG (changes requested):
Wow, much better UX when refiltering!

MattiSG (approved):
I'd still like the function name to be improved, but this works and the typos have been fixed 🙂

PR #125, opened 2020-09-10
Add stats npm command

Ndpnt:
Really simple stats command to display services and documents numbers.

Capture d’écran 2020-09-10 à 10 40 17


MattiSG (approved):
Thanks! 👍

PR #127, opened 2020-09-10
Add stats npm command

Ndpnt:


MattiSG (approved):
GTM, but I'd also be happy if we just always log this at the _beginning_ of `npm start`, which would avoid that entire script 🙂

MattiSG:
Just saw this is already provided since #112 as the first line when running `npm start`:

```
2020-09-10 17:23:46 info                                                          Refiltering 136 documents from 51 services…
```

I'm not sure this is worth adding dedicated code with no other use case than human extraction 🙂
Please re-open if I missed a use case!

PR #32, opened 2020-07-01
Few improvements

LucasVerneyDGE:
Hi,

Here is a PR to address few improvements:

* Add support for `HTTP_PROXY` / `HTTPS_PROXY` environment variables (required to be runnable in my administration) ;
* Let user overload default data directory location with `DATA_PATH` environment variable ;
* Improve documentation about initialization of `cgus-data` repository.

Best

PR #128, opened 2020-09-10
Change the recipient of error emails

Ndpnt:


MattiSG (approved):

config/production.json
```diff
@@ -6,7 +6,7 @@
   },
   "logger": {
     "sendMailOnError": {
-      "to": "nicolas.dupont@beta.gouv.fr",
+      "to": "matti.schneider@beta.gouv.fr",
```
LucasVerneyDGE, line 9: I've just noticed this configuration option thanks to this PR. I don't see it documented in the "configuration" part of the README, https://github.com/ambanum/CGUs#configuring.

Is it documented somewhere else (I probably missed it)? Same for the extra environment options for `SMTP_*` variables.

If this is not the case, I can write a PR about these settings if you want (in the place that works best for you).

MattiSG: Lovely, thanks for offering! Any help on improving the documentation would be welcome 😃

We should have updated the “Configuring” section of the README in #112 to reflect this new option 😞  This is where this documentation should go.

If you could document the environment variables there as well, perhaps splitting both parts in subtitles (`#### Config file` / `#### Environment variables`) and documenting the `HTTP*_PROXY` environment variables you added in #32, it would be awesome!

PR #132, opened 2020-09-11
Drop Facebook redirections in Facebook.filters.js

LucasVerneyDGE:
Hi,

Was keeping the `http://l.facebook.com/l.php` part on purpose? It seems we can completely get rid of the redirection with an extra step of regex matching, which might help lowering the noise.

In particular, considering https://github.com/ambanum/CGUs-versions/commit/4f66b53ee, after this change it would reduce to a simple HTTP / HTTPS noise which could be taken into account globally.

Best,

### DIFF
--- a/services/Facebook.filters.js
+++ b/services/Facebook.filters.js
@@ -20,7 +20,7 @@ export function cleanUrls(document) {
   const links = document.querySelectorAll('[href*="https://l.facebook.com/l.php?"],[href*="http://l.facebook.com/l.php?"]');
 
   links.forEach(link => {
-    link.href = link.href.replace(/&h=\S*/, '');
+    link.href = decodeURIComponent(link.href.replace(/&h=\S*/, '').replace(/(\S*)\?u=(\S*)/, '$2'));
   });
 }
 


MattiSG:
Thanks @LucasVerneyDGE! The aim was to reduce noise by removing unique tracking IDs, but you're right with that extra removal!

In general, I think we should favor minimal interventions in filters in order to minimise the risk of not tracking the actual document content, but the reference of noise you gave combined with the sort of intervention justifies in my view this improvement 🙂 

Do you agree with the general notion that we should minimise interventions? If so, I'll add it to the guidelines that will soon be consolidated from https://github.com/ambanum/CGUs/issues/106#issuecomment-681870747.

MattiSG (commented):
Looks good, but I'll approve once tested locally 🙂

LucasVerneyDGE:
> Thanks @LucasVerneyDGE! The aim was to reduce noise by removing unique tracking IDs, but you're right with that extra removal!
> 
> In general, I think we should favor minimal interventions in filters in order to minimise the risk of not tracking the actual document content, but the reference of noise you gave combined with the sort of intervention justifies in my view this improvement 🙂
> 
> Do you agree with the general notion that we should minimise interventions? If so, I'll add it to the guidelines that will soon be consolidated from [#106 (comment)](https://github.com/ambanum/CGUs/issues/106#issuecomment-681870747).

:+1: definitely agree with this approach!

eduardsmith (approved):

eduardsmith (approved):

Ndpnt (approved):

PR #134, opened 2020-09-16
Filter out country version blinking in Google Terms of Services

LucasVerneyDGE:
Hi,

I guess this should do the trick to filter out all "country version" elements blinking in Google Terms of Services. Sadly, there is no narrow selector which could be used to filter out this text, hence I have to remove the full section of the ToS by looking at section titles.

I did not find any more robust way so far, I'm welcoming any idea on this.

Best,

### DIFF
--- a/services/Google.filters.js
+++ b/services/Google.filters.js
@@ -4,3 +4,19 @@ export function removeUTMfromUrls(document) {
     link.href = link.href.replace(/utm_[^=]*=[^&]*[&]{0,1}/g, '');
   });
 }
+
+export function removeCountryVersion(document) {
+  // Remove "country version" link at top of the page
+  document.querySelector('a[data-name="country-version"]').parentNode.remove();
+
+  // Remove all the "country version" section in the CGUs
+  const countryVersionTitle = document.querySelector('#footnote-country-version');
+  const disclaimerTitle = document.querySelector('#footnote-disclaimer');
+
+  let node = countryVersionTitle;
+  while (node !== disclaimerTitle.previousSibling) {
+    const next = node.nextSibling;
+    node.remove();
+    node = next;
+  }
+}
--- a/services/Google.filters.js
+++ b/services/Google.filters.js
@@ -4,19 +4,3 @@ export function removeUTMfromUrls(document) {
     link.href = link.href.replace(/utm_[^=]*=[^&]*[&]{0,1}/g, '');
   });
 }
-
-export function removeCountryVersion(document) {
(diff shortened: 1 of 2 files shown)


Ndpnt:
Hi @LucasVerneyDGE, thanks for this noise removing suggestion !

In this case, instead of creating a new filter you can use the `remove` key of the document declaration to have something similar:
```JSON
{
  "name": "Google",
  "documents": {
    …
    "Terms of Service": {
      "fetch": "https://policies.google.com/terms?hl=en-GB",
      "filter": [
        "removeUTMfromUrls"
      ],
      "select": "div[role=article]",
      "remove": [
        {
          "startBefore": "a[data-name='country-version']",
          "endBefore": "#whats-covered"
        },
        {
          "startBefore": "#footnote-country-version",
          "endBefore": "#footnote-disclaimer"
        }
      ]
    }
  }
}

```

LucasVerneyDGE:
I missed this great addition, thanks for pointing out! This PR has been updated accordingly. Best

LucasVerneyDGE:
By the way, in this service the removed part is known in advance (either empty or some known text). What about adding tests on these remove filter to ensure we are not removing too much information (typically, such a test could be here that we do not remove more than one `<h3>` tag)?

Ndpnt:
> I missed this great addition, thanks for pointing out! This PR has been updated accordingly. Best
Thanks for the quick update!

Unfortunately, I'm sorry but I just saw that the modification that I asked to you won't work every time. When the country version content is not available, when the filter will try to remove it, it will throw an error. So in this specific case we cannot use the remove filter and we have to use a service filter like you did. 
But we also need to add a check in the code you wrote before to do nothing if the country specific content is not available.

I will handle the rollback and the code update.

Ndpnt:
> By the way, in this service the removed part is known in advance (either empty or some known text). What about adding tests on these remove filter to ensure we are not removing too much information (typically, such a test could be here that we do not remove more than one `<h3>` tag)?

For know, we can add this kind of tests directly inside the filter and made it throw if we try to remove more than one `h3`.

> Also, is there a way to run the service code on a given HTML file directly (typically for easier testing of such a change)?

Not really. Currently, in the snapshot repo, I checkout to the commit corresponding to the HTML content I want to filter, and run the refilter command to check if my filter works.

Ndpnt:
I updated the code on this branch `LucasVerneyDGE-google-country`

Ndpnt:
Merged in #140

PR #136, opened 2020-09-21
Handle errors on services filters functions

Ndpnt:
Avoid stopping the whole process if there is an error on a filter function.

### DIFF
--- a/src/logger/index.js
+++ b/src/logger/index.js
@@ -22,16 +22,26 @@ const consoleTransport = new winston.transports.Console();
 const transports = [ consoleTransport ];
 
 if (config.get('logger.sendMailOnError')) {
-  transports.push(new winston.transports.Mail({
-    level: 'error',
+  const mailerOptions = {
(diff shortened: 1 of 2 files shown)


src/logger/index.js
```diff
@@ -22,16 +22,26 @@ const consoleTransport = new winston.transports.Console();
...
+  transports.push(new winston.transports.Mail({
+    ...mailerOptions,
+    level: 'error',
+    subject: '[CGUs] Error Report',
+  }));
+
+  transports.push(new winston.transports.Mail({
+    ...mailerOptions,
+    level: 'warn',
+    subject: '[CGUs] Warning Report',
```
MattiSG, line 44: If a “warning” is always a “Content not accessible”, could we maybe:
```suggestion
    subject: '[CGUs] Inaccessible content',
```

MattiSG (approved):
Not tested locally, but looks good.

Happy to see confirmation that the check for async for unnecessary :wink:

PR #137, opened 2020-09-21
Limit attempts to restart to avoid looping on an unresolved error

Ndpnt:


ops/roles/cgus/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
 
 - name: Start CGUs
   command:
-    cmd: forever start -a --uid "{{ app }}" -c "npm run start:scheduler:prod" ./
+    cmd: forever start -m 10 -a --uid "{{ app }}" -c "npm run start:scheduler:prod" ./
```
MattiSG, line 39: > Maybe couple that with [`--spinSleepTime`](https://www.npmjs.com/package/forever) to decrease the probability of transient errors?
```suggestion
    cmd: forever start -m 10 --spinSleepTime 10000 -a --uid "{{ app }}" -c "npm run start:scheduler:prod" ./
```

MattiSG (approved):
Not tested locally but looks good.

ops/roles/cgus/tasks/main.yml
Ndpnt: I'm not sure the option `--spinSleepTime` do what you think. What I understand is correspond to the min duration to consider a script stuck. I already tried to use it but it does not make forever wait for the specified time before re-running the script.

PR #142, opened 2020-09-22
Fix services validation on CI

Ndpnt:


scripts/validation/validate.js
```diff
@@ -132,6 +132,7 @@ let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
     run();
   } catch (error) {
     console.error(error);
+    process.exit(1);
```
MattiSG, line 135: Why not just let the error be thrown, then? 🤔

.github/workflows/test.yml
```diff
@@ -14,6 +14,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+    - run: git checkout HEAD^
+    - run: git fetch origin master:master
```
MattiSG, line 21: > Why do we have to always duplicate these instructions? Can't we just have both workflows chained?

.github/workflows/deploy.yml
```diff
@@ -15,6 +15,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+    - run: git checkout HEAD^
```
MattiSG, line 21: Why do we go to the parent of `HEAD` instead of testing the last commit?

MattiSG:
Fails because `Error: Could not find any service with id "./test/helper.js"`, which is due to the fragile `arguments.slice(6)` call. See https://github.com/ambanum/CGUs/pull/110#discussion_r484869875 for more context.

.github/workflows/deploy.yml
```diff
@@ -15,6 +15,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+    - run: git checkout HEAD
+    - run: git fetch origin master:master
```
MattiSG, line 22: I don't understand why we have to do this manually. Let's 📞

Ndpnt:
Yep I know, will fix it tomorrow  :)

MattiSG:
Ok, ask me again for a review when fixed :wink:

.github/workflows/deploy.yml
Ndpnt: Because the checkout github action only checkout the current branch and we need to have the current branch and the master to allow getting all files that are modified between these two branches.
I though with the option `fetch-depth: 0` it should not be necessary as explained [here](https://github.com/actions/checkout#Checkout-a-different-branch), but if I remove `git checkout HEAD` and `git fetch origin master:master` it does not work… Not very clear for me.

package.json
```diff
@@ -18,7 +18,7 @@
     "posttest": "npm run lint",
     "test": "NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\"",
     "test:debug": "NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\" --inspect-brk --exit",
-    "validate": "NODE_ENV=validation mocha --delay --file './test/helper.js' scripts/validation/validate.js",
+    "validate": "NODE_ENV=validation node --unhandled-rejections=strict ./node_modules/mocha/bin/mocha --delay --file './test/helper.js' scripts/validation/validate.js",
```
MattiSG, line 21: Using the bin alias is more robust than navigating the filesystem within the module.
```suggestion
    "validate": "NODE_ENV=validation node --unhandled-rejections=strict ./node_modules/.bin/mocha --delay --file './test/helper.js' scripts/validation/validate.js",
```

MattiSG (approved):
Tested locally, passes both `npm run validate` and `npm run validate [service]`.

PR #145, opened 2020-09-23
Do not track images from Google Privacy / ToS

LucasVerneyDGE:
This should address https://github.com/ambanum/CGUs/issues/143.

EDIT: CI is failing, but seems independent from my change.


Ndpnt (approved):

MattiSG:
🎉 88% file size reduction, went down from 240 to 31 kB and decreased chances of noise, thanks @LucasVerneyDGE! 😃

PR #146, opened 2020-09-23
Track Facebook Data Transfer Addendum

LucasVerneyDGE:
Fixes https://github.com/ambanum/CGUs/issues/144

### DIFF
--- a/services/Facebook.json
+++ b/services/Facebook.json
@@ -41,6 +41,14 @@
         "cleanUrls"
       ],
       "select": "#content"
+    },
+    "Data Controller Agreement": {
+      "fetch": "https://www.facebook.com/legal/EU_data_transfer_addendum",
+      "filter": [
+        "numberListCorrectly",
+        "cleanUrls"
+      ],
+      "select": "#content"
     }
   }
 }
--- a/services/Facebook.json
+++ b/services/Facebook.json
@@ -42,7 +42,7 @@
       ],
       "select": "#content"
     },
-    "Data Controller Agreement": {
+    "Data Processor Agreement": {
       "fetch": "https://www.facebook.com/legal/EU_data_transfer_addendum",
       "filter": [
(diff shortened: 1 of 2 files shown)


services/Facebook.json
```diff
@@ -41,6 +41,14 @@
         "cleanUrls"
       ],
       "select": "#content"
+    },
+    "Data Controller Agreement": {
```
LucasVerneyDGE, line 45: What should be the correct type for this one, in your opinion?

Ndpnt: Not so easy, I suggest "Data Transfer Addendum" because it seems to me that "Data Controller Agreement" would be this document https://www.facebook.com/legal/terms/dataprocessing.

MattiSG: I would go with `Data Processor Agreement`, since this addendum applies to the _Commercial Terms_ (hence targeting business entities, not individuals), and applies “to the extent that Facebook Ireland Limited is acting as your Processor of EU Data under the applicable product terms”, establishes that “by providing EU Data to Facebook Ireland Limited as your Processor”, the commercial entity that provides user data agrees to keep subprocessor information confidential and a couple other things that only relate to subprocessing.

I would consider https://www.facebook.com/legal/terms/dataprocessing as the `Data Controller Agreement`, as it defines how user data should be handled jointly by both Facebook and the commercial entity that transfers user data to Facebook.

I'll ask for advice from people who know more 🙂

MattiSG (changes requested):

services/Facebook.json
Ndpnt: @MattiSG any news about this?

MattiSG (approved):
Tested & validated locally. Thanks @LucasVerneyDGE! 😃

LucasVerneyDGE:
Thanks!

PR #148, opened 2020-09-25
Clean DBs at each restart to allow restarting with a dirty state

Ndpnt:


MattiSG (approved):
I like how that simplifies the Ansible setup process, and the simplicity of the approach.

I am however concerned that this means:

1. We would lose history in case of problem with the publishing process, either in our code or on the distant server availability.
2. We lose a local clone of the history.
3. If GitHub is unavailable, we cannot deploy.

We had already discussed of the first two points, but I believe that the last one is blocking.

MattiSG (changes requested):
I like how that simplifies the Ansible setup process, and the simplicity of the approach.

I am however concerned that this means:

1. We would lose history in case of problem with the publishing process, either in our code or on the distant server availability.
2. We lose a local clone of the history.
3. If GitHub is unavailable, we cannot deploy.

We had already discussed of the first two points, but I believe that the last one is blocking.
Absolute minimal changes would be to check availability of the distant server before deleting the database.

Ndpnt:
Changed how it works to avoid the problems you're talking about.
- Do not remove anymore the whole repos.
- Create a backup of local history by creating a tag 
- Reset hard state to origin master and remove `index.lock` to ensure being in a clean state
- Clean old tags (keep only the last 20)

So it should fix 1. and 2.
For 3., if Github is down the code is not fetchable so in any case we cannot deploy. (Moreover our CD is based on Github…)

MattiSG:
> if Github is down the code is not fetchable so in any case we cannot deploy. (Moreover our CD is based on Github…)

The first part of the argument sounds relevant to me, but not the second part: we can (and do) also trigger deployment manually.

I understand it was not clear in the way I phrased my point 3, sorry about that! But what bothered me, and why I said “Absolute minimal changes would be to check availability of the public database before deleting the local database.”, was mostly the following case:

1. See that CD fails for whatever reason, in a moment of rush.
2. Manually deploy.
3. Deployment fails _after_ the local database is destroyed, because actually the reason for failure was a GitHub unavailability.
4. Cry.

ops/roles/cgus/vars/main.yml
```diff
@@ -1,3 +1,9 @@
-repo: https://github.com/ambanum/CGUs.git
+repository: https://github.com/ambanum/CGUs.git
+versions_repository: git@github.com:ambanum/CGUs-versions.git
+versions_directory: cgus-versions
+versions_deployed_branch: master
+snapshots_repository: git@github.com:ambanum/CGUs-snapshots.git
+snapshots_directory: cgus-snapshots
+snapshots_deployed_branch: master
 app: cgus
 app_port: 3000
```
MattiSG, line 9: > Is this actually used anywhere? 🤔 I see it defines a `PORT` env var, but I don't think we use it anywhere…

ops/roles/cgus/vars/main.yml
```diff
@@ -1,3 +1,9 @@
-repo: https://github.com/ambanum/CGUs.git
+repository: https://github.com/ambanum/CGUs.git
+versions_repository: git@github.com:ambanum/CGUs-versions.git
+versions_directory: cgus-versions
+versions_deployed_branch: master
+snapshots_repository: git@github.com:ambanum/CGUs-snapshots.git
+snapshots_directory: cgus-snapshots
+snapshots_deployed_branch: master
```
MattiSG, line 7: > Sort alphabetically.

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,44 @@
...
+    key_file: "/home/{{ ansible_user }}/.ssh/cgus-bot-key"
+  register: git_clone_app_finished
+  tags:
+    - setup
+    - update
+
+- name: Reset {{ name }} on origin {{ branch }}
+  command:
+    cmd: git reset --hard origin/{{ branch }}
+    chdir: "/home/{{ ansible_user }}/{{ directory }}"
```
MattiSG, line 31: If we want to always blindly trust the remote, then the following commands need to be issued:

```sh
# One does not simply `git pull`: we need to clean all leftover files and support force pushes
git fetch origin $TARGET_BRANCH  # get the remote code, do not try to apply it just yet
git checkout --force --detach origin/$TARGET_BRANCH  # ignore divergences, always trust remote
git clean --force  # remove all leftover files in tracked folders, prevent conflicts with previously ignored files
```

[Source](https://gist.github.com/MattiSG/3a5e106449a0786e4454d11338633bcd#file-deploy-sh-L27-L29).

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,44 @@
+- name: Remove index.lock in {{ name }}
+  file:
+    path: "/home/{{ ansible_user }}/{{ directory }}/.git/index.lock"
+    state: absent
+
+- name: Create backup tag for {{ name }}
```
MattiSG, line 6: Now that we simply `reset`, we don't really need these tag backups _if they are cleaned up automatically_: if we have strong confidence in our process, we can always resuscitate history with `reflog`. These are not eternal, but I'd expect them to survive them before `git gc` runs for about as long —if not longer— than 20 deployments.

MattiSG (commented):
I'm not sure that _blindly trusting the remote_ is the most appropriate answer to _clean DB to allow restarting with a dirty state_. What about simply removing the lockfile and clearing local uncommitted files? I'd feel safer knowing that divergences with the remote would trigger a failure and a manual review, rather than having a fully automated process that hides divergences and potentially loses them after 20 deploys.
Let's discuss this 🙂 🗣️

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,62 @@
...
+    repo: "{{ repository }}"
+    dest: "/home/{{ ansible_user }}/{{ directory }}"
+    depth: 1
+    accept_hostkey: yes
+    key_file: "/home/{{ ansible_user }}/.ssh/cgus-bot-key"
+  when: git_dir.before is defined and not git_dir.before
+  tags:
+    - setup
+
+- name: Update {{ name }}
```
MattiSG, line 36:
```suggestion
- name: Get latest data from public instance of {{ name }}
```

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,62 @@
...
+  command:
+    cmd: git pull
```
MattiSG, line 38:
```suggestion
    cmd: git checkout master && git pull origin master:master
```

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,62 @@
...
+    - update
+
+- name: Clone {{ name }}
+  git:
+    repo: "{{ repository }}"
+    dest: "/home/{{ ansible_user }}/{{ directory }}"
+    depth: 1
+    accept_hostkey: yes
+    key_file: "/home/{{ ansible_user }}/.ssh/cgus-bot-key"
+  when: git_dir.before is defined and not git_dir.before
```
MattiSG, line 32: I don't understand any of these conditionals, and the Ansible documentation really doesn't help me at all. We need to document this. Might be easier to talk 🗣️

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,62 @@
+- name: Get information about {{ name }}
```
MattiSG, line 1: I don't understand the use of this task. It should probably be renamed.

MattiSG:
See #181 before merging.

MattiSG (approved):
After discussion with @Ndpnt and improving tasks names and documentation, LGTM! 😥

ops/roles/cgus/tasks/database.yml
```diff
@@ -0,0 +1,72 @@
...
+    depth: 1
+    accept_hostkey: yes
+    key_file: "/home/{{ ansible_user }}/.ssh/cgus-bot-key"
+  when: existing_repository.before is defined and not existing_repository.before  # if existing_repository.before is null, then the repository already is new
+  tags:
+    - setup
+
+- name: Get latest data from public instance of {{ name }}
+  command:
+    cmd: git pull origin master:master
```
Ndpnt, line 38: If there are unpushed commits, we get this error:
```
    From github.com:ambanum/CGUs-versions
     ! [rejected]        master     -> master  (non-fast-forward)
```
It works if we replace `git pull origin master:master` by `git pull`

PR #160, opened 2020-10-05
Run CI on every push

MattiSG:
To get earlier feedback and to allow for silent CI tests (as opposed to always having to open a pull request), this changeset sets the `test` workflow to run on every push instead of running only when PRs are opened against `master`.

It also separates running tests and deploying, ensuring the bijection between a commit to `master` and a version in production, making the assumption that pushes to `master` have always be checked (or in any case that the person pushing takes the responsibility for it). This works well when combined with tests being run on every push and branch protection set up to enforce tests passing before a push to `master` is allowed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,6 @@
 name: Test
 
-on:
-  pull_request:
-    branches: [ master ]
+on: [ push, pull_request ]
 
 env:
   SENDINBLUE_API_KEY: ${{ secrets.SENDINBLUE_API_KEY }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,6 @@ name: Test
 
 on: [ push, pull_request ]
 
-env:
-  SENDINBLUE_API_KEY: ${{ secrets.SENDINBLUE_API_KEY }}
-
 jobs:
   test:
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,11 +1,10 @@
-name: Test & Deploy
+name: Deploy
 
 on:
   push:
     branches: [ master ]
 
 env:
-  SENDINBLUE_API_KEY: ${{ secrets.SENDINBLUE_API_KEY }}
   ANSIBLE_VAULT_PASSWORD: ${{ secrets.ANSIBLE_VAULT_PASSWORD }}
 
 jobs:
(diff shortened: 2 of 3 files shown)


Ndpnt (approved):

.github/workflows/test.yml
```diff
@@ -1,8 +1,6 @@
 name: Test
 
-on:
```
Ndpnt, line 3: I see that it runs test both for `push` and for `pull_request`… So I think we could only use `push`.

MattiSG:
As I understand it we need both, as we want tests to run for external contributions as well, which are not pushed to this repository but are exposed in a pull request  :)

Ndpnt:
So we will have tests ran twice on almost on every PR? 🤔

Ndpnt:
From what I understand, we can't do what you want because to run actions, the forked repository should [enable github actions on its own side ](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories).
> Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.

Do you understand the same?

MattiSG:
> Do you understand the same?

No. The paragraph above states:

> When you create a pull request from a forked repository to the base repository, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository.
> Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.

I thus understand that a forked repository will not have its own Actions enabled by default (which makes sense), but that opening a PR against a target repository will trigger Actions on the target repo.

Anyway, what do we risk trying? 🙂 

> So we will have tests ran twice on almost every PR? 🤔

That's not the behavior I expect. I expect GitHub Actions to work at commit (SHA) level, not at event level. [Source](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration):

> The workflow runs on your repository's code at the same commit SHA and Git ref that triggered the event.

Ndpnt:
> Anyway, what do we risk trying? 🙂

Already tried, and it ran tests twice, one for the `push` event and one for `pull_request` event…

MattiSG:
If we want to run `deploy` only if `test` passes, we can use [`workflow_run`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run):

```yaml
# In deploy.yaml
on:
  workflow_run:
    workflows: [ Test ]
    branches: [ master ]
    types: 
      - completed
```

MattiSG:
> Already tried, and it ran tests twice

Unfortunately I cannot find a way in the documentation to specify that we want to trigger a run for `pull_request` events only if they are sent from forks… and if we don't run on `pull_request` then I understand that forks will never be tested, which is not acceptable.

MattiSG:
I really don't understand how GitHub Actions presents information. If I go to _All workflows_, it seems there was only one run:

Capture d’écran 2020-10-15 à 11 22 35

However inline in the PR, there seems to be two:

Capture d’écran 2020-10-15 à 11 25 08

I'll go with believing there is only one run 🤷 😅

MattiSG:
([Use npm-provided alias for binaries](/ambanum/CGUs/pull/160/commits/a67075d735e5be848e595174118b759defca5c88) is a knight following #175, as seen with @Ndpnt)

PR #162, opened 2020-10-06
Add "serviceDeclarationsPath" in production config

michielbdejong:
This config var seems to be missing here?


MattiSG:
Environment-specific configurations (such as `development` or `production`) are merged with the `default` config, so it is not necessary to repeat all items in the `default` 🙂

PR #167, opened 2020-10-09
Make project compatible with Windows systems

DamienSeuxADLC:
The project does not currently run on windows when following instructions from README, because of compatibility issues in the way the directory name is obtained, and in the way service filters are imported.

This PR fixes these issues.


MattiSG:
Thanks a lot @DamienSeuxADLC! It's exciting to see support for additional platforms 😃

We'll do our best to review and merge this asap, but we'll need to find a way to check behavior on Windows as well. If it's too long to be done, we'll simply make sure this changeset is compatible with UNIX and rely on your own checks 🙂

DamienSeuxADLC:
Thanks for the answer @MattiSG . I'm aware that supporting Windows requires ressources, but it would be useful for individual contributors who want to add services.

MattiSG:
Totally! And it's super important that we expand environments support 🙂 I just wanted to outline that we'll definitely merge this, but that it might be slightly longer until we can officially claim Windows support.

But we found a solution and should be able to have this finalized this week 😃

Ndpnt:
@DamienSeuxADLC 
We succeed to use CGUs on a Windows systems without modifying the code.
Which version of Windows and which version of Node do you use?

DamienSeuxADLC:
@Ndpnt I'm using node  v12.18.4 and Windows Server 2012 R2.

I got two errors, the first when running the unmodified code, the second after fixing the first one : 
When trying to initialise git directories, simple-git is given the following path : C:\\C:\\Users\\[...]\\cgus\\data\\snapshots when there should only be one C:\\ at the beginning.
When trying to import .js filters files, and is caused by this issue : https://github.com/nodejs/node/issues/31710

Ndpnt:
Hi @DamienSeuxADLC,

I tried this changeset and it seems compatible with UNIX 👍.

For the second error you have, I saw you use `pathToFileURL` to import `filters` whereas everywhere else you use `fileURLToPath`, so I was wondering if it could comes from here?

And thanks for this PR, it allowed me to detect that our lint validation is broken for a little while, so I will fix it! 😉

DamienSeuxADLC:
Hi @Ndpnt ,
I'm not sure I understand your message correctly.
To be more specific : the two errors I mentionned in my last message happened when using the code from the current master branch. The code from this PR runs without error (at least on my machine).
I'm using a different method to import filters because import does not support absolute file paths on Windows, but works without any issue with file URL, whereas the initialisation of a git repo can be done with a file path.

Ndpnt:
Ok, I did not understood properly. I though you succeeded to fix the first error but not the second one.

So if everything is working for you in this PR it's perfect. 
I will just do a final check :).

MattiSG:
Could you please try running ` npm run validate:schema` in your Windows environment? I'm investigating limitations and there might be one there…

DamienSeuxADLC:
I tried running  `npm run validate:schema` and I got the same error. It looks like all scripts from package.json that set environnement variables will fail for the same reason.
I can look for a cross platform way to set these and add the change to this PR, unless you prefer a new one.

MattiSG:
Good, thanks for checking @DamienSeuxADLC! 😃
I'm in favor of adding in this PR, so that we have one consolidated change that actually makes the codebase compatible with Windows 🙂
Of course, if you'd prefer to have this first changeset merged so you can start some other type of work in parallel, let us know, we'll happily merge partial support!

MattiSG:
I added support for Windows testing in CI in the `ci-windows` branch. These commits should be integrated in this branch before merging 🙂

MattiSG:
I see that you copy-pasted in 99212bb the contents of the `ci-windows` branch, can you confirm this final version works on your setup? 🙂

package.json
```diff
@@ -9,16 +9,16 @@
...
     "refilter": "node src/index.js --refilter-only",
     "setup": "node scripts/setup.js",
     "lint": "./node_modules/.bin/eslint test src services",
     "posttest": "npm run lint",
-    "test": "NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\"",
-    "test:debug": "NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\" --inspect-brk --exit",
-    "validate": "NODE_ENV=validation node --unhandled-rejections=strict ./node_modules/.bin/mocha --delay --file './test/helper.js' scripts/validation/validate.js",
+    "test": "cross-env NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\"",
+    "test:debug": "cross-env NODE_ENV=test mocha --file './test/helper.js' --recursive \"./src/**/*.test.js\" --inspect-brk --exit",
+    "validate": "cross-env NODE_ENV=validation node --unhandled-rejections=strict node_modules/mocha/bin/_mocha --delay --file './test/helper.js' scripts/validation/validate.js",
```
MattiSG, line 21: Navigating in the module filesystem is not safe (it is akin to using a private API instead of the public one). Are there no binary symlink folders in Windows?

.github/workflows/test.yml
```diff
@@ -5,17 +5,23 @@ on: [ push, pull_request ]
...
+         git config --global user.email 'ambnum.bot@disinfo.quaidorsay.fr'
+         git config --global core.autocrlf false
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with: 
+          node-version: 12.x
+      - run: npm ci
+      - run: npm run validate:schema
+      - run: npm test
```
MattiSG, line 27: > Missing final newline.

scripts/validation/validate.js
```diff
@@ -6,18 +6,19 @@ import chai from 'chai';
...
 import serviceSchema from './service.schema.js';
 
 const fs = fsApi.promises;
-const __dirname = path.dirname(new URL(import.meta.url).pathname);
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const { expect } = chai;
 const rootPath = path.join(__dirname, '../..');
 const MIN_DOC_LENGTH = 100;
-const filePathRelativeToRoot = new URL(import.meta.url).pathname.replace(`${rootPath}/`, '');
+const filePathRelativeToRoot = path.relative(rootPath, fileURLToPath(import.meta.url)).split(path.sep).join('/');//Convert possible Windows path to posit version from args
```
MattiSG, line 21: Update ESlint rules to enforce space before and after `//`.

MattiSG, line 21: I guess that's a typo? 🙂
```suggestion
const filePathRelativeToRoot = path.relative(rootPath, fileURLToPath(import.meta.url)).split(path.sep).join('/');//Convert possible Windows path to POSIX version from args
```

MattiSG (changes requested):
Thanks @DamienSeuxADLC! It looks like we almost have support for Windows 😃

Before merging, we'll still need to:

- Rebase (potentially splitting 99212bb and integrating af1b93e4ded3bcf438bc712c0485601463ec479e and a391ff703b69a292f2eaf0256e316136e330a1e0 if we want to preserve sausage making).
- Document platform support in `README` (to brag) & `CONTRIBUTING` (to explain the constraints such as the use of `fileToPath` to contributors).

DamienSeuxADLC:
@MattiSG The commit history of this pull request is indeed far from optimal, I think the best way to proceed is to close this PR and open a new one with a clean history, rebased on `ci-windows`. Would this be ok with you ?
Regarding your other comments, I can fix them in the new PR.

MattiSG:
Sure, that's fine on our side! I've rebased `ci-windows` on top of `master` if you want to rebase on top of it 🙂
You can also exceptionally force-push on this branch if you feel like it!

MattiSG:
My suggestion would be to rebase on top of `ci-windows` and force-push on your original branch, so that we keep the discussion history here. If that looks too complex, feel free to close this PR and open a new one referencing it 🙂

DamienSeuxADLC:
I made the changes, including an addition to `README`.
I am not sure that `CONTRIBUTING.md` needs an addition as limitation around path handling do not matter when adding new services.

Ndpnt:
Superseded by #186

PR #173, opened 2020-10-14
Services imported from tosback2 using ambanum/TOSBack-CGUs-bridge

michielbdejong:


michielbdejong:
Ah sorry, I just noticed this doesn't fully follow https://github.com/ambanum/CGUs/issues/106#issuecomment-681870747
and that leads to e.g. creating services/Deviantart.json although services/DeviantArt.json already exists. Will fix.

Ndpnt:
Hi @michielbdejong,
I just want to clarify our needs for service importation:
- For services which pass the validation, it's important to us they comply with service naming standards described in #106 
- For services which are successfully converted (document type is supported in CGUs and xpath to css conversion succeed) but do not pass the validation, can you save the resulting declarations in a dedicated folder in the TOSBack-CGUs-bridge repository ?
- For services which are not successfully converted (document type is unsupported in CGUs or xpath to css conversion failed), can you save the resulting declarations (with the original document type and xpath selector in `select` key)  in another dedicated folder in the TOSBack-CGUs-bridge repository ?

Ndpnt:
Also, for history importation:
- Filenames and commits messages comply with service naming standards
- Commits author date should match the original ToSback snapshot date
- Import all snapshots from ToSBack, even when the types of documents are not currently supported in CGUs and even if documents do not pass validation. It will allow us to recover the history of theses documents when we will add these services and documents later.
- Can you provide a list of services and documents for which the history could not be imported in the TOSBack-CGUs-bridge repository?

Ndpnt:
Superseded by #184

Issue #156, opened 2020-09-29
Document declarations and filters versioning


Ndpnt:
We need to be able to regenerate versions from snapshots. As documents is aim to change over time (location or filters) we can't rely on the last version of the declaration to regenerate the version from an old snapshot. So we need a system to keep track of declaration changes, that's what we called **declarations and filters versioning**.

At this time, we see three solutions which have in common the following rules:
 - `history` is optional
 - the current valid declaration has no date and should be clearly identifiable
 - the `valid_until` date is an inclusive expiration date. It should be the exact authored date of the last snapshot commit for which the declaration is still valid.

## Option 1: Add an `history` field in service declaration

In `services/ASKfm.json`:
```
{
  "name": "ASKfm",
  "documents": {
    "Terms of Service": {
      "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
      "select": ".selection",
      "filter": [ "add" ]
      "history": [
        {
          "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
          "select": "body",
          "filter": [ "add" ]
          "valid_until": "2020-08-24T14:02:39Z"
        },
        {
          "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
          "select": "body",
          "valid_until": "2020-08-23T14:02:39Z"
        }
      ]
    }
  }
}
```

Note: When no historisation is needed the file may have no mention of history.

**Pros:**
- Everything is in the same file:
    - might prevent to forget to update existing history
    - might help user to know that history is a thing and encourage them to learn about it if they feel the need
    - no (pseudo-)hidden knowledge about history

**Cons:**
- Apparent complexity can discourage new contributors
- With time, the file can become huge

## Option 2: Add an `serviceId.history.json` file

In `services/ASKfm.json`:
```
{
  "name": "ASKfm",
  "documents": {
    "Terms of Service": {
      "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
      "select": ".selection",
      "filter": [ "add" ]
    }
  }
}
```

In `services/ASKfm.history.json`:
```
{
  "name": "ASKfm",
  "documents": {
    "Terms of Service": [
      {
        "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
        "select": "body",
        "filter": [ "add" ]
        "valid_until": "2020-08-24T14:02:39Z"
      },
      {
        "fetch": "https://ask.fm/docs/terms_of_use/?lang=en",
[...]

PR #174, opened 2020-10-14
Add command to run CGUs only on modified services

LucasVerneyDGE:
Hi,

When adding new services, it is convenient to be able to run in a single command the checks on the modified services. Here is a proposal for such a npm command. Let me know what you think about it and whether it might be of interest :)

Best,

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,16 +3,23 @@ import scheduler from 'node-schedule';
 import CGUs from './app/index.js';
 import logger from './logger/index.js';
 import Notifier from './notifier/index.js';
+import { getModifiedServices } from './utils/index.js';
 
 const args = process.argv.slice(2);
+const modifiedOnly = args.includes('--modified-only');
 const refilterOnly = args.includes('--refilter-only');
 const schedule = args.includes('--schedule');
-let serviceIds = args.filter(arg => !arg.startsWith('--'));
 
 (async () => {
   const app = new CGUs();
   app.attach(logger);
   await app.init();
+
+  let serviceIds = args.filter(arg => !arg.startsWith('--'));
+
+  if (modifiedOnly) {
+    serviceIds = await getModifiedServices();
+  }
 
   serviceIds = serviceIds.filter(serviceId => {
     const isServiceDeclared = app.serviceDeclarations[serviceId];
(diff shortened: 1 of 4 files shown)


LucasVerneyDGE:
Updated with this logic.

src/index.js
```diff
@@ -23,6 +32,9 @@ let serviceIds = args.filter(arg => !arg.startsWith('--'));
     return isServiceDeclared;
   });
 
+  if (modifiedOnly && !serviceIds.length) {
+    return;
+  }
```
Ndpnt, line 37: Can you add a new line after the `}`?

src/index.js
```diff
@@ -23,6 +32,9 @@ let serviceIds = args.filter(arg => !arg.startsWith('--'));
     return isServiceDeclared;
   });
 
+  if (modifiedOnly && !serviceIds.length) {
+    return;
```
Ndpnt, line 36: Add a log, something like `logger.warn('No services have been modified');`

src/tools/index.js
```diff
@@ -0,0 +1,19 @@
+import path from 'path';
```
Ndpnt, line 1: Rename `tools` in `utils`, it seems more idiomatic to me.

src/tools/index.js
```diff
@@ -0,0 +1,19 @@
...
+import simpleGit from 'simple-git';
+
+export async function getModifiedServices(rootPath) {
```
Ndpnt, line 4: As the path for `services` is always the same no matter where we call this function, I'm in favor of not passing `rootPath` as parameter but defining it directly inside this module.

src/index.js
```diff
@@ -1,19 +1,28 @@
...
 const schedule = args.includes('--schedule');
-let serviceIds = args.filter(arg => !arg.startsWith('--'));
 
 (async () => {
   const app = new CGUs();
   app.attach(logger);
   await app.init();
 
+  let serviceIds = args.filter(arg => !arg.startsWith('--'));
+  if (modifiedOnly) {
```
Ndpnt, line 20: Add a newline before the `if` to create meaningful blocs

Ndpnt (changes requested):
Hi @LucasVerneyDGE,
Thanks for this CLI improvement!
Just some minors changes and it's GTM for me.

Ndpnt:
Also, can you please rebase master branch.

LucasVerneyDGE:
Thanks for the review, PR should be updated!

MattiSG (approved):
Not tested locally, but discussed & read. Thanks @LucasVerneyDGE!

As complexity in handling services increases, for example when implementing #156, it will be appropriate to move the `getModifiedServices` logic in a `services` or `documents` module,  rather than `utils` 🙂

PR #170, opened 2020-10-12
Start tracking Airbnb, Amazon, Booking.com, eBay, Veepee, Vinted

TomHouriezDGE:
Hi,
Here are some additional services, let me know what you think of it !

Have a nice day :) 

EDIT:
Seems CI is failing.
I'm not sure what is happening, it has something to do with Veepee.json URL but it looks fine.

PR #176, opened 2020-10-15
Allow query params in URLs for PDF files

Ndpnt:
Fix the schema validation for PDF documents with query params in the URL added in PR #170.

### DIFF
--- a/scripts/validation/service.schema.js
+++ b/scripts/validation/service.schema.js
@@ -59,7 +59,7 @@ const schema = {
       properties: {
         fetch: {
           type: 'string',
-          pattern: '^https?:\/\/.+\.[pP][dD][fF]\??',
+          pattern: '^https?://.+\.[pP][dD][fF](\\?.+)?$',
           description: 'The URL where the document can be found'
         },
       },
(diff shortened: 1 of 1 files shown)


scripts/validation/service.schema.js
```diff
@@ -59,7 +59,7 @@ const schema = {
       properties: {
         fetch: {
           type: 'string',
-          pattern: '^https?://.+\.[pP][dD][fF]$',
+          pattern: '^https?:\/\/.+\.[pP][dD][fF]\??',
```
MattiSG, line 62: > I don't understand why we need to escape`/`.

Also, I'd prefer that we keep validating that the input is on one line only: it is a very good sign that the contributor forgot to close the quote.

```suggestion
          pattern: '^https?://.+\.[pP][dD][fF](\\?.+)?$',
```

(tested with :

- https://investor.apple.com/files/doc_downloads/gov_docs/Apple-Human-Rights-Policy.pdf → passes
- https://investor.apple.com/files/doc_downloads/gov_docs/Apple-Human-Rights-Policy.pdf? → fails
- https://investor.apple.com/files/doc_downloads/gov_docs/Apple-Human-Rights-Policy.pdf?a → passes
)

MattiSG (approved):
Do we have a use case for this?

MattiSG:
> Do we have a use case for this?

Yes, in #170, as mentioned in the opening comment 🙃

PR #177, opened 2020-10-15
Improve CI

MattiSG:
Following #160, we discovered that testing modified services failed on `master` because `git fetch origin master:master`, which is needed when working on a different branch, quite logically [fails](https://github.com/ambanum/CGUs/runs/1258267884?check_suite_focus=true) when the current branch is `master`.

This changes splits the code testing and the modified services validation in two separate workflows, which also has the added benefit of better readability. The services validation workflow is triggered only when the branch is not `master`, which by definition never has modified services.

I also made sure to test with the latest Node 12 version instead of sticking to Node 12.17 (for example, this workflow can now [run on Node 12.18](https://github.com/ambanum/CGUs/actions/runs/308362645)).

I considered creating a `.env.test` file instead of manually setting the git username and email, but running a `mv` command would not be cross-platform and #167 shows we should consider other platforms.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: '12.17'
+        node-version: 12.x
     - run: npm ci
     - run: |
         git config --global user.name 'AmbNum Bot'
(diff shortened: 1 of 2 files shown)


MattiSG:
I obviously tested, and failing modified services do [trigger a failure](https://github.com/ambanum/CGUs/runs/1258475002?check_suite_focus=true) 🙂 

Capture d’écran 2020-10-15 à 12 23 55

MattiSG:
If this gets merged, commits from branch `ci-windows` should be added to #167 🙂

Ndpnt (approved):

PR #178, opened 2020-10-20
Add new Google APIs service

LucasVerneyDGE:
Note that CI is failing outside of the scope of this PR.

### DIFF
--- a/services/Google APIs.json
+++ b/services/Google APIs.json
@@ -6,7 +6,7 @@
       "select": "article.devsite-article",
       "remove": ".devsite-article-meta, .devsite-thumb-rating"
     },
-    "Privacy Policy": {
+    "Developer Terms": {
       "fetch": "https://developers.google.com/terms/api-services-user-data-policy",
       "select": "article.devsite-article",
       "remove": ".devsite-article-meta, .devsite-thumb-rating"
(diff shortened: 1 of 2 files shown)


services/Google APIs.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Google APIs",
+  "documents": {
+    "Terms of Service": {
+      "fetch": "https://developers.google.com/terms",
+      "select": "article.devsite-article",
+      "remove": ".devsite-article-meta, .devsite-thumb-rating"
+    },
+    "Privacy Policy": {
```
LucasVerneyDGE, line 9: This is a bit specific policy, not strictly a "Privacy Policy" since the audience are the developers (end users of this specific services, yet still an intermediary), about personal data of the actual end users. Could be "Developer Terms" as well. I'm welcoming feedbacks on this one.

MattiSG: It's a tricky one indeed.

When reading the document, it doesn't feel right to me to qualify it as a “privacy policy” since it does not describe how the personal data of the Google APIs user (a developer) is going to be handled. It only describes principles and guidelines for handling personal data obtained through these APIs.

While these guidelines are focused on user data, I agree with you it more generally describes how the _service provider_ expects _developers_ to handle _APIs and programmatic access to content_, so I'd be in favour of using “Developer Terms”, the alternative being creating a new document type specifically focused on personal data, which seems premature to me.

LucasVerneyDGE: I agree with this, PR is updated. :)

PR #179, opened 2020-10-20
Remove redundant and unused document types

MattiSG:
“Legal Information” is not used anywhere, and “Program Policies” is too generic.

This might impact some of the import work, so would need a review from @michielbdejong.

### DIFF
--- a/services/AdSense.json
+++ b/services/AdSense.json
@@ -5,11 +5,6 @@
       "fetch": "https://www.google.com/adsense/tc/2018/UnitedKingdom.html",
       "select": "body"
     },
-    "Program Policies": {
-      "fetch": "https://support.google.com/adsense/answer/48182",
-      "select": ".article-container",
-      "remove": ".action-button, .zippy"
-    },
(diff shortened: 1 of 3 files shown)


MattiSG:
A rename in versions and snapshots will also be necessary since b9c428b changes a document type.

michielbdejong:
I don't use "Program Policies" but I do use "Legal Information" (see lines 84-86 of tosback-import.js). I could stop using it if you prefer?

MattiSG:
We'd need to check how to handle each document that uses this type. I don't want to slow down the import any longer, let's merge this after the import is done 😉

MattiSG:
Or maybe if there are not that many, I'd be happy to have a list of documents that use this type in here!

MattiSG:
We've started work on importing services from TosBack, and it is now clear that we'll need a manual review, as shown in #195. We will thus have the opportunity to change these document types to something else if needed.

Ndpnt (approved):

PR #218, opened 2020-12-08
Add Rakuten, Fnac.com Leboncoin.fr and RueduCommerce.fr services

Ndpnt:
Excluding legal questions, it fixes https://github.com/ambanum/CGUs/issues/166


Ndpnt:
Does not work from OVH server. Services protected by Datadome

PR #240, opened 2021-02-17
Fix/error handling

martinratinaud:
As sometimes the server crashes, this will catch  unhandledRejection and uncaughtException and prevent the site from crashing


clementbiron:
LGTM

PR #264, opened 2021-04-28
Update types.json

AaronjSugarman:
Adding Ad Publishing Policy

### DIFF
--- a/src/app/types.json
+++ b/src/app/types.json
@@ -173,5 +173,11 @@
       "audience": "end user",
       "object": "end user’s live service usage"
     }
-  }
+  },
+  "Ad Publishing Policy": {
+    "commitment": {
+      "writer": "service provider",
+      "audience": "end user",
+      "object": "acceptable and unacceptable usage"
+    }
 }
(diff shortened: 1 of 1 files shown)


src/app/types.json
```diff
@@ -173,5 +173,11 @@
...
       "object": "end user’s live service usage"
     }
-  }
+  },
+  "Ad Publishing Policy": {
+    "commitment": {
+      "writer": "service provider",
+      "audience": "end user",
+      "object": "acceptable and unacceptable usage"
+    }
```
clementbiron, line 182: The json is not valid, there is a } missing at end.

clementbiron (changes requested):

src/app/types.json
AaronjSugarman: Added that extra } !

clementbiron (approved):

src/app/types.json
```diff
@@ -173,5 +173,11 @@
       "audience": "end user",
       "object": "end user’s live service usage"
     }
-  }
```
maxkey12, line 176: },

PR #276, opened 2021-07-21
Send email to corresponding mailing list when change is found

martinratinaud:


clementbiron:
Not tested locally but seems ok.

Issue #294, opened 2021-08-03
OkCupid terms are changing on every request


martinratinaud:
I have been forced to remove OkCupid from the list because it has a behavior which I was not able to bypass

Here is the history of changes: https://github.com/ambanum/OpenTermsArchive-versions/commits/master/OkCupid/Privacy%20Policy.md

You can see that one time it adds the content, and the other one it removes it.

I've made a distinct branch on OpenTermsArchive-snapshots with the raw html code tidied so that changes are more obvious

https://github.com/ambanum/OpenTermsArchive-snapshots/compare/fix/okcupid

You can see that it oscillates between
- an iframe bing moved https://github.com/ambanum/OpenTermsArchive-snapshots/commit/2058a817c84d9962cf1c3600892c8e599ec279e9
- and moved back https://github.com/ambanum/OpenTermsArchive-snapshots/commit/a91c0b9a4daeef1ff88fe3045ad941c50e0cdbb1
- moved https://github.com/ambanum/OpenTermsArchive-snapshots/commit/84d269e565eeb7917ea87cb4d5e18c3dae1c6508
- and moved back https://github.com/ambanum/OpenTermsArchive-snapshots/commit/c5466a638c961fbdc85a9d90888393875b42343d
- and finally some content removed https://github.com/ambanum/OpenTermsArchive-snapshots/commit/6b874a73e8b56e31adc280d23132afe8d1e4fcc8

I have very little idea of how to handle this

@MattiSG @Ndpnt maybe you had already that kind of problems before ?

In the meantime, i will remove the service as it is creating too much clutter on the `versions` repo

martinratinaud:
I created a branch `chore/debug-snapshots` to easily debug on local with prettified snapshots

martinratinaud:
Maybe @xiaoyunyang from @okcupid can help us explain why the privacy policy page is changing from time to time.
That would be the best :-)

MattiSG:
I have no idea why this happens, this seems to be specific to the service provider. I don't believe it is worth investing too much effort in it at that stage.

Just to make it explicit though: I believe it is important that the prettifier for snapshots is not deployed into production 🙂 Snapshots should not be edited in any way, this is an important invariant in the process, that enables us to improve filters a posteriori if needed 🙂

martinratinaud:
@MattiSG I'm truly sorry but the prettifier for snapshots has been on production before I saw this comment :-(

[...]

PR #320, opened 2021-08-23
Automatically create an issue in Github when a document is not available anymore

martinratinaud:


MattiSG:
I would personally be very worried that the repo gets inundated with issues for production stuff that cannot be resolved by anyone through code, only by people who have access to ops, making contribution very hard.

martinratinaud:
@MattiSG 
maybe it's not clear enough but the issues that will be raised after this PR is merged are only if content scraped is not accessible anymore.

So any issue opened through this would be one for fixing a failing service, which is, IMO, what we need to keep the database as updated as possible.

martinratinaud:
I renamed the PR which was not good at all and will put extra attention in the naming of PRs.
I feel a bit alone but need to remember that it's by having clear PRs and issues that we might gain some more contributors.
Thanks!

clementbiron (approved):
LGTM

martinratinaud:
Thanks for the clarification.

I believe this way of keeping track of issues has already benefited to the team as it helped us keeping real track of what's failing (aka issues)

I understand notifications is the big problem and unfortunately, I checked but there is no such feature today as to be notified only for specific tags on github (feature requests exists here though https://github.community/t/feature-request-subscribe-to-notifications-for-specific-labels-github-issues/1606/10)

- for PRs, what do you suggest? Please note that we are 2 developers here working on this. Should we not do PRs at all?
- I have the opposite vision :-) I feel it's good to see a project is living and that I could contribute easily and rapidly. Also it shows how badly we need support. This, of course, is valid if issues are closed often (which we try to do every week).
- I propose we switch these to the "Discussions" tab and this way you can only be notified about stuff in discussions and not in issues. 
 
I believe this would be the way to go as the terms used (issues/discussions) should be used for their named purpose.

I will move what appears like discussions to me to the discussions tab. Please feel free to turn on notifications only for this @MattiSG. And @LVerneyPEReN also as I saw you liked Matti's comment.

Thanks for the help
@clementbiron FYI

MattiSG:
Based on this issue as well as https://github.com/ambanum/OpenTermsArchive/issues/294#issuecomment-908298263 or https://github.com/ambanum/OpenTermsArchive-api/issues/36, I understand that I mostly had a misunderstanding on what is the process for contribution and decision-making 🙂 I had understood that the aim was to build a common governance, but it is now clear that this is a team working publicly, which is great too.

Thank you for having clarified what is the scope of the team 🙂

PR #427, opened 2021-09-08
Add NordVPN

karnauskas:


martinratinaud:
Thank you so much for contributing!

I added `"executeClientScripts": true` which, instead of making a raw http request, will enable puppeteer.

Can you please rebase on the master branch and I will take care of merging it

PR #637, opened 2021-10-28
Add `Eslint` rule to have consistent order import

Ndpnt:


martinratinaud (approved):
Good
Can you please separate the change commit from the formatting commit next time soi that it's easier to understand what has really changed

Thanks

PR #649, opened 2021-11-03
Disallow passing headers object to fetcher

Ndpnt:


services/Instagram.history.json
```diff
@@ -1,52 +1,37 @@
 {
   "Developer Terms": [
     {
-      "fetch": "https://developers.facebook.com/terms",
+      "fetch": "https://developers.facebook.com/terms?locale=en_GB",
       "select": "#faq-page",
-      "headers": {
```
martinratinaud, line 6: This is wrong as you may know. The history should contain all the historical configuration.
But I'm fine with it if you are

martinratinaud (approved):
As discussed before, I still think we will have to have a functionality at some point where it's easy to bypass certain limitations that we may encounter.
passing headers or cookies will at some point be needed IMO.

services/Instagram.history.json
```diff
@@ -1,52 +1,37 @@
 {
   "Developer Terms": [
     {
-      "fetch": "https://developers.facebook.com/terms",
+      "fetch": "https://developers.facebook.com/terms?locale=en_GB",
```
martinratinaud, line 4: Nice!

martinratinaud (approved):

services/Instagram.history.json
Ndpnt: > As discussed before, I still think we will have to have a functionality at some point where it's easy to bypass certain limitations that we may encounter. passing headers or cookies will at some point be needed IMO.

I may agree with that, but currently we do not have the need, so for contribution and code simplicity we remove the feature for now.

Ndpnt:
> As discussed before, I still think we will have to have a functionality at some point where it's easy to bypass certain limitations that we may encounter. passing headers or cookies will at some point be needed IMO.

I may agree with that, but currently we do not have the need, so for contribution and code simplicity we remove the feature for now.

PR #650, opened 2021-11-03
Fix Kraken - Privacy Policy

Ndpnt:


martinratinaud (approved):
Not sure too understand what it fixes to add a trailing "/".

Can you please elaborate?
Also, aren't we supposed to add it to "history"

Ndpnt:
> Not sure too understand what it fixes to add a trailing "/".
> 
> Can you please elaborate? Also, aren't we supposed to add it to "history"

Currently, without the `/`, the server goes in an infinite loop of redirection.

Not necessary to add it to history as it does not impact the rewriting.

PR #655, opened 2021-11-04
Update required node version

Ndpnt:


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "type": "module",
   "engines": {
-    "node": ">=12.17"
+    "node": ">=14.0.0"
```
martinratinaud, line 8: I think it would be safer to bind to the same version as the one in ops.

So maybe instead of latest in ops, we should use 14

martinratinaud (approved):

package.json
Ndpnt: In ops we do this: 
```yaml
- name: Add the NodeSource repository to the system
  shell: curl -sL https://deb.nodesource.com/setup_14.x | sudo bash -

- name: Install NodeJS and Npm
  apt:
    name: nodejs
    update_cache: yes
    state: latest
```
Which already install only the version 14. `latest` only refer to the latest patch of the version 14.

martinratinaud: ok sorry I saw "state: latest" and did not see the above script

PR #675, opened 2021-11-09
Terminate browser on stop

### LINKED ISSUE #658: Terminate all processes on stop
Currently, the headless browser process that is instantiated stays alive.
When the engine is stopped, it must stop all its helpers, including its browser instance.

martinratinaud:
Fixes #658 

This should be work but

- I did not manage to make a test out of it
- I believe process.on can only be used once in the whole code
- on my local, when app crashes, browser is correctly destroyed (nothing in `ps aux | grep chromium`

TBD then


Ndpnt:
There is a mix with `Rename GitHub variables and add documentation` PR content, please remove it.

Ndpnt:
> on my local, when app crashes, browser is correctly destroyed (nothing in ps aux | grep chromium

Yep but the problem is that the `puppeteer` instance prevent app from closing properly when there is no crash.
Do the test on your machine, if you run a single service which have `executeClientScripts: true` in its declaration, the app will never stop. Otherwise, it will stop properly.

Ndpnt:
@martinratinaud I made an alternative implementation without using `process.on`: #677

Ndpnt:
Closed in favor of #677

PR #677, opened 2021-11-09
Terminate headless processes on stop

### LINKED ISSUE #658: Terminate all processes on stop

Ndpnt:
Alternative implementation of #675 
Fixes #658

### DIFF
--- a/scripts/validation/validate.js
+++ b/scripts/validation/validate.js
@@ -8,7 +8,7 @@ import chai from 'chai';
 import config from 'config';
 import jsonSourceMap from 'json-source-map';
 
-import fetch from '../../src/app/fetcher/index.js';
+import fetch, { launchHeadlessBrowser, closeHeadlessBrowser } from '../../src/app/fetcher/index.js';
 import filter from '../../src/app/filter/index.js';
 import * as services from '../../src/app/services/index.js';
 
@@ -49,6 +49,15 @@ let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
         throw new Error(`Could not find any service with id "${serviceId}"`);
       }
 
+      before(async function () {
+        this.timeout(5000);
+        await launchHeadlessBrowser();
+      });
+
+      after(async () => {
+        await closeHeadlessBrowser();
+      });
+
       describe(serviceId, async () => {
         it('has a valid declaration', async () => {
           const declaration = JSON.parse(await fs.readFile(path.join(rootPath, config.get('serviceDeclarationsPath'), `${serviceId}.json`)));
@@ -84,6 +93,8 @@ let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
                   url: location,
                   executeClientScripts,
                   cssSelectors: service.getDocumentDeclaration(type).getCssSelectors(),
+                }, {
+                  keepBrowserAlive: true
                 });
 
                 content = document.content;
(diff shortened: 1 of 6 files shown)


martinratinaud:
This looks good to me and more reliable than my implementation and we should keep yours

Two minor things though
- tests do not pass anymore (timeout problem I think when testing ssl errors)
- it's not clear that options passed to `fetch` are only for puppeteer and are not even passed to  the htmlOnly implementation, maybe we should differentiate them or put the "keepBrowerAlive" by default

TBD

Ndpnt:
> it's not clear that options passed to fetch are only for puppeteer and are not even passed to the htmlOnly implementation, maybe we should differentiate them or put the "keepBrowerAlive" by default.

It's wanted. Consider `fetch` module as a black box, clients of this module do not have to know how it works internally. They pass options to the exposed `fetch` function and inside it we pass where it is needed. For know the only option is `keepBrowerAlive` but if we have option useful for the `HTMLOnly` fetcher we will also pass it the options object.

I hesitated to keep browser alive by default, and maybe we should do this as we do not have a use case for an auto managed browser. I am in favor of completely removing the auto browser management if we have to control headless browser manually by default.

scripts/validation/validate.js
```diff
@@ -49,6 +51,10 @@ let servicesToValidate = args.filter(arg => !arg.startsWith('--'));
         throw new Error(`Could not find any service with id "${serviceId}"`);
       }
 
+      before(() => launchHeadlessBrowser());
+
+      after(() => closeHeadlessBrowser());
```
MattiSG, line 56:
```suggestion
      before(launchHeadlessBrowser);

      after(closeHeadlessBrowser);
```

src/app/fetcher/fullDomFetcher.js
```diff
@@ -61,5 +64,21 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
...
       await page.close();
     }
+
+    if (!keepBrowserAlive && browser) {
+      await browser.close();
+    }
+  }
+}
+
+export async function launchHeadlessBrowser() {
```
MattiSG, line 74: Use a symmetric function name: `open` / `close` or `launch` / `stop` 🙂

MattiSG:
As seen orally: let's just remove the automatic management option, this feature is not part of any public API 🙂

martinratinaud (approved):
LGTM

src/app/fetcher/fullDomFetcher.js
```diff
@@ -63,3 +64,16 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
...
+  if (!browser) {
+    browser = await puppeteer.launch({ headless: true, args: [ '--no-sandbox', '--disable-setuid-sandbox' ] });
+  }
+}
+
+export async function stopHeadlessBrowser() {
+  if (browser) {
+    await browser.close();
+    browser = null;
+  }
```
MattiSG, line 78: > I like early exits 🙂 
```suggestion
  if (! browser) return;
  
  await browser.close();
  browser = null;
```

src/app/fetcher/fullDomFetcher.js
```diff
@@ -63,3 +64,16 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
     }
   }
 }
+
+export async function launchHeadlessBrowser() {
+  if (!browser) {
+    browser = await puppeteer.launch({ headless: true, args: [ '--no-sandbox', '--disable-setuid-sandbox' ] });
```
MattiSG, line 70: > Multiline options would increase readability and ease diff if it changes in the future 🙂

MattiSG (approved):
I realize I had not fully understood that we had arbitrated for `fetch`, an abstraction layer, to expose the headless browser instance control. This seems a bit much to expose to the clients of this module, but this is not blocking 👍

PR #686, opened 2021-11-12
Do not stop the whole process on unfetchable documents

### LINKED ISSUE #657: Do not stop the whole process on unfetchable documents
Document in the CONTRIBUTING the fact that errors should be fatal if they are not expected.

This explains the architecture illustrated in #651.

Ndpnt:
Fixes #657.

Also document how we handle errors.


CONTRIBUTING.md
```diff
@@ -407,3 +411,37 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
...
 
 - Do not rely on GitHub issue reference numbers in commit messages, as we have no guarantee the host system and its autolinking will be stable in time. Make sure the context is self-explanatory. If an external reference is given, use its full URL.
+
+## Practices
+
+### Errors handling
+
+First of all it's important to distinguish two fundamentally different kinds of errors: **operational** and **programmer** errors.
+
+- **Operational errors** represent run-time problems experienced by **correctly-written programs**. These are not bugs in the program. These are usually problems with something else: the system itself (e.g., out of memory or too many open files), the system’s configuration (e.g., no route to a remote host), the network (e.g., socket hang-up), or a remote service (e.g., a 500 error, failure to connect, or the like).
```
MattiSG, line 421:
```suggestion
- **Operational errors** represent run-time problems experienced by **correctly-written programs**. These are not bugs in the program. These are usually problems with something else: the system itself (e.g. out of memory or too many open files), the system’s configuration (e.g. no route to a remote host), the network (e.g. socket hang-up), or a remote service (e.g. a 500 error, failure to connect, or the like).
```

CONTRIBUTING.md
```diff
@@ -407,3 +411,37 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
...
+
+Also noteworthy, failure to handle an operational error is itself a programmer error.
+
+#### Handling operational errors
+
+There are five ways to handle operational errors:
+- **Deal with the failure directly**. For example, create directory if it's missing.
+- **Propagate the failure**. If you don’t know how to deal with the error, the simplest thing to do is to abort whatever operation you’re trying to do, clean up whatever you’ve started, and propagate the error.
+- **Retry the operation**. For example, try to reconnect if the connection is lost.
+- **Blow up**. If the error cannot be handled and can affect data integrity.
```
MattiSG, line 435: “Blow up” is not clear.

CONTRIBUTING.md
```diff
@@ -407,3 +411,37 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
...
+- **Log the error — and do nothing else**. If it's a minor error and there’s nothing you can do about, and there is no reason to stop the whole process.
+
+In our case, we consider all `fetch`-related errors as expected, so as operational errors and we handle them by logging but we do not stop the whole process. We handle errors related to the `notifier` in the same way.
+In contrast, we consider errors from the `history` module as fatal, and we crash immediately.
+
+#### Handling programmer errors
+
+**The best way to recover from programmer errors is to crash immediately.** Indeed, it is not recommended to attempt to recover from programmer errors — that is, allow the current operation to fail, but keep handling requests. Consider that a programmer error is a case that you didn’t think about when you wrote the original code. How can you be sure that the problem won’t affect the program itself and the data integrity?
```
MattiSG, line 443:
```suggestion
**The best way to handle programmer errors is to crash immediately.** Indeed, it is not recommended to attempt to recover from programmer errors — that is, allow the current operation to fail, but keep handling requests. Consider that a programmer error is a case that you didn’t think about when you wrote the original code. How can you be sure that the problem won’t affect the program itself and the data integrity?
```

CONTRIBUTING.md
```diff
@@ -407,3 +411,37 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
...
+
+So, **_shut the process gracefully when a stranger comes to town_**!
```
MattiSG, line 445:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -407,3 +411,37 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
...
+
+More info [about error handlings](https://www.joyent.com/node-js/production/design/errors).
```
MattiSG, line 447: Make source more clear: “This section is highly inspired, and in part extracted, from …”

src/logger/index.js
```diff
@@ -88,7 +88,7 @@ logger.onRecordsPublished = () => {
 };
 
 logger.onInaccessibleContent = (error, serviceId, type) => {
-  logger.warn({ message: `Content inaccessible: ${error.message}`, serviceId, type });
+  logger.warn({ message: `${error.message}`, serviceId, type });
```
MattiSG, line 91:
```suggestion
  logger.warn({ message: error.message, serviceId, type });
```

src/app/fetcher/htmlOnlyFetcher.js
```diff
@@ -3,7 +3,7 @@ import HttpProxyAgent from 'http-proxy-agent';
 import HttpsProxyAgent from 'https-proxy-agent';
 import nodeFetch, { AbortError } from 'node-fetch';
 
-import { InaccessibleContentError } from '../errors.js';
+import { FetchError } from './errors.js';
```
MattiSG, line 6: This name is very technical. What about `InaccessibleDocumentError`? 🙂

src/app/index.js
```diff
@@ -61,9 +61,8 @@ export default class CGUs extends events.EventEmitter {
         return;
       }
 
-      this.emit('error', error, service.id, type);
```
MattiSG, line 64: I'm not sure about that 😕 
(discussed orally)

MattiSG (commented):
I would request changes even for a few minor things, but I won't be available in the next two weeks and don't want to block 😉

CONTRIBUTING.md
martinratinaud: Maybe `Fail` or `Crash`

src/app/fetcher/htmlOnlyFetcher.js
martinratinaud: I believe t's clearer if it's a `FetchError`

src/app/fetcher/htmlOnlyFetcher.js
```diff
@@ -24,26 +24,28 @@ export default async function fetch(url) {
...
+  let response;
 
-    const mimeType = response.headers.get('content-type');
-
-    return {
-      mimeType,
-      content: await (mimeType.startsWith('text/') ? response.text() : response.buffer()),
-    };
+  try {
+    response = await nodeFetch(url, options);
```
martinratinaud, line 30: as we are using a user agent on puppeteer, maybe add it it here also

src/app/index.js
```diff
@@ -95,14 +94,25 @@ export default class CGUs extends events.EventEmitter {
...
+      if (error instanceof FetchError) {
+        throw new InaccessibleContentError(error.message);
+      }
+
+      throw error;
+    }
 
     if (!content) {
-      return;
+      throw new InaccessibleContentError('Content is empty');
```
martinratinaud, line 115: 👍

martinratinaud (approved):
LGTM

src/app/fetcher/htmlOnlyFetcher.js
Ndpnt: Why? We do not have issues with the `htmlOnlyFetcher`?
In any case, if we want to do that, we will do it in another PR.

PR #695, opened 2021-11-16
Improve Eslint configuration

### LINKED ISSUE #663: Handle ESlint warnings with `.eslintrc` setup
We have “unnamed functions” warnings in `test` and “extraneous dependency” warnings in `scripts`. These warnings are not relevant in the context of their folder.
Since `.eslintrc` setup files are hierarchical, they should be defined at folder-level rather than within line-level comments.

Ndpnt:
Fixes #663 

- Update `object-curly-newline` and `comma-dangle` rules.
- Remove warning on tests files
- Remove errors on scripts files

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -21,7 +21,9 @@ rules:
       singleValue: false
   eqeqeq: 0
   chai-friendly/no-unused-expressions: 2
-  comma-dangle: 0
+  comma-dangle:
+    - error
+    - always-multiline
   consistent-return: 0
   function-paren-newline:
     - error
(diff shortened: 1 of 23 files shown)


.eslintrc.yaml
```diff
@@ -21,7 +21,9 @@ rules:
       singleValue: false
   eqeqeq: 0
   chai-friendly/no-unused-expressions: 2
-  comma-dangle: 0
+  comma-dangle:
+    - error
+    - always-multiline
```
martinratinaud, line 26: I approve and merge so that we do not get too many files conflict 

But I want to rediscuss `always-multiline` which I find pretty weird but fine

martinratinaud (approved):

PR #698, opened 2021-11-18
Unify snapshot recording

### LINKED ISSUE #683: Unify snapshot recording behavior across the `executeClientScripts` option
Currently, if the `executeClientScripts` option is set, the `fullDomFetcher` will wait for the selected elements to be loaded on page before saving a snapshot. This means that if the element is never found, the snapshot is never recorded. This is not the case for the default fetcher.

These behaviors should be unified: a snapshot should always be saved.

martinratinaud:
Fixes #683


src/app/fetcher/fullDomFetcher.js
```diff
@@ -42,9 +38,24 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
       throw new FetchDocumentError(`Received HTTP code ${statusCode} when trying to fetch '${url}'`);
     }
 
-    await Promise.all(selectors.map(selector => page.waitForSelector(selector, { timeout: config.get('fetcher.waitForElementsTimeout') })));
+    const mimeType = (response.headers() || [])['content-type'];
+
+    if (mimeType.startsWith('text/')) {
+      content = await response.text();
```
martinratinaud, line 44: I reproduced the behavior that was existent on the htmlOnlyFetcher.

`page.content` will return a blank html page even if content is empty
`response.text` returns the exact data retrieved

src/app/index.js
```diff
@@ -101,45 +101,38 @@ export default class CGUs extends events.EventEmitter {
...
     try {
-      ({ mimeType, content } = await fetch({
+      const { mimeType, content } = await fetch({
         url: location,
         executeClientScripts,
         cssSelectors: documentDeclaration.getCssSelectors(),
-      }));
+      });
+
+      const snapshotId = await this.recordSnapshot({
```
Ndpnt, line 111: Why moving this in the `try/catch`?

src/app/fetcher/htmlOnlyFetcher.js
```diff
@@ -35,9 +35,15 @@ export default async function fetch(url) {
 
     const mimeType = response.headers.get('content-type');
 
+    const content = await (mimeType.startsWith('text/') ? response.text() : response.buffer());
+
+    if (!content) {
```
Ndpnt, line 40: Why moving this inside the `fetch` module rather than leaving it to the caller to decide whether they consider it a error or not?

src/app/fetcher/fullDomFetcher.js
Ndpnt: I'm not sure to understand, TBD.

src/app/fetcher/fullDomFetcher.js
```diff
@@ -42,9 +38,24 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
...
+    } else {
+      try {
+        await Promise.all(selectors.map(selector => page.waitForSelector(selector, { timeout: config.get('fetcher.waitForElementsTimeout') })));
+        content = await page.content();
+      } catch (e) {
+        // Selectors have not been found in the page
+        // Return the complete snapshot that have been retrieved
+        content = await response.text();
```
Ndpnt, line 52: I think we should only do this when the error is effectively a `waitForElementsTimeout` error like this:

```
if (error instanceof puppeteer.pptr.errors.TimeoutError) { // Expected elements are not present on the web page
    content = await page.content(); // Get the whole page content as content selectors may be wrong
}
```

martinratinaud (commented):
@Ndpnt I moved all tests from specific files to common file and harmonized all.

Still one question though.
In case a pdf file is processed with "executeClientScripts": true, it will break but I see that it is handled by the json schema.

I believe this is enough but wanted to make sure it is.

src/app/fetcher/fullDomFetcher.js
```diff
@@ -42,13 +37,21 @@ export default async function fetch(url, cssSelectors, { retry } = { retry: 0 })
...
     }
 
-    await Promise.all(selectors.map(selector => page.waitForSelector(selector, { timeout: config.get('fetcher.waitForElementsTimeout') })));
-
-    content = await page.content();
+    try {
+      await Promise.all(selectors.map(selector => page.waitForSelector(selector, { timeout: config.get('fetcher.waitForElementsTimeout') })));
+    } catch (error) {
+      if (error instanceof puppeteer.pptr.errors.TimeoutError) {
+        // Expected elements are not present on the web page
```
Ndpnt, line 44: Add this to the `eslintrc.yaml` tp ensure multiline comment consistency: 
```
  multiline-comment-style:
    - error
    - separate-lines
```

Ndpnt (approved):

PR #705, opened 2021-11-18
Document why parallelism is deactivated

### LINKED ISSUE #664: Document why parallelism is deactivated
The parallel handling feature is currently set to a parallelism of 1.
The main reason is that too many requests on the same endpoint yield 403, and that sometimes `simple-git` creates a commit without returning an SHA for unknown reasons.
This reason should be documented in the code to avoid it being turned on too eagerly.

Ndpnt:
Fixes #664.

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -9,6 +9,10 @@ import filter from './filter/index.js';
 import * as history from './history/index.js';
 import * as services from './services/index.js';
 
+// The parallel handling feature is currently set to a parallelism of 1 on document tracking
+// because when it's higher there are two issues:
+// - too many requests on the same endpoint yield 403
+// - sometimes when creating a commit no SHA are returned for unknown reasons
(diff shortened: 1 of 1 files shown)


martinratinaud (changes requested):
I'm sorry but I do not get what happens

This PR comments the fact that it stays on 1
And the other PR puts a parallelism of 10 but the comment is not present

Can you please elaborate a bit more?

Ndpnt:
The comment added on this PR is only for the `MAX_PARALLEL_DOCUMENTS_TRACKS` variable which is associated with the "document tracking" step. 
On the PR #706, the parallelism is enabled only for the `MAX_PARALLEL_REFILTERS` variable (so the refiltering step) as issues we have with parallelism happens only for document tracking.

martinratinaud (approved):

PR #708, opened 2021-11-25
Extract services in OpenTermsArchive/services-all

### LINKED ISSUE #684: Move services declarations to separate repository
- [x] The `services` folder is moved to a repository `OpenTermsArchive/services-all`.
- [x] Its contents are stored in a `declarations` folder at the root of the repo.
- [x] Move document types declaration into `document-types.json` at the root of the repository.
- [x] Dependencies are declared in a `package.json`.
- [x] The `package.json` has no version and name metadata.
- [x] OTA-Core loads services declarations from a folder given in config.
- [x] Ops loads the services in production.
- [x] Rewrite filters to be independent from OTA-Core API.
- [x] Remove the labels `fix-document` and `add-document`.
- [ ] ~Move validation scripts into the new repository. This implies adding `https://github.com/ambanum/OpenTermsArchive` as a dependency to `OpenTermsArchive/services-all`.~
- [x] Move CI setup to validate service declarations.
- [x] Validate systematically all services in CI, not only those changed.
- [x] Apply a EUPL LICENSE.
- [x] Add a README explaining how this repository is used.
- [x] Move the relevant CONTRIBUTING sections of OTA-Core into Services-All.
- [x] Move the relevant wiki sections of OTA-Core into a `Guidelines.md` file at the root of Services-All.

### Accepted limitations

- OTA-Core is not versioned at this stage. In case of API and document declaration format changes, manual synchronisation between services declaration and Core will have to be handled. Proper semver will be introduced later, for the moment we prefer to move fast without handling versioning.

Ndpnt:
With https://github.com/OpenTermsArchive/services-all/pull/98, fixes #684.


bin/validate.js
```diff
@@ -0,0 +1,26 @@
...
+  mocha.addFile(VALIDATE_PATH);
+  await mocha.loadFilesAsync();
+
+  let hasFailedTests = false;
+
+  mocha.run()
+    .on('fail', () => { hasFailedTests = true; })
+    .on('end', () => {
+      if (hasFailedTests) {
+        process.exit(1);
```
MattiSG, line 23: Prefer [setting the exit code](https://nodejs.org/api/all.html#process_processexitcode_1) instead of forcing immediately exit to ensure any async process closes properly 🙂 

```suggestion
        process.exitCode = 1;
```

ops/roles/ota/tasks/main.yml
```diff
@@ -25,6 +25,18 @@
   tags:
     - setup
 
+- name: Clone services declarations Repo
```
MattiSG, line 28:
```suggestion
- name: Install services declarations
```

src/app/fetcher/fullDomFetcher.js
```diff
@@ -8,7 +8,7 @@ import { FetchDocumentError } from './errors.js';
 puppeteer.use(StealthPlugin());
 
 const PUPPETEER_TIMEOUT = 30 * 1000; // 30 seconds in ms
-
+const WAIT_FOR_ELEMENTS_TIMEOUT = config.has('fetcher.waitForElementsTimeout') ? config.get('fetcher.waitForElementsTimeout') : 10 * 1000; // 10 seconds in ms
```
MattiSG, line 11: Why do we set a default value here instead of relying on defaults being set at config level?

src/app/services/index.js
```diff
@@ -192,24 +187,36 @@ async function loadServiceHistoryFiles(serviceId) {
...
-    .map(filePath => path.basename(filePath, '.json'));
+export function getServicesDeclarationsPath() {
+  let serviceDeclarationsPath = path.resolve(
+    process.cwd(),
+    './declarations',
+  );
+
+  if (config.has('services.declarationsPath')) {
+    serviceDeclarationsPath = path.resolve(
+      process.cwd(),
```
MattiSG, line 198: Relying on `cwd` seems very brittle to me. It is also ambiguous for the `services.declarationsPath`: it means it has to be given relative to `cwd`, whereas I would rather expect relative to the package base directory.

MattiSG (changes requested):
Let's have a synchronous discussion on config keys and the usage of `cwd` 🙂

scripts/validation/validate.js
```diff
@@ -27,16 +27,13 @@ if (args.includes('--schema-only')) {
...
 let servicesToValidate = args;
 
 (async () => {
   const SERVICE_DECLARATIONS_PATH = services.getServicesDeclarationsPath();
   const serviceDeclarations = await services.loadWithHistory();
 
-  if (modifiedOnly) {
-    servicesToValidate = await services.getIdsOfModified();
-  } else if (!servicesToValidate.length) {
+  if (!servicesToValidate.length) {
```
martinratinaud, line 36: This should be replaced with the possibility to validate one service only

Like "npm run validate Amazon"

If we do not have this, it will be very difficult to add a service and validate it without running the whole test suite

martinratinaud (approved):
Very nice and smooth implementation, congratulations

README.md
```diff
@@ -165,7 +172,10 @@ The default configuration can be read and changed in `config/default.json`.
 
 ```json
 {
-  "serviceDeclarationsPath": "Directory containing services declarations and associated filters.",
+  "service": {
```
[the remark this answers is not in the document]
Ndpnt: It is like the module name. I renamed it to be consistent with all others keys. They all are the name of a module where the config apply.

README.md
```diff
@@ -165,7 +172,10 @@ The default configuration can be read and changed in `config/default.json`.
...
+    "declarationsPath": "Directory containing services declarations and associated filters.",
+    "documentTypesPath": "File containing document types."
```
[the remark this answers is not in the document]
Ndpnt: Because it is used in the `service` module. Maybe we should rename the module if it brings too much confusion?

bin/validate.js
Ndpnt: 👍

src/app/fetcher/fullDomFetcher.js
Ndpnt: Yes, I knew that this point would be discussed :).
I did that to avoid the consumer of the command validate outside this repo, like the repo `service-all`, to have to define a config. Let's discuss about that.

test/fixtures/termsFromPDF.md
```diff
@@ -1,12 +1,7 @@
------ page 1 -----
+THIS IS A LEGAL AGREEMENT BETWEEN YOU AND TESTCOMPANYTESTCOMPANY Developer Agreement
```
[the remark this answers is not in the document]
Ndpnt: Because I had to upgrade the dependencies which convert PDF to markdown, and they change the way they convert.

bin/validate.js
```diff
@@ -0,0 +1,26 @@
...
+import { fileURLToPath } from 'url';
+
+import Mocha from 'mocha';
+
+const mocha = new Mocha({});
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const VALIDATE_PATH = path.resolve(__dirname, '../scripts/validation/validate.js');
+
+(async () => {
+  mocha.delay();
```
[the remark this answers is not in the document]
Ndpnt: Totally right! 👍

scripts/validation/validate.js
Ndpnt: It's already the case :wink:

test/fixtures/termsFromPDF.md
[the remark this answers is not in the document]
MattiSG: @martinratinaud see https://github.com/ambanum/OpenTermsArchive/pull/708/commits/dec5f24e413f2e4341e503e5dc393e2a1a1c0e6b 🙂

README.md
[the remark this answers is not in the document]
Ndpnt: Handled synchronously

README.md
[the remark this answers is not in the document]
Ndpnt: Handled synchronously

bin/validate.js
```diff
@@ -0,0 +1,41 @@
+#! /usr/bin/env node
+import './.env.js'; // Workaround to ensure `SUPPRESS_NO_CONFIG_WARNING` is set before config is imported
```
MattiSG, line 2: > Why can't we just assign `process.env` in this file? 😮

src/archivist/services/index.js
```diff
@@ -3,18 +3,16 @@ import path from 'path';
...
+
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const SERVICE_DECLARATIONS_PATH = path.resolve(
-  __dirname,
-  '../../..',
-  config.get('serviceDeclarationsPath'),
-);
+
+export const SERVICE_DECLARATIONS_PATH = path.resolve(__dirname, '../../..', config.get('services.declarationsPath'));
+export const DOCUMENT_TYPES_PATH = path.resolve(__dirname, '../../..', config.get('services.documentTypesPath'));
```
MattiSG, line 15: As those are exported and rely on config, I feel a bit weird reading them as all-caps constants.
```suggestion
export const declarationsPath = path.resolve(__dirname, '../../..', config.get('services.declarationsPath'));
export const documentTypesPath = path.resolve(__dirname, '../../..', config.get('services.documentTypesPath'));
```

MattiSG (approved):
👍

bin/validate.js
Ndpnt: Is the comment not explicit enough?

MattiSG: I mean I understand we need a workaround, I just don't understand why it is needed. I'm fine with that level of documentation in the code, I'm just surprised!

Ndpnt: It's because all imports are evaluated prior to the body of module doing the importing. So even if we write `process.env.SUPPRESS_NO_CONFIG_WARNING = 'y'` before the `import config`, the instruction is evaluated after, so `config` will still warn.

PR #713, opened 2021-12-02
GitHub put back cache

martinratinaud:
In order to prevent API rate limits as search quota limits are very low


martinratinaud:
Closed in favor of https://github.com/ambanum/OpenTermsArchive/pull/717

PR #722, opened 2021-12-09
Support MongoDB as a storage mechanism

Ndpnt:
Important changes:
- Remove setup script. Now git adapter will automatically create repository if it not exists.
- Versions date will now be the snapshot fetch date.
- Add `startRefilterRun`, `refilterRunCompleted`, `startTrackingChangesRun`, `trackingChangesRunCompleted` events and remove `recordsPublished` event.
- `history` module does not exits anymore, it's replaced by `recorder`
- Rewrite snapshots script cannot add extra content to git message anymore as it was only useful when importing TosBack history.


Ndpnt:
Note: `adapter` has a specific meaning in term of Design pattern and it doesn't seem to fit our case. I think the pattern here is rather the Repository pattern. In this case, we should maybe change the terms we use… 
What do you think?

src/archivist/index.js
```diff
@@ -22,7 +22,10 @@ export const AVAILABLE_EVENTS = [
   'versionRecorded',
   'firstVersionRecorded',
   'versionNotChanged',
-  'recordsPublished',
+  'startRefilterRun',
+  'refilterRunCompleted',
```
martinratinaud, line 26: At first sight it looked weird to me to not have a `start` and `end`
I understand you've thought this through but i'd like to undestand why you picked this instead of
`refilterRunStart` or `refilterRunStarted` (to keep the verb at the end)

src/archivist/index.js
```diff
@@ -95,17 +98,22 @@ export default class Archivist extends events.EventEmitter {
     });
   }
 
-  async trackChanges(servicesIds) {
+  async trackChanges(servicesIds = this.serviceIds) {
+    servicesIds.sort((a, b) => a.localeCompare(b)); // Sort service ids by lowercase name to have more intuitive logs
```
martinratinaud, line 102: 👍  but maybe we should shuffle them

src/storage-adapters/mongo/index.js
```diff
@@ -0,0 +1,91 @@
...
+  async record({ serviceId, documentType, content: passedContent, mimeType, fetchDate, isRefilter, snapshotId }) {
+    let content = passedContent;
+
+    if (mimeType == PDF_MIME_TYPE) {
+      content = passedContent.toString('utf-8'); // Serialize PDF
+    }
+
+    const previousRecord = await this.getLatestRecord(serviceId, documentType);
+
+    if (previousRecord.id && previousRecord.content == content) {
```
martinratinaud, line 34: `===` ?

src/storage-adapters/mongo/index.js
```diff
@@ -0,0 +1,91 @@
...
+    const insertResult = await this.collection.insertOne({ ...recordProperties, created_at: new Date() });
+
+    return {
+      id: insertResult.insertedId.toString(),
+      isFirstRecord,
+    };
+  }
+
+  async getLatestRecord(serviceId, documentType) {
+    const [record] = await this.collection.find({ serviceId, documentType }).limit(1).sort({ fetchDate: -1 }).toArray();
```
martinratinaud, line 68: I think findOne would work and be more easy to read

`const record = await this.collection.findOne({ serviceId, documentType }).sort({ fetchDate: -1 })`

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+    collection = db.collection(config.get('recorder.snapshots.storage.mongo').collection);
+  });
+
+  describe('#record', () => {
+    let record;
+    let mongoRecord;
+    let numberOfRecordsBefore;
+    let numberOfRecordsAfter;
+
+    context('when it is the first record', () => {
```
martinratinaud, line 48: **W**hen ? to be consistent?

src/storage-adapters/git/git.js
```diff
@@ -1,21 +1,26 @@
...
     this.path = repositoryPath;
-    this.git = simpleGit(repositoryPath, { maxConcurrentProcesses: 1 });
+    this.author = author;
   }
 
-  async init() {
-    return this.git.init();
-  }
+  async initialize() {
+    if (!await fileExists(this.path)) {
```
martinratinaud, line 17: As it's not a file but a directory, maybe using `existsSync` would be better

martinratinaud (commented):
This looks good to me, congratulations.

I still need to launch it and try on my local but already made some comments.

Will approve after testing

scripts/rewrite/rewrite-snapshots.js
```diff
@@ -66,13 +82,12 @@ let history;
...
     ({ serviceId, documentType } = renamer.applyRules(serviceId, documentType));
 
-    const { id: snapshotId } = await history.recordSnapshot({
+    const { id: snapshotId } = await recorder.recordSnapshot({
       serviceId,
       documentType,
       content,
       mimeType,
-      authorDate: commit.date,
-      extraChangelogContent: commit.body,
```
MattiSG, line 75: > We could have a generic, “storage adapter”-level metadata storage mechanism that would store them as properties in MongoDB and as a text list in Git commit messages.
> But there is no use case for that at the moment 😀

src/archivist/index.js
```diff
@@ -23,7 +22,10 @@ export const AVAILABLE_EVENTS = [
   'versionRecorded',
   'firstVersionRecorded',
   'versionNotChanged',
-  'recordsPublished',
+  'startRefilterRun',
+  'refilterRunCompleted',
+  'startTrackingChangesRun',
+  'trackingChangesRunCompleted',
```
MattiSG, line 28: Unify with other forms of events?
```suggestion
  'refilteringStarted',
  'refilteringCompleted',
  'trackingStarted',
  'trackingCompleted',
```

src/archivist/index.js
```diff
@@ -90,16 +98,22 @@ export default class Archivist extends events.EventEmitter {
...
   }
 
-  async trackChanges(servicesIds) {
+  async trackChanges(servicesIds = this.serviceIds) {
+    servicesIds.sort((a, b) => a.localeCompare(b)); // Sort service ids by lowercase name to have more intuitive logs
+
+    this.emit('startTrackingChangesRun', servicesIds.length, this.getNumberOfDocuments(servicesIds));
+
     await launchHeadlessBrowser();
+    await this.recorder.initialize();
```
MattiSG, line 107: > These could be parallelized for improved performance.

test/data/README.md
```diff
@@ -1 +1 @@
-This directory does not contain actual data, but it should be left in place. It is here to support testing.
+This directory, like `versions` and `snapshots` directories, does not contain actual data, but it should be left in place. It is here to support testing.
```
MattiSG, line 1:
```suggestion
This directory, like the `versions` and `snapshots` directories, does not contain actual data, but it should be left in place. It is here to support testing.
```

test/data/snapshots/README.md
```diff
@@ -1 +0,0 @@
-This directory does not contain actual data, but it should be left in place. It is here to support testing.
```
MattiSG, line 1: I understand that this file is not necessary anymore due to the way `initialize` now works, creating the folder hierarchy as needed, whereas it used to be mandatory for ensuring presence of the folders.
If that is indeed the case, then do we still need to mention those files in `data/README.md`? Do we still need the `test/data/README.md` altogether?

README.md
```diff
@@ -157,14 +157,6 @@ cd ..
 git clone https://github.com/OpenTermsArchive/services-all.git
 ```
 
-### Setting up the database
-
-Initialize the database:
-
-```sh
-npm run setup
```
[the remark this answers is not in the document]
Ndpnt: Yep but I think this option is not relevant anymore. 
With MongoDB, I'm think we do not want to create a script to clone the Mongo database from our server to a local database. Moreover as our snapshots repository is now very heavy and slow, I do not find it a nice option. If the user really wants a copy of the official database, he still could clone it manually.

ops/roles/infra/tasks/mongo.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Import the MongoDB public key used by the package management system
+  shell: wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | apt-key add -
```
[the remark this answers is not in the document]
Ndpnt: I think because it's not available for debian. I will check that.

src/archivist/index.js
Ndpnt: You're right. It's better with `refilterRunStarted`.

src/archivist/index.js
Ndpnt: Yep but not in this PR

src/storage-adapters/mongo/index.js
Ndpnt: Not really useful but we can if you prefer

src/storage-adapters/mongo/index.js
Ndpnt: Are you sure it works like that? I think I tried and it did not worked…

src/storage-adapters/mongo/index.test.js
Ndpnt: Nope, at the contrary, I try to remove all caps!

src/storage-adapters/git/git.js
Ndpnt: I though it was deprecated but `exists` is, not `existsSync`. So :+1:

scripts/rewrite/rewrite-snapshots.js
Ndpnt: I had the same idea and came to the same conclusion! :D

src/archivist/index.js
Ndpnt: 👍

README.md
```diff
@@ -216,15 +212,42 @@ An example of a production configuration file can be found in `config/production
 }
 ```
 
-- Public URLs to the snapshots and versions repositories, used to automate the initial database setup (`publicSnapshotsRepository` and `publicVersionsRepository` which are used by `npm setup`) and for the links in notifications (`snapshotsBaseUrl`):
+##### Storage adapters configuration
+
+Two storage adapters are currently supported: Git and MongoDB. Each one can be used independently for versions and snapshots.
```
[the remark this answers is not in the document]
Ndpnt: > Note: `adapter` has a specific meaning in term of Design pattern and it doesn't seem to fit our case. I think the pattern here is rather the Repository pattern. In this case, we should maybe change the term we use… What do you think?

https://en.wikipedia.org/wiki/Data_mapper_pattern

What about `persistence-mappers` where `git` and `mongo` are mappers?

Ndpnt:
> Note: `adapter` has a specific meaning in term of Design pattern and it doesn't seem to fit our case. I think the pattern here is rather the Repository pattern. In this case, we should maybe change the term we use… What do you think?

https://en.wikipedia.org/wiki/Data_mapper_pattern

What about persistence-mappers where git and mongo are mappers?

ops/roles/infra/tasks/mongo.yml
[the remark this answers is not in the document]
Ndpnt: Updated

README.md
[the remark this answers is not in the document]
Ndpnt: Also remove related dependency

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+      ...config.get('recorder.snapshots.storage.git'),
+      path: RECORDER_PATH,
+      fileExtension: 'html',
+    });
+
+    return subject.initialize();
+  });
+
+  describe('#_save', () => {
+    context('when service’s directory already exist', () => {
```
MattiSG, line 55:
```suggestion
    context('when service’s directory already exists', () => {
```

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+      before(async () => subject._save({
+        serviceId: SERVICE_PROVIDER_ID,
+        documentType: DOCUMENT_TYPE,
+        content: CONTENT,
+      }));
+
+      after(async () => subject._removeAllRecords());
+
+      it('creates a file for the given service', async () => {
+        expect(fs.readFileSync(EXPECTED_FILE_PATH, { encoding: 'utf8' })).to.equal(CONTENT);
```
MattiSG, line 65: Why is this `async` yet `readFileSync`? 🙂

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+      it('creates a directory and file for the given service', async () => {
+        await subject._save({ serviceId: NEW_SERVICE_ID, documentType: DOCUMENT_TYPE, content: CONTENT });
+
+        expect(fs.readFileSync(NEW_SERVICE_EXPECTED_FILE_PATH, { encoding: 'utf8' })).to.equal(CONTENT);
+      });
+    });
+  });
+
+  describe('#_commit', () => {
+    const commitMessage = 'Message to check if the commit message is properly saved';
```
MattiSG, line 103:
```suggestion
    const COMMIT_MESSAGE = 'Message to check if the commit message is properly saved';
```

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+      it('returns the file path with given extension for the given service provider’s document type', async () => {
+        expect(subject._getPathFor(SERVICE_PROVIDER_ID, DOCUMENT_TYPE, 'pdf')).to.equal(EXPECTED_PDF_FILE_PATH);
+      });
+    });
+  });
+
+  describe('#_isTracked', () => {
+    after(async () => subject._removeAllRecords());
+
+    context('when the file does not exists', () => {
```
MattiSG, line 147:
```suggestion
    context('when the file does not exist', () => {
```

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+      });
+
+      context('when provided', () => {
+        it('stores the fetch date', () => {
+          expect(new Date(commit.date).getTime()).to.equal(FETCH_DATE.getTime());
+        });
+
+        it('stores the mime type', () => {
+          expect(mime.getType(EXPECTED_FILE_PATH)).to.equal(MIME_TYPE);
+        });
```
MattiSG, line 230: This is made mandatory in checks, it should not be in a “when provided” block.

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+        expect(latestRecord.id).to.include(lastSnapshotId);
+      });
+
+      it('returns the latest record content', async () => {
+        expect(latestRecord.content.toString('utf8')).to.equal(UPDATED_FILE_CONTENT);
+      });
+
+      it('returns the latest record mime type', async () => {
+        expect(latestRecord.mimeType).to.equal(MIME_TYPE);
+      });
```
MattiSG, line 509: None of these comparisons seem to be `async` 🙂

src/storage-adapters/git/index.js
```diff
@@ -0,0 +1,159 @@
+/**
+ * This file is the boundary beyond which the usage of git is abstracted.
```
MattiSG, line 2: 😍

src/archivist/recorder/index.test.js
```diff
@@ -0,0 +1,509 @@
...
+            mimeType: 'mime type',
+          };
+
+          Object.entries(validParams).forEach(([testedRequiredParam]) => {
+            context(`when "${testedRequiredParam}" is missing`, () => {
+              it('throws an error', async () => {
+                try {
+                  const validParamsExceptTheOneTested = Object.fromEntries(Object.entries(validParams).filter(([paramName]) => paramName != testedRequiredParam));
+
+                  await recorder.recordSnapshot(validParamsExceptTheOneTested);
```
MattiSG, line 93: > This is smart but complex.

src/archivist/recorder/index.test.js
```diff
@@ -0,0 +1,509 @@
...
+                } catch (e) {
+                  expect(e).to.be.an('error');
```
MattiSG, line 95: What about this instead of try/catch? 🙂 

```js
expect(() => recorder.recordSnapshot(validParamsExceptTheOneTested)).to.throw(testedRequiredParam);

File: src/archivist/recorder/index.js [hunk 24]
```diff
@@ -0,0 +1,86 @@
+export default class Recorder {
+  constructor({ versionsStorageAdapter, snapshotsStorageAdapter }) {
+    if (!versionsStorageAdapter && !snapshotsStorageAdapter) {
+      throw new RangeError('storage adapters should be defined both for versions and snapshots');
```
MattiSG, line 4: This is not what is currently checked.
```suggestion
    if (!versionsStorageAdapter || !snapshotsStorageAdapter) {
      throw new RangeError('Storage adapters should be defined both for versions and snapshots');
```

src/archivist/recorder/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+    this.versionsStorageAdapter = versionsStorageAdapter;
+    this.snapshotsStorageAdapter = snapshotsStorageAdapter;
+  }
+
+  async initialize() {
+    return Promise.all([ this.versionsStorageAdapter, this.snapshotsStorageAdapter ].map(storageAdapter => {
+      if (typeof storageAdapter.initialize != 'function') {
+        return Promise.resolve();
+      }
```
MattiSG, line 15: I am not sure this is useful: we expect an adapter to comply with an API that includes `initialize`, that's all. If we want to be smart, it should be by providing an abstract Storage Adapter class that implements a no-op for both `initialize` and `finalize`, not in clients.

src/index.js
```diff
@@ -6,40 +10,47 @@ import Archivist from './archivist/index.js';
...
-    const isServiceDeclared = archivist.serviceDeclarations[serviceId];
+  if (extraArgs.length) {
+    serviceIds = extraArgs.filter(serviceId => {
+      const isServiceDeclared = archivist.serviceDeclarations[serviceId];
 
-    if (!isServiceDeclared) {
-      logger.warn(`Service ${serviceId} does not exist and will be ignored.`);
-    }
+      if (!isServiceDeclared) {
+        logger.warn(`Service ${serviceId} does not exist and will be ignored.`);
```
MattiSG, line 46:
```suggestion
        logger.warn(`Parameter "${serviceId}" was interpreted as a service ID to update, but no matching declaration was found. It will be ignored.`);
```

scripts/release/README.md
```diff
@@ -0,0 +1,9 @@
+### Releasing a dataset
+
+To release a data to the [OpenTermsArchive-versions GitHub repository](https://github.com/ambanum/OpenTermsArchive-versions/releases) run:
```
MattiSG, line 3:
```suggestion
To release a dataset to the [OpenTermsArchive-versions GitHub repository](https://github.com/ambanum/OpenTermsArchive-versions/releases), run:
```

ops/roles/infra/tasks/mongo.yml
```diff
@@ -0,0 +1,18 @@
+---
+# See https://docs.mongodb.com/manual/tutorial/install-mongodb-on-debian/
+- name: Import the MongoDB public key used by the package management system
+  shell: wget -qO - https://www.mongodb.org/static/pgp/server-5.0.asc | sudo apt-key add -
```
MattiSG, line 4: > Should we not, for security reasons, include this key in our repository?

README.md
```diff
@@ -216,15 +212,42 @@ An example of a production configuration file can be found in `config/production
...
 {
-  "history": {
-    "publicSnapshotsRepository": "https://github.com/ambanum/OpenTermsArchive-snapshots.git",
-    "publicVersionsRepository": "https://github.com/ambanum/OpenTermsArchive-versions.git",
-    "snapshotsBaseUrl": "https://github.com/ambanum/OpenTermsArchive-snapshots/commit/"
+  …
+  "storage": {
+    "git": {
+      "path": "Versions database directory path, relative to the root of this project",
+      "publish": "Boolean. Set to true to publish changes to the shared, global database. Should be true only in production.",
```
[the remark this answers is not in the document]
MattiSG:
```suggestion
      "publish": "Boolean. Set to true to push changes to the origin of the cloned repository at the end of every run. Recommended for production only.",
```

MattiSG (changes requested):
Wow, that was one long review! 😵

Thank you very much @Ndpnt, this is very clean code and the changes are very readable 👏 👍
The main general comment I would have is that the tests between first record and not are maybe a bit too redundant.

My requests for changes are minor.

ops/roles/infra/tasks/mongo.yml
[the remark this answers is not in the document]
Ndpnt: I just follow the installation instructions from MongoDB, see https://docs.mongodb.com/manual/tutorial/install-mongodb-on-debian/

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+      it('stores the service id', () => {
+        expect(mongoRecord.serviceId).to.include(SERVICE_PROVIDER_ID);
+      });
+
+      it('stores the document type', () => {
+        expect(mongoRecord.documentType).to.include(DOCUMENT_TYPE);
+      });
+
+      it('stores information that it is the first record for this specific document', () => {
+        expect(mongoRecord.isFirstRecord).to.be.undefined;
```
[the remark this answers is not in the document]
Ndpnt: Because `expect` library does not provide something like `expect(mongoRecord.isFirstRecord).to.be.falsy` the only alternative here is `expect(mongoRecord.isFirstRecord).to.not.be.ok`, which I found confusing.

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+      });
+
+      context('when provided', () => {
+        it('stores the fetch date', () => {
+          expect(new Date(mongoRecord.fetchDate).getTime()).to.equal(FETCH_DATE.getTime());
+        });
+
+        it('stores the mime type', () => {
+          expect(mongoRecord.mimeType).to.equal(PDF_MIME_TYPE);
+        });
```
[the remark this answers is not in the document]
Ndpnt: Adapters do not implement mandatory fields, everything is permitted as this level. It's at the application level where the notion of mandatory is implemented inside the `recorder` module. So here, it's really `when provided`.

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+
+describe('GitAdapter', () => {
+  let subject;
+
+  before(async () => {
+    git = new Git({
+      path: RECORDER_PATH,
+      author: {
+        name: config.get('recorder.snapshots.storage.git.author.name'),
+        email: config.get('recorder.snapshots.storage.git.author.email'),
```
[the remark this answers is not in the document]
Ndpnt: The other configuration keys (like the git path, etc...) are already retrieved from the configuration, so I don't see the point of not retrieving the author as well. 
I think the next step to limit external dependency will be to have a configuration folder per module that could be overridden at the application level.

src/archivist/recorder/index.test.js
Ndpnt: It does not work

Ndpnt:
> This looks good to me, congratulations.
> 
> I still need to launch it and try on my local but already made some comments.
> 
> Will approve after testing

@martinratinaud did you succeed to test locally?

martinratinaud:
> > This looks good to me, congratulations.
> > I still need to launch it and try on my local but already made some comments.
> > Will approve after testing
> 
> @martinratinaud did you succeed to test locally?

Currently on it and tests are passing correctly.
I will try soon with some services to see how it goes

src/storage-adapters/mongo/index.test.js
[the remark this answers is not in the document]
MattiSG: I was not so concerned about checking `undefined` instead of falsy, but more about checking that the Mongo object has a `isFirstRecord` property set to some specific value. This is whitebox testing, relying on the implementation of the database storage layer. The API that I am concerned with checking is that the `isFirstRecord` property of the Record object is `false`, not how it is implemented 🙂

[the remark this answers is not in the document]
Ndpnt: Maybe I'm wrong but I think you are confused between the MongoAdapter API and the Recorder API. Let's talk about this.

src/archivist/recorder/index.js
```diff
@@ -1,31 +1,19 @@
...
 
   async initialize() {
-    return Promise.all([ this.versionsStorageAdapter, this.snapshotsStorageAdapter ].map(storageAdapter => {
-      if (typeof storageAdapter.initialize != 'function') {
-        return Promise.resolve();
-      }
-
-      return storageAdapter.initialize();
-    }));
+    return Promise.all([ this.versionsStorageAdapter, this.snapshotsStorageAdapter ].map(storageAdapter => storageAdapter.initialize()));
```
MattiSG, line 12: 😉 
```suggestion
    return Promise.all([ this.versionsStorageAdapter.initialize(), this.snapshotsStorageAdapter.initialize() ]);
```

src/archivist/recorder/index.js
```diff
@@ -1,31 +1,19 @@
...
 
   async finalize() {
-    return Promise.all([ this.versionsStorageAdapter, this.snapshotsStorageAdapter ].map(storageAdapter => {
-      if (typeof storageAdapter.finalize != 'function') {
-        return Promise.resolve();
-      }
-
-      return storageAdapter.finalize();
-    }));
+    return Promise.all([ this.versionsStorageAdapter, this.snapshotsStorageAdapter ].map(storageAdapter => storageAdapter.finalize()));
```
MattiSG, line 16:
```suggestion
    return Promise.all([ this.versionsStorageAdapter.finalize(), this.snapshotsStorageAdapter.finalize() ]);
```

src/archivist/recorder/index.test.js
```diff
@@ -193,7 +193,7 @@ describe('Recorder', () => {
           after(async () => adapters.snapshots._removeAllRecords());
 
           it('does not record the document', async () => {
-            expect(id).to.equal(null);
+            expect(id).to.equal(undefined);
```
MattiSG, line 196: > This would minimize test code churn:
```suggestion
            expect(id).to.not.be.ok;
```

bin/validate.js
```diff
@@ -36,11 +36,11 @@ const VALIDATE_PATH = path.resolve(__dirname, '../scripts/validation/validate.js
       .on('fail', () => { hasFailedTests = true; })
       .on('end', () => {
         if (hasFailedTests) {
-          process.exitCode = 1;
+          process.exit(1);
```
MattiSG, line 39: > Thanks to the commit log, I understand this is because of Windows. I would love to have a short synchronous discussion to clarify what we're doing, because I have a growing feeling we don't really understand what's happening in this file 😉

MattiSG (approved):
👏 Congratulations. This important technical advancement also ships major improvements of the code in several areas. I am glad we are retiring some of the fancier setup system. I love how abstracting things away sharpens the focus on what really matters. This paves the way towards a distributable library. This is a great job 👍

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+let collection;
+
+describe('MongoAdapter', () => {
+  let subject;
+
+  before(async () => {
+    subject = new MongoAdapter(config.get('recorder.snapshots.storage.mongo'));
+    await subject.initialize();
+    await client.connect();
+    const db = client.db(config.get('recorder.snapshots.storage.mongo').database);
```
martinratinaud, line 37: `config.get('recorder.snapshots.storage.mongo.database')`
would be more redable IMO

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+
+    collection = db.collection(config.get('recorder.snapshots.storage.mongo').collection);
```
martinratinaud, line 39: same

src/storage-adapters/mongo/index.test.js
```diff
@@ -0,0 +1,501 @@
...
+          documentType: DOCUMENT_TYPE,
+        }).count();
+
+        (mongoDocument = await collection.findOne({
+          serviceId: SERVICE_PROVIDER_ID,
+          documentType: DOCUMENT_TYPE,
+        }));
+      });
+
+      after(() => collection.drop());
```
martinratinaud, line 75: maybe `after(async () => subject._removeAllRecords());`

To be consistent with GIT adapter

config/production.json
```diff
@@ -3,10 +3,32 @@
...
-    "publish": true
+  "recorder": {
+    "versions": {
+      "storage": {
+        "git": {
+          "path": "../ota-versions",
+          "publish": true,
+          "prefixMessageToSnapshotId": "This version was recorded after filtering snapshot https://github.com/ambanum/OpenTermsArchive-snapshots/commit/",
+          "author": {
+            "name": "Open Terms Archive Bot",
```
martinratinaud, line 14: I would keep OTA Bot as it is smaller on commit lines
Screen Shot 2021-12-14 at 14 08 03

martinratinaud:
So I just tested on my local and it's working really good.
This is an impressive piece of work and implementation congratulations

Apart from what has been noted in comments 

- Tried to save a snapshot of a PDF but I never could because content has not changed. I think it should still record a snapshot even if there were no changes (as it is a recurrent question I have, we should discuss this during planification)
- tested on local with mongoDB and it works like a charm
- at some point the issue tracking system could also be implemented through adapters to be able to store files with problems within mongo

👏

martinratinaud (approved):
👏

src/storage-adapters/mongo/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const recordProperties = Object.fromEntries(Object.entries({
+      serviceId,
+      documentType,
+      content: passedContent,
+      mimeType,
+      fetchDate,
+      isRefilter,
+      snapshotId,
+    }).filter(([ , value ]) => value)); // Remove empty values
```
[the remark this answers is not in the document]
Ndpnt: Why casting to boolean?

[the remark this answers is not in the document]
martinratinaud: I believe filter works with a boolean so it's more accurate IMO.
But not  sure my assumption is exact https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
even though all examples are with booleans

Not mandatory anyway

src/archivist/index.js
```diff
@@ -37,19 +39,25 @@ export default class Archivist extends events.EventEmitter {
...
   }
 
+  constructor({ storage: { versions, snapshots } }) {
+    super();
+    this.recorder = new Recorder({ versionsStorageAdapter: versions, snapshotsStorageAdapter: snapshots });
+  }
+
   async init() {
     if (this.services) {
       return this.services;
```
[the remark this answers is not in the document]
Ndpnt: removed.

src/archivist/recorder/index.test.js
```diff
@@ -0,0 +1,509 @@
...
+          const paramsNameToExpectedTextInError = {
+            serviceId: 'service ID',
+            documentType: 'document type',
+            fetchDate: 'fetch date',
+            content: 'content',
+            mimeType: 'mime type',
+          };
+
+          Object.entries(validParams).forEach(([testedRequiredParam]) => {
+            context(`when "${testedRequiredParam}" is missing`, () => {
```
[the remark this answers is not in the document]
Ndpnt: nope, at the contrary, I removed them

src/index.js
```diff
@@ -6,40 +10,47 @@ import Archivist from './archivist/index.js';
...
 
-    return isServiceDeclared;
-  });
-
-  serviceIds = serviceIds.length ? serviceIds : archivist.serviceIds;
-
-  const numberOfDocuments = serviceIds.reduce((acc, serviceId) => acc + archivist.serviceDeclarations[serviceId].getNumberOfDocuments(), 0);
-
-  serviceIds = serviceIds.sort((a, b) => a.localeCompare(b));
+      return isServiceDeclared;
```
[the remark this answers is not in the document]
Ndpnt: Everything in JS is evaluable as truthy or falsy, it's not necessary to cast in boolean. What improvement it brings?

src/storage-adapters/git/index.js
```diff
@@ -0,0 +1,159 @@
...
+    this.author = author;
+    this.fileExtension = fileExtension;
+    this.needsPublication = publish;
+    this.prefixMessageToSnapshotId = prefixMessageToSnapshotId;
+  }
+
+  async initialize() {
+    this.git = new Git({ path: this.path, author: this.author });
+
+    return this.git.initialize();
```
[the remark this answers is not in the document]
Ndpnt: It's just here to return the promise.
It avoids to write:

```
async initialize() {
    this.git = new Git({ path: this.path, author: this.author });

    await this.git.initialize();
    return;
}
```

src/storage-adapters/mongo/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+  async initialize() {
+    await this.client.connect();
+    const db = this.client.db(this.databaseName);
+
+    this.collection = db.collection(this.collectionName);
+  }
+
+  async finalize() {
+    return this.client.close();
```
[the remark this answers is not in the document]
Ndpnt: it is because it's a promise, otherwise it won't wait until the end.

src/storage-adapters/mongo/index.js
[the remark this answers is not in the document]
MattiSG: It works with a truth value. If the value is falsy it will be removed 🙂

src/storage-adapters/git/index.test.js
```diff
@@ -0,0 +1,559 @@
...
+const FETCH_DATE = new Date('2000-01-01T12:00:00.000Z');
+const SNAPSHOT_ID = 'snapshot_id';
+const MIME_TYPE = 'text/html';
+const PDF_CONTENT = fs.readFileSync(path.resolve(__dirname, '../../../test/fixtures/terms.pdf'), { encoding: 'utf8' });
+const PDF_MIME_TYPE = 'application/pdf';
+
+let git;
+
+describe('GitAdapter', () => {
+  let subject;
```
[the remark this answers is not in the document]
MattiSG: I use it everywhere, I find this convention much easier to read and write tests: wherever I am in my tests, I know that if I mention `subject` it is an instance of the tested module / class 🙂

config/production.json
MattiSG: Since Git and its metadata is part of our user-visible UI, I prefer that we expand the name:

Capture d’écran 2021-12-14 à 09 44 21

MattiSG: Thanks to this being in config, maybe we can have the best of both world with using `OTA Bot` in the dev config and `Open Terms Archive Bot` in production 🙂

src/storage-adapters/git/index.js
```diff
@@ -0,0 +1,159 @@
...
+    let message = `${prefix} ${serviceId} ${documentType}`;
+
+    if (snapshotId) {
+      message = `${message}\n\n${this.prefixMessageToSnapshotId}${snapshotId}`;
+    }
+
+    return message;
+  }
+
+  async _removeAllRecords() {
```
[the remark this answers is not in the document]
Ndpnt: I'm not an English specialist, so  @MattiSG `clear` or `remove`?

src/storage-adapters/git/index.js
```diff
@@ -0,0 +1,159 @@
...
+    } catch (error) {
+      throw new Error(`Could not commit ${filePath} with message "${message}" due to error: "${error}"`);
+    }
+  }
+
+  _getPathFor(serviceId, documentType, fileExtension) {
+    return `${this.path}/${serviceId}/${documentType}.${fileExtension || this.fileExtension}`;
+  }
+
+  _isTracked(serviceId, documentType) {
```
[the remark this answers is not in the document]
Ndpnt: `async` will wrap the method inside a promise. When we do not use `await`, there is not need to add `async` even if the function is asynchronous as long as it returns a promise.

src/storage-adapters/git/index.js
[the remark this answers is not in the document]
MattiSG: PR was merged, let me know if you still need inputs on the naming 😉

PR #727, opened 2022-01-05
Improve developer experience

MattiSG:
When deploying a new instance of Open Terms Archive for [Dating services](https://github.com/OpenTermsArchive/services-dating/), I encountered many small surprises. I documented the bigger ones so we can discuss them, but I also had a few suggestions that I compiled in this branch aiming at improving the developer / contributor experience. I will open separate PRs for changes that I expect to be more controversial 🙂

**I recommend to review this PR commit by commit instead of consolidated, as each change is atomic and hopefully duly justified by the commit message, whereas the consolidated changes have no global meaning.**

### Pure refactor

- Fix typos and whitespace
- Use long options in shell commands to improve their readability
- Clean up npm scripts
- Improve logging in tests, ops and run: add details, unify phrasings, make slow tests detection relevant

### Minor functional changes that should have no negative impact but are worth having in mind

- Use `npm ci` instead of `npm install` to increase speed and stability of unsupervised installs
- Fail validation if no service is provided
- Exit with a different status code depending if tests fail or error
- Decrease default fetch timeout to 30 seconds instead of 45, to increase run and tests speed
- Upgrade a deprecated buffer syntax for downloading PDF files

### Run

- Inconsistent punctuation in progress logging: “The document cannot be accessed or its content can not be selected: The provided selector ".StaticPage-599770692c57b3b7" has no match in the web page https://deliveroo.ie/legal.” vs “The document cannot be accessed or its content can not be selected: Received HTTP code 403 when trying to fetch 'https://www.coinbase.com/legal/cookie'”.
- Inconsistent quotes in progress logging: “The provided selector ".TermsOfUse-module--terms-of-use--1aCf6" has no match in the web page https://tresorit.com/terms-of-use.”
- “Start tracking” inconsistent with “Refiltering” for long tasks: do we “start” or execute them?
- Filter function failure logs whole filter body.
- Deprecated logs: 
```
2021-12-20 11:51:49 info  App Store — Review Guidelines                           Recorded first version with id 0f4ef19.
(node:97378) [node-fetch#buffer] DeprecationWarning: Please use 'response.arrayBuffer()' instead of 'response.buffer()'
(Use `node --trace-deprecation ...` to show where the warning was created)
2021-12-20 11:51:51 info  App Store — Developer Terms                             Recorded first snapshot with id 48f9335.
```
- Timeout was decreased to 30 seconds in 056b9cf48a26a006b7dc913d25a92036340c372d but not for page fetch.

### Validation

- “has a resulting filtered content with at least 100”: 100 what?
- Subcategory “when fetched and filtered twice in a row” decreases readability because it is the only category that is not a name of a service / document.
- Repeated “has a” is not helpful.
- Timings in red / yellow are not realistic.

### Ops

- The branch for databases is hardcoded to `master`, which is not compatible with more recent default `main`.

### Before

Capture d’écran 2021-12-20 à 16 29 33

### After

Capture d’écran 2021-12-20 à 16 36 28

### Before

Capture d’écran 2021-12-20 à 16 31 34

### After
Capture d’écran 2021-12-20 à 16 32 53

### Before
Capture d’écran 2021-12-20 à 19 19 20

### After

Capture d’écran 2021-12-20 à 20 10 46

### Before
Capture d’écran 2021-12-28 à 22 13 09

### After

Capture d’écran 2021-12-28 à 22 13 20


ops/roles/ota/tasks/main.yml
```diff
@@ -36,6 +37,7 @@
     force: yes
     accept_hostkey: yes
     key_file: '/home/{{ ansible_user }}/.ssh/ota-bot-key'
+    depth: 1
```
martinratinaud, line 40: For reference for other reviewers

```Create a shallow clone with a history truncated to the specified number or revisions. The minimum possible value is 1, otherwise ignored. Needs git>=1.9.1 to work correctly.```

Comment (martinratinaud) [reply ops/roles/ota/tasks/main.yml:40]: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html

**Review by martinratinaud — APPROVED** [review]
Thanks a lot @MattiSG for this useful cleanup

Good way to start a new year :-)

File: src/logger/index.js [hunk 2]
```diff
@@ -88,7 +88,7 @@ logger.onRefilteringStarted = (numberOfServices, numberOfDocuments) => {
 };
 
 logger.onRefilteringCompleted = (numberOfServices, numberOfDocuments) => {
-  logger.info(`👆  Refiltered ${numberOfDocuments} documents from ${numberOfServices} services.\n`);
+  logger.info(`👆  Examined ${numberOfDocuments} documents from ${numberOfServices} services for refiltering.\n`);
```
Ndpnt: Should we also change the `onRefilteringStarted` log as well? 
Like this:
```
logger.onRefilteringStarted = (numberOfServices, numberOfDocuments) => {
  logger.info(`👇  Examining ${numberOfDocuments} documents from ${numberOfServices} services for refiltering.…`);
 };
```

src/logger/index.js
```diff
@@ -85,18 +92,23 @@ logger.onVersionNotChanged = (serviceId, type) => {
...
 logger.onTrackingStarted = (numberOfServices, numberOfDocuments) => {
   logger.info(`👇  Tracking changes of ${numberOfDocuments} documents from ${numberOfServices} services…`);
+  recordedSnapshotsCount = 0;
+  recordedVersionsCount = 0;
 };
 
 logger.onTrackingCompleted = (numberOfServices, numberOfDocuments) => {
-  logger.info(`👆  Tracked changes of ${numberOfDocuments} documents from ${numberOfServices} services.\n`);
+  logger.info(`👆  Tracked changes of ${numberOfDocuments} documents from ${numberOfServices} services.`);
+  logger.info(`    Recorded ${recordedSnapshotsCount} new snapshots and ${recordedVersionsCount} new versions.`);
```
Ndpnt, line 111: I'm in favor of removing spaces, like that:  
Capture d’écran 2022-01-11 à 10 22 01
If you agree, I already made the commit, I just have to push.

src/logger/index.js
```diff
@@ -53,6 +53,9 @@ if (config.get('logger.sendMailOnError')) {
   }
 }
 
+let recordedSnapshotsCount;
```
Ndpnt, line 56: > Not really a big fan of having logic in the logger

ops/roles/ota/tasks/database.yml
```diff
@@ -13,7 +13,7 @@
     - update
     - setup
 
-- name: Obtain {{ name }} initial data from {{ repository }}
+- name: Obtain {{ name }} initial data from branch {{ branch }} of {{ repository }}
```
Ndpnt, line 16: Where is defined `branch` variable? I do not find it, I think it does not exist.

Ndpnt (changes requested):
Great improvements! 👏 
I have just a question regarding the `branch` variable.

src/logger/index.js
MattiSG: Why not. I agree it would be nicer to have it aligned, but I don't like so much the idea of having a log line that contains only an emoji and that is not significant if taken away from its context 😕 Maybe we coud then use a different log level, like `debug`, for these lines?

ops/roles/ota/tasks/database.yml
MattiSG: Sorry, this should have been imported from https://github.com/ambanum/OpenTermsArchive/commit/b2cf6280691545af8e5c060f0c7a83773dc84935#diff-dedd8f47cf84f35c9e2aa4e11ba9e512921ac00f50ce1ca6deaefb7355fa964e!

MattiSG:
As discussed, I'm happy to delegate finalisation and merging of this PR to @Ndpnt considering my workload for this sprint. Thanks @Ndpnt! 😌

src/logger/index.js
Ndpnt: I'm in favor or removing the emoji and just adding a newline line after refiltering and tracking. @martinratinaud agree with that?

Ndpnt: Like that
Capture d’écran 2022-01-12 à 09 24 21

martinratinaud: Sure

Ndpnt (approved):

PR #728, opened 2022-01-05
Remove “prod” options from NPM scripts

MattiSG:
Consider that Ansible manages production environments, not NPM scripts. This allows other environments to be considered as “production”: otherwise, Ansible sets up pm2 to use `npm start:prod` and its `NODE_ENV` is ignored.

I did not ship this as a fix, even though it is needed for deploying different instances, because removing the `:prod` options from NPM scripts altogether might impact other developers in ways I can not foresee.

### DIFF
--- a/ops/roles/ota/tasks/main.yml
+++ b/ops/roles/ota/tasks/main.yml
@@ -91,7 +91,7 @@
 
 - name: Start Open Terms Archive
   command:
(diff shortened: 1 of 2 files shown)


martinratinaud (commented):
Hi @MattiSG 
I'm not sure to understand where you now pass this NODE_ENV=production ?

Ndpnt (approved):

PR #736, opened 2022-01-12
Centralise document types declaration file

Ndpnt:
Should be merged with https://github.com/OpenTermsArchive/services-all/pull/282.

### DIFF
--- a/scripts/validation/service.schema.js
+++ b/scripts/validation/service.schema.js
@@ -1,14 +1,6 @@
-import { createRequire } from 'module';
-import path from 'path';
-import { fileURLToPath } from 'url';
+import { DOCUMENT_TYPES } from '../../src/archivist/services/index.js';
 
-import config from 'config';
-
-const require = createRequire(import.meta.url);
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const TYPES = require(path.resolve(__dirname, '../../', config.get('services.documentTypesPath'))); // eslint-disable-line import/no-dynamic-require
-
-const AVAILABLE_TYPES_NAME = Object.keys(TYPES);
+const AVAILABLE_TYPES_NAME = Object.keys(DOCUMENT_TYPES);
 
 const documentsProperties = () => {
   const result = {};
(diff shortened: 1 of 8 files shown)


scripts/validation/service.schema.js
```diff
@@ -1,12 +1,4 @@
-import { createRequire } from 'module';
-import path from 'path';
-import { fileURLToPath } from 'url';
-
-import config from 'config';
-
-const require = createRequire(import.meta.url);
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const TYPES = require(path.resolve(__dirname, '../../', config.get('services.documentTypesPath'))); // eslint-disable-line import/no-dynamic-require
+import { DOCUMENT_TYPES as TYPES } from '../../src/archivist/services/index.js';
```
martinratinaud, line 1: not sure what benefit there is to `as TYPES`

maybe 
```
import { DOCUMENT_TYPES } from '../../src/archivist/services/index.js';
const AVAILABLE_TYPES_NAME = Object.keys(DOCUMENT_TYPES);
```

Not mandatory of course

martinratinaud (approved):

src/archivist/services/index.js
```diff
@@ -11,6 +12,10 @@ const fs = fsApi.promises;
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const declarationsPath = path.resolve(__dirname, '../../..', config.get('services.declarationsPath'));
 
+const require = createRequire(import.meta.url);
+
+export const DOCUMENT_TYPES = require(path.resolve(__dirname, './documentTypes.json')); // eslint-disable-line import/no-dynamic-require
```
MattiSG, line 17: Why make this a dynamic require?

```suggestion
export const DOCUMENT_TYPES = require(./documentTypes.json');
```

However, as outlined in [this article](https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/#option-1%3A-read-and-parse-json-files-yourself), I'm in favour of using `JSON.parse` for readability reasons, since neither option feels right.

MattiSG (approved):

src/archivist/services/index.js
Ndpnt: Because I had to choose one, and I stuck to the method already chosen in the codebase. 
But I understand your point so I will migrate the other dynamic require to `JSON.parse`.

PR #737, opened 2022-01-12
Adapt urls in Github issues to new contribution tool

martinratinaud:
Even though there is a fallback in place right now, github issues should be created linking to the right contribution tool

This follow up https://github.com/ambanum/OpenTermsArchive.org/issues/92

### DIFF
--- a/src/github/index.js
+++ b/src/github/index.js
@@ -12,8 +12,8 @@ const ISSUE_STATE_ALL = 'all';
 
 const UPDATE_DOCUMENT_LABEL = process.env.GITHUB_LABEL_UPDATE || 'update';
 
-const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/contribute/service';
-const CONTRIBUTE_URL = 'https://opentermsarchive.org/en/contribute/service';
+const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/en/service';
+const CONTRIBUTE_URL = 'https://contribute.opentermsarchive.org/en/service';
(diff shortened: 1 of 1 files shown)


src/github/index.js
```diff
@@ -12,8 +12,8 @@ const ISSUE_STATE_ALL = 'all';
 
 const UPDATE_DOCUMENT_LABEL = process.env.GITHUB_LABEL_UPDATE || 'update';
 
-const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/contribute/service';
-const CONTRIBUTE_URL = 'https://opentermsarchive.org/en/contribute/service';
+const LOCAL_CONTRIBUTE_URL = `http://localhost:3000/contribute/service?${process.env.GITHUB_REPO}`;
```
Ndpnt, line 15: Set where your construct the `urlQueryParams` string instead of here. Moreover an `&` is missing:

```
const urlQueryParams = `${process.env.GITHUB_REPO}&step=2&url=${encodedUrl}&name=${encodedName}&documentType=${encodedType}${noiseSelectorsQueryString}${contentSelectorsQueryString}&expertMode=true`;
```

Ndpnt: Also I see you did not give a name to your query param. I think `repository=${process.env.GITHUB_REPO}` should be better

martinratinaud: You're all right! (late commit after a too long day)

Everything is fixed

Ndpnt (approved):

PR #748, opened 2022-02-08
Industrialize dataset publishing

### LINKED ISSUE #742: Industrialize dataset publishing
As a contributor,
I get my tracked terms published automatically in my versions repository,
So my data can be reused and I can link to it

- [x] Add tests
- [x] Expose an `npm run export $filename` command that generates a zipfile with the name `filename.zip`
- [x] Include README in dataset
- [x] Include LICENSE in dataset
- [x] `export/makedataset.js` -> `export/main.js`
- [ ] Move `scripts/export/README` instructions into main `README`
- [x] Use ISO date in filename, with `T` and `Z`
- [x] Use `renamer` module instead of renaming manually
- [ ] Correct root README on publishing frequency
- [ ] Bonus if times allow: replace `gh` with the GitHub Releases REST API
- [x] Improve release message:

> # Open Terms Archive ${servicesRepositoryName} ${releaseDate} dataset
>
> This dataset consolidates the contractual documents of ${servicesCount} service providers, in all their versions that were accessible online between ${firstCommitDate} and ${lastCommitDate}.
>
> This dataset is tailored for datascientists and other analysts. You can also explore all these versions interactively on ${versionsRepositoryURL}.
> 
> ### Dataset format
> 
> This dataset represents each version of a document as a separate [Markdown](https://spec.commonmark.org/0.30/) file, nested in a directory with the name of the service provider and in a directory with the name of the document type. The filesystem layout will look like below.
>
> ```
> ␜
> ├ README.md
> ├┬ Service provider 1 (e.g. “Facebook”)
> │├┬ Document type 1 (e.g. “Terms of Service”)
> ││├ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-08-01T01-03-12Z.md)
> ┆┆┆
> ││└ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-10-03T08-12-25Z.md)
> ┆┆
> │└┬ Document type X (e.g. “Privacy Policy”)
> │ ├ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-05-02T03-02-15Z.md)
> ┆ ┆
> │ └ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-11-14T12-36-45Z.md)
> ┆
> └┬ Service provider Y (e.g. Google”)
>  ├┬ Document type 1 (e.g. “Developer Terms”)
>  │├ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2019-03-12T04-18-22Z.md)
>  ┆┆
>  │└ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-12-04T22-47-05Z.md)
>  ┆
>  └┬ Document type Z (e.g. “Privacy Policy”)
>   ├ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-05-02T03-02-15Z.md)
>   ┆
>   └ YYYY-DD-MMTHH-MM-SSZ.md (e.g. “2021-11-14T12-36-45Z.md)
> ```
>
> ### License
>
> This dataset is made available under an [Open Database (OdBL) License](https://opendatacommons.org/licenses/odbl/1.0/) by Open Terms Archive Contributors.

Ndpnt:
Fixes #742.

Still missing:
- [x] Update dataset release documentation
- [x] Add README to dataset script?
- [x] Update config keys in documentation


scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+    addRecordToArchive({
+      record: { serviceId: renamedServiceId, documentType: renamedDocumentType, ...othersRecordAttributes },
+      periodDates,
+      services,
+      archive,
+      archiveName,
+    });
+  }
+
+  addReadmeAndLicenseToArchive({ releaseDate, periodDates, services, archive, archiveName });
```
MattiSG, line 39:
```suggestion
  addAssets({ releaseDate, periodDates, services, archive, archiveName });
```

…but honestly, I am not convinced of the added value of this function. I'd recommend to either put all the body in this function, or split into `addReadme` and `addLicense`.

MattiSG (commented):
I'll send this batch of comments already in order to not block review for too long 🙂 

For the file additions to the archive, I wonder if it would not be safer and more readable to have each function return a `{ content, name }` object and let the `generateArchive` main function call `archive.append` on each return. This way, each function would be pure and only the main loop would have side effects.

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+import GitAdapter from '../../../src/storage-adapters/git/index.js';
+import MongoAdapter from '../../../src/storage-adapters/mongo/index.js';
+import * as renamer from '../../utils/renamer/index.js';
+import readme from '../assets/README.template.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const locale = 'en-EN';
+const dateOptions = { year: 'numeric', month: 'long', day: 'numeric' };
+
+export default async function generateArchive({ archivePath, releaseDate }) {
```
MattiSG, line 17: Why do we need `releaseDate`? Shouldn't it be obtained locally from the current date?
```suggestion
export default async function generate({ archivePath, releaseDate }) {
```

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+import mime from 'mime';
+import StreamZip from 'node-stream-zip';
+
+import GitAdapter from '../../../src/storage-adapters/git/index.js';
+
+import generateArchive from './index.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const { expect } = chai;
```
MattiSG, line 17: > Move these imports before our local imports.

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+      let message = `expected ${givenContentPath} to have the same content as ${expectedContentPath}
+
+     Statistics - equal entries: ${result.equal}, distinct entries: ${result.distinct}, left only entries: ${result.left}, right only entries: ${result.right}, differences: ${result.differences}\n`;
+
+      result.diffSet.forEach(diff => {
+        if (diff.state == 'equal') {
+          return;
+        }
+
+        message += `     Reason: ${diff.reason}: File - name1: ${diff.name1}, type1: ${diff.type1}, name2: ${diff.name2}, type2: ${diff.type2}\n`;
```
MattiSG, line 64:
```suggestion
        message += `     Reason: ${diff.reason}: File - name is "${diff.name1}" of type ${diff.type1} vs "${diff.name2}" of type ${diff.type2}\n`;
```

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+  });
+});
+
+describe('Export', () => {
+  describe('#generateArchive', () => {
+    let storageAdapter;
+    const archiveName = 'test-dataset';
+    const archivePath = path.resolve(__dirname, `./tmp/${archiveName}.zip`);
+    const tmpPath = path.resolve(__dirname, './tmp');
```
MattiSG, line 77: Why not use [`os.tmpdir()`](https://nodejs.org/api/os.html#ostmpdir) and avoid leftovers? 🙂

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  if (!periodDates.first || new Date(fetchDate) < new Date(periodDates.first)) {
+    periodDates.first = fetchDate;
+  }
+
+  services.add(serviceId);
+
+  const fileName = fetchDate.toISOString()
+    .replace(/\.\d{3}/, '') // Remove milliseconds
+    .replace(/:|\./g, '-'); // Replace `:` and `.` by `-` to be compliant with the file system
```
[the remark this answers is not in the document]
Ndpnt: Why adding space and why removing caps. What is your rules to choose if we use a caps or not when commenting?

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+        fetchDate: FIRST_FETCH_DATE,
+        snapshotId: SNAPSHOT_ID,
+      });
+
+      await storageAdapter.record({
+        serviceId: FIRST_SERVICE_PROVIDER_ID,
+        documentType: FIRST_DOCUMENT_TYPE,
+        content: SECOND_CONTENT,
+        fetchDate: SECOND_FETCH_DATE,
+        snapshotId: SNAPSHOT_ID,
```
MattiSG, line 104:
```suggestion
        snapshotId: `${SNAPSHOT_ID}_second`,
```

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+
+    after(async () => {
+      await fs.rm(tmpPath, { recursive: true });
+      await storageAdapter._removeAllRecords();
+    });
+
+    it('is an archive', () => {
+      const mimeType = mime.getType(archivePath);
+
+      expect(mimeType).to.be.equal('application/zip');
```
MattiSG, line 141:
```suggestion
      expect(mimeType).to.equal('application/zip');
```

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+        fetchDate: FOURTH_FETCH_DATE,
+        snapshotId: SNAPSHOT_ID,
+      });
+
+      await generateArchive({
+        archivePath,
+        releaseDate: new Date(RELEASE_DATE),
+      });
+
+      zip = new StreamZip.async({ file: archivePath }); // eslint-disable-line new-cap
```
MattiSG, line 128: I would prefer that we use the [`properties: false`](https://eslint.org/docs/rules/new-cap#properties) option of the `new-cap` ESlint rule instead of a `disable-line` statement 🙂

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+  });
+});
+
+describe('Export', () => {
+  describe('#generateArchive', () => {
+    let storageAdapter;
+    const archiveName = 'test-dataset';
+    const archivePath = path.resolve(__dirname, `./tmp/${archiveName}.zip`);
+    const tmpPath = path.resolve(__dirname, './tmp');
+    const datasetFixturesPath = path.resolve(__dirname, './test/fixtures/dataset');
```
MattiSG, line 78:
```suggestion
    const EXPECTED_DATASET_PATH = path.resolve(__dirname, './test/fixtures/dataset');
```

scripts/dataset/export/test/fixtures/dataset/ServiceB/Privacy Policy/2022-01-01T12-12-24Z.md
```diff
@@ -0,0 +1 @@
+Second Content
```
MattiSG, line 1: It is very misleading that for Service B the content start with the second one. Is this truly necessary? 🙁

MattiSG (commented):
Second batch of comments 😉

scripts/dataset/index.js
```diff
@@ -0,0 +1,55 @@
...
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import generateRelease from './export/index.js';
+import logger from './logger/index.js';
+import publishRelease from './publish/index.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+export async function release({ publicationEnabled, removeLocalCopyEnabled, fileName }) {
```
MattiSG, line 11: Use consistent flag names: either `publicationEnabled` and `localCopyRemovalEnabled`, or `publishEnabled` and `removeLocalCopyEnabled`. Or maybe `shouldPublish` and `shouldRemoveLocalCopy`? 🙂

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+  await renamer.loadRules();
+
+  const periodDates = { first: null, last: null };
+  const services = new Set();
+
+  for await (const record of versionsStorageAdapter.iterate()) {
+    const { serviceId, documentType, ...othersRecordAttributes } = record;
+    const { serviceId: renamedServiceId, documentType: renamedDocumentType } = renamer.applyRules(serviceId, documentType);
+
+    addRecordToArchive({
```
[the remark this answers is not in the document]
Ndpnt: In this case, shouldn't  we also rename the variable `record` in `version`?

scripts/dataset/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+export async function generate({ archivePath, releaseDate }) {
+  logger.info('Start exporting dataset…');
+
+  const { servicesCount, firstCommitDate, lastCommitDate } = await generateRelease({ archivePath, releaseDate });
+
+  logger.info(`Dataset exported in ${archivePath}`);
+
+  return { servicesCount, firstCommitDate, lastCommitDate };
+}
```
MattiSG, line 33: 4 lines of code, out of which 1 is to return a result and all variables are instantiated just for that. I suggest to bring it back to the main body 🙂

scripts/dataset/index.js
```diff
@@ -0,0 +1,55 @@
...
+    return;
+  }
+
+  return publish({ archivePath, releaseDate, servicesCount, firstCommitDate, lastCommitDate, publicationEnabled, removeLocalCopyEnabled });
+}
+
+export async function generate({ archivePath, releaseDate }) {
+  logger.info('Start exporting dataset…');
+
+  const { servicesCount, firstCommitDate, lastCommitDate } = await generateRelease({ archivePath, releaseDate });
```
MattiSG, line 28:
```suggestion
  const result = await generateRelease({ archivePath, releaseDate });
```

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  if (config.has('recorder.versions.storage.git')) {
+    versionsStorageAdapter = new GitAdapter({
+      ...config.get('recorder.versions.storage.git'),
+      path: path.resolve(__dirname, '../../../', config.get('recorder.versions.storage.git.path')),
+      fileExtension: 'md',
+    });
+  } else if (config.has('recorder.versions.storage.mongo')) {
+    versionsStorageAdapter = new MongoAdapter(config.get('recorder.versions.storage.mongo'));
+  }
```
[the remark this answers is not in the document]
Ndpnt: I prefer to throw at the adapter level. When we implement it, there will be less code to refactor.

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  await versionsStorageAdapter.initialize();
+
+  return versionsStorageAdapter;
+}
+
+async function initializeArchive(archivePath) {
+  fsApi.mkdirSync(path.dirname(archivePath), { recursive: true });
```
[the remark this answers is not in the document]
Ndpnt: Because it has to exists before we can create the archive

scripts/dataset/logger/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+const { combine, timestamp, printf, colorize } = winston.format;
+
+const alignedWithColorsAndTime = combine(
+  colorize(),
+  timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
+  printf(({ level, message, timestamp }) => `${timestamp} ${level.padEnd(15)} ${message}`),
+);
+
+const consoleTransport = new winston.transports.Console();
```
MattiSG, line 11: So no email notifications in case of publishing failures… This is not very satisfactory. Couldn't we use the main notifier setup?

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+
+import logger from './logger/index.js';
+
+import { release } from './index.js';
+
+(async () => {
+  const args = process.argv.slice(2);
+  const publicationEnabled = args.includes('--publish');
+  const removeLocalCopyEnabled = args.includes('--remove-local-copy');
+  const scheduleEnable = args.includes('--schedule');
```
MattiSG, line 11:
```suggestion
  const scheduleEnabled = args.includes('--schedule');
```

scripts/dataset/publish/index.js
```diff
@@ -0,0 +1,50 @@
...
+import config from 'config';
+import dotenv from 'dotenv';
+import { Octokit } from 'octokit';
+
+import * as readme from '../assets/README.template.js';
+
+dotenv.config();
+
+const locale = 'en-EN';
+const dateOptions = { year: 'numeric', month: 'long', day: 'numeric' };
```
MattiSG, line 14: This is duplicated and should thus probably be in the `readme` template.

scripts/dataset/publish/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+export default async function publish({ archivePath, releaseDate, servicesCount, firstCommitDate, lastCommitDate }) {
+  const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
+
+  const [ owner, repo ] = url.parse(config.get('dataset.versionsRepositoryURL')).pathname.split('/').filter(element => element !== '');
```
MattiSG, line 19:
```suggestion
  const [ owner, repo ] = url.parse(config.get('dataset.versionsRepositoryURL')).pathname.split('/').filter(component => component);
```

scripts/dataset/publish/index.js
```diff
@@ -0,0 +1,50 @@
...
+    data: fsApi.readFileSync(archivePath),
+    headers: {
+      'content-type': 'application/zip',
+      'content-length': fsApi.statSync(archivePath).size,
+    },
+    name: path.basename(archivePath),
+    url: uploadUrl,
+  });
+
+  return { releaseUrl };
```
MattiSG, line 49:
```suggestion
  return releaseUrl;
```

scripts/dataset/export/index.js
Ndpnt: It's to share the exact same date between dataset generation and publication on github. To ensure the message that will be with the release and the readme inside the dataset will have the same date.

src/storage-adapters/git/index.js
```diff
@@ -76,6 +76,45 @@ export default class GitAdapter {
...
+  async* iterate() {
+    const initialCommitHash = (await this.git.raw([ 'rev-list', '--max-parents=0', 'HEAD' ])).trim();
+    const currentBranchName = (await this.git.raw([ 'rev-parse', '--abbrev-ref', 'HEAD' ])).trim();
+
+    try {
+      let previousCommitHash;
+
+      /* eslint-disable no-await-in-loop */
+      while (previousCommitHash != initialCommitHash) {
+        const [{ hash, date, message, diff: { files: [{ file: relativeFilePath }] } }] = await this.git.log([ '-1', '--stat=4096' ]); // Get current commit informations
```
MattiSG, line 88:
```suggestion
        const [{ hash, date, message, diff: { files: [{ file: relativeFilePath }] } }] = await this.git.log([ '-1', '--stat=4096' ]);  // get current commit information
```

src/storage-adapters/git/index.js
```diff
@@ -76,6 +76,45 @@ export default class GitAdapter {
...
+            documentType,
+            content: await fs.readFile(absoluteFilePath, { encoding: 'utf8' }),
+            fetchDate: new Date(date),
+          };
+        }
+
+        previousCommitHash = hash;
+
+        if (initialCommitHash != hash) {
+          await this.git.checkout(['HEAD~1']); // Checkout to the parent commit
```
MattiSG, line 109:
```suggestion
          await this.git.checkout(['HEAD^']);  // checkout the parent commit
```

src/storage-adapters/git/index.js
```diff
@@ -76,6 +76,45 @@ export default class GitAdapter {
...
+            serviceId,
+            documentType,
+            content: await fs.readFile(absoluteFilePath, { encoding: 'utf8' }),
+            fetchDate: new Date(date),
+          };
+        }
+
+        previousCommitHash = hash;
+
+        if (initialCommitHash != hash) {
```
MattiSG, line 108: I'm not sure this condition is necessary since the loop already checks it 🙂

scripts/dataset/export/index.test.js
Ndpnt: Great idea!

src/storage-adapters/git/index.js
```diff
@@ -76,6 +76,45 @@ export default class GitAdapter {
...
+          await this.git.checkout(['HEAD~1']); // Checkout to the parent commit
+        }
+      }
+      /* eslint-enable no-await-in-loop */
+    } finally {
+      await this.git.checkout([currentBranchName]); // Checkout back to the top of the branch
```
MattiSG, line 114:
```suggestion
      await this.git.checkout([currentBranchName]);
```

scripts/dataset/export/index.test.js
Ndpnt: This is not important

src/storage-adapters/git/index.test.js
```diff
@@ -473,4 +473,47 @@ describe('GitAdapter', () => {
...
+    after(async () => subject._removeAllRecords());
+
+    it('iterates through all records', async () => {
+      let numberOfRecordCovered = 0;
+
+      for await (const record of subject.iterate()) { // eslint-disable-line no-unused-vars
+        numberOfRecordCovered++;
+      }
+
+      expect(numberOfRecordCovered).to.equal(3);
```
MattiSG, line 516: This is a bit too weak a test: we are expecting the iteration to be ordered. This should be checked, for example with a `push(record.content)`.

src/storage-adapters/mongo/index.js
```diff
@@ -85,6 +85,12 @@ export default class MongoAdapter {
     };
   }
 
+  /* eslint-disable */
```
MattiSG, line 88: Why do we disable eslint here? This is smelly 😉

src/storage-adapters/mongo/index.js
```diff
@@ -85,6 +85,12 @@ export default class MongoAdapter {
...
+  async* iterate() {
+    throw new Error('`iterate` function is not yet implemented for MongoDB storage adapter');
```
MattiSG, line 90:
```suggestion
    throw new Error('#iterate is not yet implemented in the MongoDB storage adapter');
```

scripts/dataset/export/test/fixtures/dataset/ServiceB/Privacy Policy/2022-01-01T12-12-24Z.md
Ndpnt: Not at all :)

scripts/dataset/export/index.js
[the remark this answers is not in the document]
MattiSG: Sorry, indeed this should be linted if we reach an agreement 🙂

I personally use the following rule:

- Multi-line comments start with `/**`, end with a line containing only `*/` and are full sentences, with capitals and punctuation.
- Single line comments are always at the end of the code section they comment (no single line comment on its own) to avoid rot, are separated from the code by two spaces, and have no capital nor punctuation. This helps me feel when my single-line comment is getting too long: if it feels like it should be two sentences, I should either refactor, improve the doc, or make that a multi-line comment.

src/storage-adapters/mongo/index.js
Ndpnt: Because as the function is not implemented, there are no `yield` inside and Eslint cries.

scripts/dataset/export/index.js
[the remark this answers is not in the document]
MattiSG: I understand that. I meant: couldn't we `await mkdir` instead and free the main event loop? 🙂

src/storage-adapters/git/index.js
Ndpnt: It is because it is not exactly the same condition and on an edge case it will try to checkout to non existent commit.

scripts/dataset/logger/index.js
Ndpnt: You mean the main logger setup. Yep I will do that.

scripts/dataset/index.js
Ndpnt: It was to export it independently to the `publish` function but as we do not use it, I agree. 
I can also do the same for the `publish` function. Ok for that?

scripts/dataset/index.js
Ndpnt: Go for `should…`.

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const archiveExtension = path.extname(archivePath);
+  const archiveName = path.basename(archivePath, archiveExtension);
+
+  const output = fsApi.createWriteStream(archivePath);
+  const archive = archiver('zip', { zlib: { level: 9 } });
```
martinratinaud, line 77: why level 9 ?
Maybe you can add a quick note

scripts/dataset/export/index.test.js
martinratinaud: By the way, shouldn't eslint take care of sorting the imports itself?

scripts/dataset/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+import generateRelease from './export/index.js';
+import logger from './logger/index.js';
+import publishRelease from './publish/index.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+export async function release({ publicationEnabled, removeLocalCopyEnabled, fileName }) {
+  const releaseDate = new Date();
+  const archiveName = fileName || `dataset-${releaseDate.toISOString().replace(/T.*Z/, '')}`;
```
martinratinaud, line 13: Again, difficult to read, so you remove everything after `T`? Keeping only the main year, month and day, right?

I would use dayjs, but else `/T.*$/` to be sure to remove everything

scripts/dataset/index.js
martinratinaud: I agree

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+  const argsWithoutOptions = args.filter(arg => !arg.startsWith('--'));
+  const [fileName] = argsWithoutOptions;
+
+  if (removeLocalCopyEnabled && !publicationEnabled) {
+    logger.warn('--remove-local-copy is only available with --publish option');
```
martinratinaud, line 16: shouldn't we throw here ? it's unclear what will be the the resulting behavior
I believe it will not remove local copy then? or even not create it?

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  if (!scheduleEnable) {
+    return release({ publicationEnabled, removeLocalCopyEnabled, fileName });
+  }
+
+  logger.info('Release will be created every Monday at 03h30');
+
+  scheduler.scheduleJob({ hour: 3, minute: 30, dayOfWeek: 1 }, async () => release({ publicationEnabled, removeLocalCopyEnabled, fileName }));
```
martinratinaud, line 25: this means it's every monday at 3:30am?
It's never obvious what dayOfWeek is (starting sunday or monday)
Maybe add a comment

martinratinaud (approved):
Looks pretty good!

martinratinaud:
The diff is too long though and it breaks the browser, had to review it commit by commit :-)

scripts/dataset/export/index.js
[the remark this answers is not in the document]
Ndpnt: Ok, I understand

scripts/dataset/export/index.test.js
Ndpnt: I encounter errors when I try to delete the folder

scripts/dataset/main.js
Ndpnt: It's in the log line above

martinratinaud: oups sorry :-)

scripts/dataset/export/index.js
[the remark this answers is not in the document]
Ndpnt: I tried `daysjs` but the format we need is not already defined and at the end we still have to do manual formatting. So I prefer avoid adding a dependency in this case.

src/storage-adapters/git/index.js
```diff
@@ -76,6 +76,45 @@ export default class GitAdapter {
     };
   }
 
+  async* iterate() {
```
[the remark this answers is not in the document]
Ndpnt: The goal here is to return an object that implements the [iteration protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) and to hide the complexity of the iteration in the object from the caller.

scripts/dataset/main.js
Ndpnt: The comment is 2 lines above ;)

MattiSG:
@ndpnt is this PR still a draft?

Ndpnt:
> @Ndpnt is this PR still a draft?

Yes because, there are still the following to do:

- [ ] Update documentation
- [ ] Add log for progression as it is long enough with the service-all versions

martinratinaud:
This looks weird to me that datasets are exported in `scripts/dataset`

Would better have a config variable `dataset.exportFolder` that would be by default `exports` at the root
And change the gitignore

martinratinaud (approved):
👍

README.md
```diff
@@ -291,9 +292,26 @@ npm start $service_id
...
 See [Ops Readme](ops/README.md).
 
-### Exporting a dataset
+### Dataset
 
-See [`export` script Readme](scripts/export/README.md).
+To generate a dataset:
+
+```
+npm run dataset:generate
```
MattiSG, line 300: The `filename` option should be either documented or removed from implementation.

src/index.js
```diff
@@ -1,126 +1,32 @@
...
+    storageAdapter = new GitAdapter({
+      ...config.get(`recorder.${recordType}.storage.git`),
+      fileExtension,
     });
+    break;
+  case 'mongo':
+    storageAdapter = new MongoAdapter(config.get(`recorder.${recordType}.storage.mongo`));
+    break;
+  default:
+    throw new Error('No config were found for snapshots storage adapter');
```
MattiSG, line 28:
```suggestion
    throw new Error('No configuration found for snapshots storage adapter');
```

src/main.js
```diff
@@ -0,0 +1,69 @@
...
+  if (!schedule) {
+    return;
+  }
+
+  logger.info('The scheduler is running…');
+
+  const TRACK_CHANGES_HOURS_INTERVAL = 24;
+
+  logger.info(`Documents will be tracked every ${TRACK_CHANGES_HOURS_INTERVAL} hours\n`);
+  setInterval(async () => {
```
MattiSG, line 66: If I understand correctly, this means that if the initial run fails for any reason, the scheduler will not be set up, am I right? In the same fashion, would any following run failure prevent the next ones from running?

scripts/dataset/logger/index.js
```diff
@@ -0,0 +1,56 @@
+import 'winston-mail';
```
MattiSG, line 1: This is a third-party lib, move it below `import winston`?

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+      });
+
+      return message;
+    }
+  });
+});
+
+describe('Export', () => {
+  describe('#generateArchive', () => {
+    let storageAdapter;
```
MattiSG, line 74: > `let` after `const`?

scripts/dataset/export/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+    const archiveName = 'test-dataset';
+    const archivePath = path.resolve(__dirname, `./tmp/${archiveName}.zip`);
+    const tmpPath = path.resolve(__dirname, './tmp');
+    const EXPECTED_DATASET_PATH = path.resolve(__dirname, './test/fixtures/dataset');
```
MattiSG, line 78: Why is this `const` in all caps and not the others? 😕

scripts/dataset/export/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+    if (fetchDate > lastVersionDate) {
+      lastVersionDate = fetchDate;
+    }
+
+    services.add(serviceId);
+
+    const versionPath = generateVersionPath({ serviceId, documentType, fetchDate });
+
+    logger.info({ message: versionPath, counter: index, hash: version.id });
```
MattiSG, line 47: > Do we really need the counter? I am not sure I understand why we want to log every single entry. For `services-all`, for example, that's over 16 thousand lines. What do I gain from knowing that it is version number 13241 in the run?
>
> Removing the counter would simplify the code and remove the need for the `index` variable.

MattiSG (changes requested):
Thank you for this work.

I am not very satisfied with the amount of duplication we get from `src` in this “script”, it makes me wonder if we drew the line between the Core and the script properly.

### Refilters

I see an unhandled point we should discuss: what do we do with refilter entries? If I understand correctly, they would currently be wrongly presented as a new version.

Let's consider the following history:

1. Version N is a new addition.
2. Version A is recorded as an update to N.
3. Version R is recorded as a refilter of version A.
4. Version B is recorded as a new version, changed from A with R applied.

I understand that the current release process will publish N, A, R, B.
I believe the ideal would be to publish N, R (dated with A date), B.
If this is too complicated for now, then I believe at least we should either:

A. Publish only N, A, B. This will make the diff between A and B wrong, but will preserve date and churn data.
B. Publish only N, R, B. This will make date data wrong but will preserve the diff and churn data.
C. Publish all but mark R as a refilter.

A useful bit of information would be to know how we currently handle refilters in the published datasets.

MattiSG:
As pointed out by @Ndpnt, we should also decrease the release frequency back to 24h, because it is unlikely that we don't deploy for 7 days and #766 switched to `setInterval`, meaning we will only release after the given time has expired since last deploy.

scripts/dataset/export/index.js
Ndpnt: I added this counter for two reasons:
- When manually running the export, it's handy to have a feedback with a progress log. As this can take some time (~15min for the full services-all versions history), it informs us that it is not blocked and shows us the progress.
- When it fails, it is very useful to find out where it fails so that we can understand why.

src/index.js
```diff
@@ -1,126 +1,32 @@
...
-  }
+export function instantiateVersionsStorageAdapter() {
+  return instantiateStorageAdapter('versions', 'md');
+}
 
-  if (config.has('recorder.versions.storage.mongo')) {
-    versionsStorageAdapter = new MongoAdapter(config.get('recorder.versions.storage.mongo'));
-  }
+export function instantiateSnapshotsStorageAdapter() {
+  return instantiateStorageAdapter('snapshots', 'html');
```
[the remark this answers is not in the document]
Ndpnt: It is overridden when it is a PDF file. But this should be refactored in another PR as it is a legacy.

src/main.js
Ndpnt: - If it is a programmer error that we do not know how to recover, yes the app will crash immediately and thus the scheduler will not be set up. But the process manager will restart the process.
- In the same way, if a programmer error happen inside a scheduled job the app will crash immediately.

config/default.json
```diff
@@ -46,11 +51,12 @@
   },
   "tracker": {
     "githubIssues": {
-      "repository": "OpenTermsArchive/sandbox",
+      "repository": "ambanum/test-repo",
```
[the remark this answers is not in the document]
Ndpnt: Because today, the repo `OpenTermsArchive/sandbox` does not exist and I need one when I want to test if the tracker works properly. Ideally we should create this repo and update rights and tokens accordingly.

Ndpnt:
> ### Refilters
> I see an unhandled point we should discuss: what do we do with refilter entries? If I understand correctly, they would currently be wrongly presented as a new version.
> 
> Let's consider the following history:
> 
> 1. Version N is a new addition.
> 2. Version A is recorded as an update to N.
> 3. Version R is recorded as a refilter of version A.
> 4. Version B is recorded as a new version, changed from A with R applied.
> 
> I understand that the current release process will publish N, A, R, B. I believe the ideal would be to publish N, R (dated with A date), B. If this is too complicated for now, then I believe at least we should either:
> 
> A. Publish only N, A, B. This will make the diff between A and B wrong, but will preserve date and churn data. B. Publish only N, R, B. This will make date data wrong but will preserve the diff and churn data. C. Publish all but mark R as a refilter.
> 
> A useful bit of information would be to know how we currently handle refilters in the published datasets.

I understand the issue.
Currently, refilters are handle in the same way in the published datasets. So I suggest to handle the issue in another PR as this one already has too much embedded in it.

src/index.js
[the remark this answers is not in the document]
MattiSG: But since this function is exported, how can the override happen? I understand how the Core can, in its use of this adapter, override it when needed. But if I'm a third party using this adapter, how would it work when I try to load a snapshot that is a PDF file?

I understand we introduce this in this PR because we use the versions adapter in the dataset export script. If we don't need to consume the snapshots adapter from the outside of the app now, let's just not export this function and add a comment explaining that the html one is overridden by a pdf one in the core.

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+  const shouldRemoveLocalCopy = args.includes('--remove-local-copy');
+  const shouldSchedule = args.includes('--schedule');
+  const argsWithoutOptions = args.filter(arg => !arg.startsWith('--'));
+  const [fileName] = argsWithoutOptions;
+
+  if (!shouldSchedule) {
+    return release({ shouldPublish, shouldRemoveLocalCopy, fileName });
+  }
+
+  logger.info(`A release will be published every ${RELEASE_HOURS_INTERVAL} hours\n`);
```
MattiSG, line 19: This is not entirely true: if `shouldPublish` is `false` but `shouldSchedule` is `true`, I understand that a local archive will be created every 24 hours but nothing will be published 🤪

I suggest that `shouldSchedule` implies `shouldPublish`: either override it to `true` after line 18, or `throw` if `shouldSchedule && !shouldPublish`. Another (lazier) option is to just remove this logging instruction and have a release silently created at the given interval, leaving it to the user to not ask for stupid things.

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+
+  setInterval(async () => {
+    logger.info('Start creating the release…');
+    await release({ shouldPublish, shouldRemoveLocalCopy, fileName })();
```
MattiSG, line 23: Oopsie, this crashes!

```suggestion
    await release({ shouldPublish, shouldRemoveLocalCopy, fileName });
```

scripts/dataset/main.js
```diff
@@ -0,0 +1,26 @@
...
+    logger.info('Release published');
+  }, RELEASE_HOURS_INTERVAL * 60 * 60 * 1000);
```
MattiSG, line 25: I don't see the point of the additional logger instructions considering that the `release` function very appropriately logs already.
```suggestion
  setInterval(
    release,
    RELEASE_HOURS_INTERVAL * 60 * 60 * 1000,
    { shouldPublish, shouldRemoveLocalCopy, fileName }
  );
```

scripts/dataset/logger/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+const alignedWithColorsAndTime = combine(
+  colorize(),
+  timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
+  printf(({ level, message, counter, hash, timestamp }) => {
+    let prefix = '';
+
+    if (counter && hash) {
+      prefix = `${counter.toString().padEnd(6)} ${hash.padEnd(40)}`;
+    }
```
MattiSG, line 15: > A ternary would seem appropriate to me here.

scripts/dataset/logger/index.js
```diff
@@ -0,0 +1,23 @@
+import winston from 'winston';
+
+import logger from '../../../src/logger/index.js';
+
+const { combine, timestamp, printf, colorize } = winston.format;
+
+const alignedWithColorsAndTime = combine(
```
MattiSG, line 7:
```suggestion
logger.format = combine(
```

MattiSG (changes requested):
So many important refactors! …but the recurring publishing itself fails because of two small characters 😅

I don't dare to re-apply all my phrasing corrections since I did in my last review and my commit got lost in the following force push 🙁 I “unresolved” them but they won't appear as new comments in this review.

scripts/dataset/main.js
martinratinaud: IMO you can want to save the realeas file but not publish it anywhere and mount an ftp endpoint. So it makes sense to have both.

MattiSG: True! But then the log is misleading :)

MattiSG (approved):
Last commit pair-programmed with @Ndpnt.

PR #749, opened 2022-02-11
Fix broken link

pdehaye:


pdehaye:
The suggested link is mistaken.

PR #751, opened 2022-02-14
Throw inacessible content when data found is empty

martinratinaud:
For now, it is only an Error that is thrown and everytime it happens, it kills the PID
Here, we consider having an empty content is an inaccesible content error

### DIFF
--- a/src/archivist/recorder/index.js
+++ b/src/archivist/recorder/index.js
@@ -1,3 +1,5 @@
+import { InaccessibleContentError } from '../errors.js';
+
 export default class Recorder {
   constructor({ versionsStorageAdapter, snapshotsStorageAdapter }) {
     if (!versionsStorageAdapter || !snapshotsStorageAdapter) {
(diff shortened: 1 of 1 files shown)


MattiSG (commented):
I am surprised that this was not already the case, and prefer to assume that there was good reason for it until we can have a discussion with @Ndpnt.

For reference, the origin of this problem can be reproduced by reverting https://github.com/OpenTermsArchive/services-all/pull/395

martinratinaud:
@Ndpnt can you please look at this?

Ndpnt:
We indeed should throw an `InaccessibleContentError` when a matching selector return empty content. It was not already implemented just because we never met the issue where a declaration has a correct selector but without content.

I open this [alternative PR](https://github.com/ambanum/OpenTermsArchive/pull/767) for consistency (all InaccessibleContentError are thrown from this module) and because I think it is more related to the filtering than the recording.

martinratinaud:
Decline in favor of https://github.com/ambanum/OpenTermsArchive/pull/767

PR #762, opened 2022-03-02
Add ability to customize language requests for instance

martinratinaud:
On UFC instance, we always want to retrieve french version of documents.

For this, it is sometimes necesary to pass a request header.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -28,7 +28,8 @@
   },
   "fetcher": {
     "waitForElementsTimeout": 10000,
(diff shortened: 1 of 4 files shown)


martinratinaud:
I intentionally kept the language config access in the `fetcher` file as we are in for a quick fix.

I believe it's the fetcher caller which should pass the language (and maybe more cookies, or headers) and, even if the main language of a repo should be set, we should be able to set a particular document for a particular language.

Let me know if it's fine like this

PR #763, opened 2022-03-02
Mark refilters in history

### LINKED ISSUE #761: Mark refilters in history
Since at least two weeks, version changes that are a consequence of a refilter are not marked as such anymore. This makes it harder to track actual version changes.

Add a test.

Ndpnt:
Fixes #761

### DIFF
--- a/src/storage-adapters/mongo/index.js
+++ b/src/storage-adapters/mongo/index.js
@@ -75,13 +75,14 @@ export default class MongoAdapter {
       return {};
     }
 
-    const { _id, content, mimeType, fetchDate } = record;
+    const { _id, content, mimeType, fetchDate, isRefilter } = record;
 
     return {
       id: _id.toString(),
       content,
       mimeType,
       fetchDate: new Date(fetchDate),
+      isRefilter: !!isRefilter,
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/storage-adapters/mongo/index.js
```diff
@@ -75,13 +75,14 @@ export default class MongoAdapter {
...
 
-    const { _id, content, mimeType, fetchDate } = record;
+    const { _id, content, mimeType, fetchDate, isRefilter } = record;
 
     return {
       id: _id.toString(),
       content,
       mimeType,
       fetchDate: new Date(fetchDate),
+      isRefilter: !!isRefilter,
```
martinratinaud, line 85: Why do you need to cast it to boolean here ?

martinratinaud (approved):

src/storage-adapters/mongo/index.js
Ndpnt: To avoid returning `undefined` when the property does not exist. I think when a variable is named `isSomething` or `hasSomething`, it should always be a boolean.

MattiSG: I'm not such a fan of double-bangs 🙂
Details: https://medium.com/@MattiSG/when-should-i-use-a-double-bang-1672bf8e677f
```suggestion
      isRefilter: Boolean(isRefilter),
```

Ndpnt: I think `!!` is not so cryptic and is very common, so I'm not really sensible to the argument of the related article but if you have a personal preference of using `Boolean()` I'm ok with that :)

martinratinaud: Also not finding it cryptic but fine with `Boolean` :-)

PR #764, opened 2022-03-02
Do not commit local development config files

martinratinaud:


.gitignore
```diff
@@ -7,8 +7,11 @@ data
 yarn.lock
 tmp
 
+# local config files
```
Ndpnt, line 10: `files` => `file`

Ndpnt (approved):

Issue #753, opened 2022-03-01
Prevent Git concurrency


MattiSG:
There is an increasing amount of cases where a lock file cannot be obtained on the Git repository and where the process crashes. Meanwhile, versions are still created and pushed.

Below are example error logs and a count of each error type.

Observations:

- It is unexpected that multiple Git or OTA instances are run in parallel. It has been demonstrated that only a single OTA instance exists on the faulty servers (stopping it stops all change trackings, and `ps` shows that there is no other Node process running).
- Since the failures appear on random services, it is impossible that the lock files are leftovers that are not cleaned.

### 28/02/2022

Many errors during the week end
```
2022-02-28 06:30:15 �[31merror�[39m Amazon — Privacy Policy Error: Could not commit /home/debian/ota-versions-france/Amazon/Privacy Policy.md with message "Update Amazon Privacy Policy

This version was recorded after filtering snapshot https://github.com/OpenTermsArchive/snapshots-france/commit/621c6bf6b5b4f9ba1cab1611" due to error: "Error: fatal: Unable to create '/home/debian/ota-versions-france/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
"
at GitAdapter._commit (file:///home/debian/ota/src/storage-adapters/git/index.js:99:13)
at async GitAdapter.record (file:///home/debian/ota/src/storage-adapters/git/index.js:35:17)
at async Archivist.recordVersion (file:///home/debian/ota/src/archivist/index.js:231:46)
at async Archivist.trackDocumentChanges (file:///home/debian/ota/src/archivist/index.js:147:29)
```
#### actions taken

@martinratinaud: restart services
- ansible-playbook --inventory ops/inventories/production.yml ops/app.yml --limit france  --tag restart

### 24/02/2022

Seems to be fine now, only one error email

**3 errors**
```
2022-02-24 06:30:09 �[31merror�[39m Amazon — Terms of Service Error: Could not commit /home/debian/ota-versions-france/Amazon/Terms of Service.md with message "Update Amazon Terms of Service

This version was recorded after filtering snapshot https://github.com/OpenTermsArchive/snapshots-france/commit/621725f005318ee21546c454" due to error: "Error: fatal: Unable to create '/home/debian/ota-versions-france/.git/index.lock': File exists.

[...]

PR #766, opened 2022-03-02
Prevent scheduled jobs from running multiple times in parallel

### LINKED ISSUE #756: Prevent short-term blinking
Sometimes, a full deletion and addition are stored in the same run at the same minute, generating two versions instead of no changes.

Example: https://github.com/OpenTermsArchive/versions-france/commit/46a58ae708dec70daf15b444fc1991df352e2568 and https://github.com/OpenTermsArchive/versions-france/commit/98ee801768a46fa257b2aad9d1294837fe49e82a

Ndpnt:
Should fix #756 and probably #753.

Notable changes:
- Remove `node-scheduler` dependency as it is [buggy](https://github.com/node-schedule/node-schedule/issues/635)
- Use native `setInterval` instead of using an external dependency
- Track changes every 6 hours instead of at minute 30 past every 6 hours

[`toad-scheduler`](https://github.com/kibertoad/toad-scheduler) and [`croner`](https://github.com/Hexagon/croner) were considered as `node-scheduler` replacement but native `setInterval` was chose for simplicity even if it does not offer same flexibility, but for now we do not need more than that.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,6 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 import config from 'config';
-import scheduler from 'node-schedule';
 
 import { publishRelease } from '../scripts/release/releasedataset.js';
 
@@ -71,18 +70,23 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
   }
 
   logger.info('The scheduler is running…');
-  logger.info('Documents will be tracked at minute 30 past every 6 hours.');
-  scheduler.scheduleJob('30 */6 * * *', async () => {
+
+  const TRACK_CHANGES_HOURS_INTERVAL = 2;
+  const RELEASE_HOURS_INTERVAL = 24;
+
+  logger.info(`Documents will be tracked every ${TRACK_CHANGES_HOURS_INTERVAL} hours\n`);
+  setInterval(async () => {
     await archivist.trackChanges(serviceIds);
-  });
+  }, TRACK_CHANGES_HOURS_INTERVAL * 60 * 60 * 1000);
 
   if (config.get('dataset.publish')) {
-    logger.info('Release will be created every 24 hours at 04h15');
-    scheduler.scheduleJob('15 4 * * *', async () => {
+    logger.info(`Release will be created every ${RELEASE_HOURS_INTERVAL} hours\n`);
+
+    setInterval(async () => {
       logger.info('Start creating the release…');
       await publishRelease();
       logger.info('Release published');
-    });
+    }, RELEASE_HOURS_INTERVAL * 60 * 60 * 1000);
   }
 }());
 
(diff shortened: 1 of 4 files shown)


martinratinaud (approved):
Good idea!

src/index.js
```diff
@@ -43,7 +42,7 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
       const isServiceDeclared = archivist.serviceDeclarations[serviceId];
 
       if (!isServiceDeclared) {
-        logger.warn(`Parameter "${serviceId}" was interpreted as a service ID to update, but no matching declaration was found. It will be ignored.`);
+        logger.warn(`Parameter "${serviceId}" was interpreted as a service ID to update, but no matching declaration was found. It will be ignored`);
```
MattiSG, line 45:
```suggestion
        logger.warn(`Parameter "${serviceId}" was interpreted as a service ID to update, but no matching declaration was found; it will be ignored`);
```

src/index.js
```diff
@@ -71,18 +70,23 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
   }
 
   logger.info('The scheduler is running…');
-  logger.info('Documents will be tracked at minute 30 past every 6 hours.');
-  scheduler.scheduleJob('30 */6 * * *', async () => {
+
+  const TRACK_CHANGES_HOURS_INTERVAL = 2;
```
MattiSG, line 74: Currently, we generate too much noise on each tracking run. Let's decrease this until we make it configurable.
```suggestion
  const TRACK_CHANGES_HOURS_INTERVAL = 24;
```

src/index.js
```diff
@@ -71,18 +70,23 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+  logger.info(`Documents will be tracked every ${TRACK_CHANGES_HOURS_INTERVAL} hours\n`);
+  setInterval(async () => {
     await archivist.trackChanges(serviceIds);
-  });
+  }, TRACK_CHANGES_HOURS_INTERVAL * 60 * 60 * 1000);
 
   if (config.get('dataset.publish')) {
-    logger.info('Release will be created every 24 hours at 04h15');
-    scheduler.scheduleJob('15 4 * * *', async () => {
+    logger.info(`Release will be created every ${RELEASE_HOURS_INTERVAL} hours\n`);
```
MattiSG, line 83:
```suggestion
    logger.info(`A release will be published every ${RELEASE_HOURS_INTERVAL} hours\n`);
```

src/index.js
```diff
@@ -71,18 +70,23 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
   }
 
   logger.info('The scheduler is running…');
-  logger.info('Documents will be tracked at minute 30 past every 6 hours.');
-  scheduler.scheduleJob('30 */6 * * *', async () => {
+
+  const TRACK_CHANGES_HOURS_INTERVAL = 2;
+  const RELEASE_HOURS_INTERVAL = 24;
```
MattiSG, line 75: Currently, we often publish datasets that need manual cleaning. Let's decrease this until we make it configurable.
```suggestion
  const RELEASE_HOURS_INTERVAL = 24 * 7;
```

MattiSG (approved):

PR #767, opened 2022-03-02
Throw InacessibleContentError on empty matched content

Ndpnt:
Alternative to #751

### DIFF
--- a/src/archivist/filter/index.js
+++ b/src/archivist/filter/index.js
@@ -83,7 +83,7 @@ export async function filterHTML({ content, documentDeclaration }) {
   const markdownContent = transform(domFragment);
 
   if (!markdownContent) {
-    throw new InaccessibleContentError(`The provided selector "${contentSelectors}" match an empty content in the web page at '${location}'`);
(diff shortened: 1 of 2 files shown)


martinratinaud (approved):
Nice thanks

src/archivist/filter/index.js
```diff
@@ -80,7 +80,13 @@ export async function filterHTML({ content, documentDeclaration }) {
     }
   });
 
-  return transform(domFragment);
+  const markdownContent = transform(domFragment);
+
+  if (!markdownContent) {
+    throw new InaccessibleContentError(`The provided selector "${contentSelectors}" match an empty content in the web page at '${location}'`);
```
MattiSG, line 86:
```suggestion
    throw new InaccessibleContentError(`The provided selector "${contentSelectors}" matches no content in the web page at '${location}'`);
```

MattiSG (changes requested):
Can we please add another test with whitespace only? 🙂

MattiSG:
Feel free to dismiss my “request changes” when my request is handled, I might not be reactive enough today 🙂

PR #774, opened 2022-03-10
Add Conditions of Carriage document type

clementbiron:


src/archivist/services/documentTypes.json
```diff
@@ -1,191 +1,198 @@
...
+    }
+  },
+  "Complaints Policy": {
+    "commitment": {
+      "writer": "service provider",
+      "audience": "anyone",
+      "object": "how generic complaints will be handled"
+    }
+  },
+  "Transport Policy": {
```
MattiSG, line 191: Based on some [example airline](https://www.transavia.com/en-EU/conditions/conditions-of-carriage/) website, I found that “Conditions de transport” seem to be “Conditions of Carriage” in English.
When I look this term up in a search engine, I find indeed a [definition](https://www.icao.int/sustainability/Documents/ConsumerInterests_B.pdf) for this term by the ICAO (UN agency for civil aviation), which seems very fit as a source:

>  “Conditions of carriage” mean the terms and conditions established by an air carrier in respect of its carriage, and are filed as applicable tariff rules.

On Wikipedia, I find, for [rail travel](https://en.wikipedia.org/wiki/CIV_(rail_travel)):

> The full current title is the Uniform Rules concerning the Contract of International Carriage of Passengers by Rail (CGT-CIV, [French](https://en.wikipedia.org/wiki/French_language): Conditions générales de transport pour le transport international ferroviaire des voyageurs), formerly the Uniform Rules concerning the Contract for International Carriage of Passengers and Luggage by Rail.

I thus suggest:

```suggestion
  "Conditions of Carriage": {
```

src/archivist/services/documentTypes.json
```diff
@@ -1,191 +1,198 @@
...
+    "commitment": {
+      "writer": "service provider",
+      "audience": "end user",
```
MattiSG, line 194: The ICAO indeed [uses the term “user”](https://www.icao.int/sustainability/Documents/ConsumerInterests_B.pdf) in their definition, but also uses “passenger”. The CIT in the [International Convention for the transportation of Passengers](https://cit-rail.org/secure-media/files/documentation_de/passenger/civ/civ1999-f-d-e.pdf?cid=224515) uses “passenger”. I believe it is more precise and more specific to use this term 🙂

```suggestion
      "audience": "passenger",
```

src/archivist/services/documentTypes.json
```diff
@@ -1,191 +1,198 @@
...
+      "object": "end user’s transport service usage"
```
MattiSG, line 195: Based on the [ICAO’s definition](https://www.icao.int/sustainability/Documents/ConsumerInterests_B.pdf):

> Both conditions of carriage/contract spell out the various benefits and limitations associated with the air transportation being provided. These benefits and limitations, along with the price, constitute a “contract for carriage” between the air carrier and the user.

(this could probably be improved by being made a bit shorter)

```suggestion
      "object": "benefits and limitations associated with the transportation being provided"
```

MattiSG (changes requested):
Why change the indentation of the whole file? 😕 I'd prefer to have a commit with only a clean addition 🙂

src/archivist/services/documentTypes.json
clementbiron: 👍 ✅

src/archivist/services/documentTypes.json
```diff
@@ -1,191 +1,198 @@
...
+  "Complaints Policy": {
+    "commitment": {
+      "writer": "service provider",
+      "audience": "anyone",
+      "object": "how generic complaints will be handled"
+    }
+  },
+  "Transport Policy": {
+    "commitment": {
+      "writer": "service provider",
```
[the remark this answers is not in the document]
clementbiron: 👍

src/archivist/services/documentTypes.json
clementbiron: 👍

clementbiron:
I didn't pay any attention to what the indentation had changed and i think it's because the file was originaly wrong formated.

Thanks for your feedback Matti !

MattiSG (approved):
Nice, thanks @clementbiron!

The best would be to rebase and remove the two commits where one reverts the other 🙂

PR #790, opened 2022-03-22
Disable automatic restart in production

Ndpnt:
In 99% of cases, an automatic restart will not resolve the error as the application will get stuck on unhandled errors, which are bugs in the code. So it's preferable to completely avoid automatic restarts.

Note: pm2 `max-restarts` option does not work as expected and cannot be used here.

### DIFF
--- a/ops/roles/ota/tasks/main.yml
+++ b/ops/roles/ota/tasks/main.yml
@@ -126,7 +126,7 @@
 
 - name: Start Open Terms Archive
   command:
-    cmd: pm2 start --name "{{ ota_app_name }}" --exp-backoff-restart-delay=1000 npm -- run start:scheduler
+    cmd: pm2 start --name "{{ ota_app_name }}" --no-autorestart npm -- run start:scheduler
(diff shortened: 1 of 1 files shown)


martinratinaud:
I find it really harsh to not even try to restart once as it could help us not having interruption of service.

I made some research and here is what I found
- `max_restarts` can only be used with `min_uptime`, otherwise, une environment variable `unstable_restarts` will never be populated AND thus `max_restarts` will never be triggered. See https://github.com/Unitech/pm2/pull/3990/files
- `min_uptime` is not available on the command line and will not be I guess. See https://github.com/Unitech/pm2/issues/2877

So I approve this version but ask @MattiSG if it was not better to create a configuration file in the ops section in order to get this max-restart functionnality

martinratinaud (approved):

MattiSG (approved):
This is acceptable for me and better than the current situation, but I'm in favour of switching to a config file. We already saw that it would enable using `startOrRestart`, and it would just make the options more discoverable. I also fully agree with @martinratinaud's point that not restarting at all is putting a lot of pressure on the team in case of a transient error.

PR #791, opened 2022-03-22
Schedule runs

Ndpnt:


scripts/dataset/main.js
```diff
@@ -1,6 +1,8 @@
-import { release } from './index.js';
+import cron from 'croner';
```
martinratinaud, line 1: I imagine croner will not have the same problems as node-schedule but how can you be sure ?

martinratinaud (approved):

src/main.js
```diff
@@ -61,9 +61,9 @@ const TRACK_CHANGES_HOURS_INTERVAL = 24;
   }
 
   logger.info('The scheduler is running…');
+  logger.info('Documents will be tracked at minute 30 past hour 03, 09, 15, and 21 every day');
 
-  logger.info(`Documents will be tracked every ${TRACK_CHANGES_HOURS_INTERVAL} hours\n`);
-  setInterval(async () => {
+  cron('30 3,9,15,21 * * *', async () => {
```
MattiSG, line 66: This sounds more complicated than necessary to me 🙂 

```suggestion
logger.info('Documents will be tracked every six hours starting at midnight');

  cron('0 0/6 * * *', async () => {
```

src/main.js
```diff
@@ -61,9 +61,9 @@ const TRACK_CHANGES_HOURS_INTERVAL = 24;
...
     await archivist.trackChanges(serviceIds);
```
MattiSG, line 67: Why would we `await` this? 🤔 
```suggestion
    archivist.trackChanges(serviceIds);
```

MattiSG (approved):
I hope this one lib doesn't let us down 😄

But it does seem well documented.

Hexagon:
> I hope this one lib doesn't let us down 😄
> 
> But it does seem well documented.

Should work! If it don't, just drop an issue and we'll sort it out ;)

Issue #788, opened 2022-03-21
Validate JSON syntax in declarations


martinratinaud:
After a wrongly formatted JSON file introduced in `dating` repo, we ended up with 1000s of emails sent as the service was not working correctly anymore.

Despite the error, the CI was showing passing tests

After some digging, I realized that in case of a wrongly formatted JSON, 

`const serviceDeclarations = await services.loadWithHistory();`

is throwing an error that is ignored silently by ota-validate

PR #798, opened 2022-03-23
Fail tests on json error

martinratinaud:
It appears mocha does not throw `unhandledException` s and fails silently

This PR will fix this

Thanks @Ndpnt for the pair

### DIFF
--- a/src/archivist/services/index.js
+++ b/src/archivist/services/index.js
@@ -19,7 +19,14 @@ export async function load() {
   const serviceFileNames = fileNames.filter(fileName => path.extname(fileName) == '.json' && !fileName.includes('.history.json'));
 
   await Promise.all(serviceFileNames.map(async fileName => {
-    const serviceDeclaration = JSON.parse(await fs.readFile(path.join(declarationsPath, fileName)));
+    const jsonDeclarationFilePath = path.join(declarationsPath, fileName);
+    let serviceDeclaration;
+
+    try {
+      serviceDeclaration = JSON.parse(await fs.readFile(jsonDeclarationFilePath));
+    } catch (e) {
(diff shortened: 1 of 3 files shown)


src/archivist/services/index.js
```diff
@@ -19,7 +19,14 @@ export async function load() {
...
 
   await Promise.all(serviceFileNames.map(async fileName => {
-    const serviceDeclaration = JSON.parse(await fs.readFile(path.join(declarationsPath, fileName)));
+    const jsonDeclarationFilePath = path.join(declarationsPath, fileName);
+    let serviceDeclaration;
+
+    try {
+      serviceDeclaration = JSON.parse(await fs.readFile(jsonDeclarationFilePath));
+    } catch (e) {
+      throw new Error(`JSON for document ${jsonDeclarationFilePath} could not be parsed`);
```
Ndpnt, line 28: I suggest:
`The "${path.basename(fileName, '.json')}" service declaration is malformed and cannot be parsed in ${jsonDeclarationFilePath}`

Example:
`The "Facebook" service declaration is malformed and cannot be parsed in /Users/ndpnt/Workspace/OTA/services-all/declarations/Facebook.json`

martinratinaud: I changed even though I believe reusing `path.basename(fileName, '.json')` is overkill

martinratinaud:
Tests should pass once https://github.com/OpenTermsArchive/services-all/pull/524 is merged

martinratinaud:
@Ndpnt I added your suggestions

Ndpnt (approved):

martinratinaud:
Fixes #788

PR #800, opened 2022-03-23
Switch snapshots on `france` instance to Git storage

Ndpnt:


MattiSG (approved):
I assume it is voluntary to rely on the default value for `type` instead of explicitly stating `git` 🙂

PR #802, opened 2022-03-24
Use different mailing lists for elections-france

martinratinaud:
Fixes https://github.com/OpenTermsArchive/versions-france-elections/issues/1


Ndpnt (approved):

Ndpnt:
I fixed the url of the main button in the mail template but now is ok.

PR #804, opened 2022-03-24
Download entire history on instance deployment

### LINKED ISSUE #792: Download entire history on instance setup
Fix https://github.com/ambanum/OpenTermsArchive-versions/issues/9

martinratinaud:
Also

- [x] test deployment on vagrant
- [x] stop ota on `services-all` server
- [x] remove the snapshot repository from the `services-all` server
- [x] remove the versions repository from the `services-all` server
- [x] merge and deploy
- [ ] verify that dataset contains all data

Fixes #792

### DIFF
--- a/ops/roles/ota/tasks/database.yml
+++ b/ops/roles/ota/tasks/database.yml
@@ -19,7 +19,6 @@
     repo: '{{ repository }}'
     version: '{{ branch }}'
     dest: '/home/{{ ansible_user }}/{{ directory }}'
-    depth: 1
     accept_hostkey: yes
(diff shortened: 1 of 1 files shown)


Ndpnt (approved):

martinratinaud:
Just checked and we have
- 11511 version commits
- 1056779 snapshots commits which I'm not sure should even be there as we have this in mongo now

MattiSG:
> snapshots commits which I'm not sure should even be there as we have this in mongo now

I guess ops doesn't know about the snapshots format and always tries to set it up… @Ndpnt is there a way to specify to not set up snapshots?

Ndpnt:
No, not at the moment.

Issue #794, opened 2022-03-23
Unify repository naming


MattiSG:
- `services-…` → `declarations-…`
- Prefix with ID instead of suffixing.
- `experimental` → `experiment` in ops config.
- Document repository naming pattern in CONTRIBUTING.
- `all` → `contrib` 
- `ambanum/OpenTermsArchive-versions` → `OpenTermsArchive/contrib-versions`

martinratinaud:
Before merging, some actions are necessary

**NOTE** I propose to move those folders instead of removing them to save the time of redownloading when deploying

- [ ] rename services-all to contrib-declarations
- [ ] rename ambanum/OpenTermsArchive-versions to OpenTermsArchive/contrib-versions
- [ ] rename ambanum/OpenTermsArchive-snapshots to OpenTermsArchive/contrib-snapshots
- [ ] rename declarations-france to france-declarations with snapshots and versions
- [ ] rename declarations-france-elections to france-declarations-elections with snapshots and versions
- [ ] rename services-dating to dating-declarations with snapshots and versions

- [ ] stop OTA on `services-all` and rename folders
```
mv ota-services-all ota-contrib-declarations
mv ota-snapshots ota-contrib-snapshots
mv ota-versions ota-contrib-versions
```

- [ ] stop OTA on `dating` and rename folders 
```
mv ota-versions-dating ota-dating-versions
mv ota-snapshots-dating ota-dating-snapshots
mv ota-services-dating ota-dating-declarations
```

- [ ] stop OTA on `france` and rename folders 
```
mv ota-versions-france ota-france-versions
mv ota-snapshots-france ota-france-snapshots
mv ota-services-france ota-france-declarations
```

- [ ] stop OTA on `france-elections` and rename folders 
```
mv ota-versions-france-elections ota-france-elections-versions
mv ota-snapshots-france-elections ota-france-elections-snapshots
mv ota-services-france-elections ota-france-elections-declarations
```

- [ ] merge https://github.com/OpenTermsArchive/declarations-france/pull/318
- [ ] merge https://github.com/OpenTermsArchive/services-dating/pull/77
- [ ] merge https://github.com/OpenTermsArchive/declarations-france-elections/pull/86
- [ ] no PR needed for services-all (did not find any reference to a github account name)


## On Local

You will need to

- rename your local folder from `services-all` to `contrib-declarations`
- rename your local folder from `services-dating` to `dating-declarations`
[...]

PR #808, opened 2022-03-28
Unify naming

martinratinaud:
Fixes https://github.com/ambanum/OpenTermsArchive/issues/794


CONTRIBUTING.md
```diff
@@ -11,6 +12,21 @@ First of all, thanks for taking the time to contribute! 🎉👍
   - [Errors handling](#errors-handling)
 
 ---
+
+## New instance naming
+
+In case you want to create a new instance on the `OpenTermsArchive` organization, please choose an instance name as descriptive as possible.
+
+Here are some examples of existing instances and the kind of declaration files they contain:
```
MattiSG, line 20: I think the point of this documentation should be to explain what makes an instance “descriptive” or “understandable”, not so much to insist on the fact that it should be 😉 The example we give below do not explain in which way they are descriptive.

ops/inventories/dev.yml
```diff
@@ -1,18 +1,24 @@
 all:
+  vars:
+    ansible_user: vagrant
+    snapshots_branch: main
+    versions_branch: main
+    snapshots_directory: snapshots
+    versions_directory: version
+    declarations_directory: declarations
+    app_directory: ota
```
MattiSG, line 9: > Why do we need to add this now here? If it was not needed before, do we really need to duplicate these variables?

ops/roles/ota/vars/main.yml
```diff
@@ -1,2 +0,0 @@
-ota_app_name: ota
-ota_repository: https://github.com/ambanum/OpenTermsArchive.git
```
MattiSG, line 2: If this was a shared default between production and dev, why not keep it? 🙂

MattiSG (changes requested):
Thanks for this changeset! I'll create a separate PR based on this one to suggest documentation 🙂

config/ota-contrib-declarations-production.json
```diff
@@ -1,12 +1,12 @@
...
   "services": {
-    "declarationsPath": "../ota-services-all/declarations"
+    "declarationsPath": "../ota-contrib-declarations/declarations"
   },
   "recorder": {
     "versions": {
       "storage": {
         "git": {
-          "path": "../ota-versions",
+          "path": "../ota-contrib-versions",
```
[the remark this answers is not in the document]
martinratinaud: Good catch, forgot those ones

ops/roles/ota/vars/main.yml
martinratinaud: I believed it was really weird to have only those 2 variables in another lonely file

More over, the ota repository and app name should be changeable
- app name to be able to launche several instances on the same machine
- repository to make it possible to use a fork

Again, having dispateched variables feels not right to me but I can rollback if needed

Ndpnt: I agree with Martin, I think the factorization he made it's a good improvement. 
It is easier to discover the variables and it is possible to override these variables for each instance if necessary.

config/ota-contrib-declarations-production.json
```diff
@@ -1,12 +1,12 @@
 {
   "services": {
-    "declarationsPath": "../ota-services-all/declarations"
+    "declarationsPath": "../declarations/declarations"
```
Ndpnt, line 3: There is duplications between instance production config files.
I suggest to create a `production.json` config file with the factorized configs keys shared by all instances and rename `ota-contrib-declarations-production` to `contrib` (and so on…) with only the overridden keys.

config/ota-contrib-declarations-production.json
```diff
@@ -1,12 +1,12 @@
 {
```
Ndpnt, line 1: Renaming these files will break deployment as they match the hostname of each instance.
I still think it's a good idea to rename these files but you will have to also rename hostnames on each instances.
I also suggest to rename hostname like the following:
- ota-contrib-declarations --> contrib
- ota-dating-declarations --> dating
- ota-france-declarations  --> france
- ota-france-elections-declarations --> france-elections

MattiSG: That would be really good indeed! And part of this unification effort 👍

MIGRATING.md
```diff
@@ -1,13 +1,13 @@
 ## 2021-11-25 Externalize services declarations
 
-Services declarations are now declared in their [own repository](https://github.com/OpenTermsArchive/services-all).
+Services declarations are now declared in their [own repository](https://github.com/OpenTermsArchive/contrib-declarations).
```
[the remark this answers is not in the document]
martinratinaud: Done

config/ota-contrib-declarations-production.json
martinratinaud: Done

config/ota-contrib-declarations-production.json
martinratinaud: DONE

I added a new variable `config_file_name` because relying on the hostname of the machine (that needs a by-hand configuration) is too magical for me

ops/inventories/dev.yml
martinratinaud: As explained in other comment
I find it less confusing to have all deployment related variables at the same spot.

In the dev.yml is where you might want to make some tests when deploying on vagrant.
I believe all config variables should be listed there

martinratinaud:
@MattiSG @Ndpnt 

I added some commits for
- factorize all production config in one file
- add entry in the migration guide
- specifying the config you want to use for deployment in ops

I tested by deploying on vagrant 
- with a custom `vagrant.json` that overrides correctly other config items
- I then removed this file to test without any config file (as it is the case for dev now)

Seems to be all fine

Ndpnt (approved):

PR #810, opened 2022-03-28
Rename Orange Money France

clementbiron:


MattiSG:
This is indeed the destination name, but I am not sure how we should handle the fact that this service _used to be actually called_ Orange Money France. Should we rename it for all of its history or only from the moment it has been renamed?

In any case, I need a review from @Ndpnt since I don't know how the renamer works.

MAKA02:
carding and loading is an act to be done by legitimate people 
need verification on how to get quick cc , PayPal transfers and more ?
have you been looking for a place to buy cc and get legit money TRANSFER services like PayPal,  western union and more ?
visit https://t.me/credit_card_uk

MattiSG:
We'll instead consider that the service `Orange Money France` closed at some date, and that another service called `Orange Money Europe` opened at the same date. This will not allow tracking changes from this service to the other and makes us lose the information that they are the same in some sense, but this seems better than losing the historical name of the original service.

Issue #805, opened 2022-03-24
Document retrieved is different from one time to another


clementbiron:
The document retrieved for SHEIN Commercial Terms is sometimes the mobile version, sometimes the desktop version. View the associated diff https://github.com/OpenTermsArchive/versions-france/commit/e4742be0c637500454bf2f8c0d1b522fad7dff00 and https://github.com/OpenTermsArchive/versions-france/commit/a9cca60e67e855c66009b823b6b5f1c94793ff6d

As changes are discovered, we need to ensure that we consistently provide the same version of a document.

clementbiron:
Yes 👍

MattiSG:
Indeed, it would be good to restrict the device category to stabilise the documents we receive 🙂
But, more generally, why do we generate a random user agent altogether? I understand the intention, but do we have any idea how many documents less we would get by using a legit `Open Terms Archive` user agent? 🙂

MattiSG:
I wonder how much this is also the source of the blinking Meta pages such as in https://github.com/OpenTermsArchive/france-elections-versions/commit/709edec29cd516aa33bee475a7bea071d9fae029. Indeed, beyond the `?locale2=fr_FR` query strings, I also see that some [destination domains](https://github.com/OpenTermsArchive/france-elections-versions/commit/709edec29cd516aa33bee475a7bea071d9fae029#diff-e43a59590f4f051a818725f7b8453d6db4b26ce8ad524d71bc8d8293d72bc28cR248) switch to a `m` subdomain.

Ndpnt:
I am in favour of removing the random user agent generation. Also, I suggest avoiding randomness in code as much as possible: Random things lead to random bugs, which are very difficult to debug.

martinratinaud:
This is part of best practices of scraping

https://www.blog.datahut.co/post/web-scraping-best-practices-tips

Also, I think I implemented it because I had problems on a declaration (but did not document it)

i'd rather fix it to one version than remove it completely, or use "OpenTermsArchive/v1" but I'm pretty sure many servers will kick us out with this

clementbiron:
By removing the user agent generation, the risk of getting kicked increases considerably.

Ndpnt:
Currently, random generation of the user agent is only performed for documents retrieved with the headless browser. If we remove it, it can only affect these documents, which is :
- 72/607 (~12%) documents for `contrib`
- 38/209 (~18%) documents for `france`.
[...]

PR #814, opened 2022-03-29
Remove random user agent generation

Ndpnt:
Related to https://github.com/ambanum/OpenTermsArchive/issues/805

### DIFF
--- a/src/archivist/fetcher/fullDomFetcher.js
+++ b/src/archivist/fetcher/fullDomFetcher.js
@@ -1,7 +1,6 @@
 import config from 'config';
 import puppeteer from 'puppeteer-extra';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
-import UserAgent from 'user-agents';
 
 import { FetchDocumentError } from './errors.js';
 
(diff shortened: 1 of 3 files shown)


src/archivist/fetcher/fullDomFetcher.js
```diff
@@ -25,7 +24,6 @@ export default async function fetch(url, cssSelectors) {
   try {
     page = await browser.newPage();
 
-    await page.setUserAgent(new UserAgent().toString());
```
MattiSG, line 28: Could we please use `OpenTermsArchive/${package.version}`? 🙂
I'd like our crawlers to start appearing in logs.

MattiSG (commented):
Thanks @Ndpnt!

Before shipping this to production, could we please run it locally (or on an experiment server) once and measure how many documents are not fetched correctly anymore? Either on Contrib or on France, where we have more certainty on the maintenance level, ideally on both.

src/archivist/fetcher/fullDomFetcher.js
MattiSG: Suggestion made in #815.

martinratinaud, line 28: I think this is a bad idea

See https://brightdata.com/blog/how-tos/user-agents-for-web-scraping-101

And I've read many articles that all explained that scraping should be done with a user-agent.

I believe if we remove this, we will be kicked by many servers

You'd better use it with `{"deviceCategory": "desktop"}`

martinratinaud (changes requested):

src/archivist/fetcher/fullDomFetcher.js
Ndpnt: See https://github.com/ambanum/OpenTermsArchive/issues/805#issuecomment-1082793637

PR #815, opened 2022-03-29
Use Open Terms Archive user agent across fetcher implementations

MattiSG:
Considering the lack of current problems on the HTML fetcher with no user agent at all, I doubt this would make things worse 🙂

### DIFF
--- a/src/archivist/fetcher/fullDomFetcher.js
+++ b/src/archivist/fetcher/fullDomFetcher.js
@@ -3,6 +3,7 @@ import puppeteer from 'puppeteer-extra';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
 
 import { FetchDocumentError } from './errors.js';
+import USER_AGENT from './userAgent.js';
 
 puppeteer.use(stealthPlugin());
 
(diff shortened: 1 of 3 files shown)


MattiSG:
This is a suggestion for #814, I probably won't have the time to make this go all the way to the end properly. @Ndpnt feel free to re-implement it however you see fit if you agree with the intention 🙂

martinratinaud:
> Considering the lack of current problems on the HTML fetcher with no user agent at all, I doubt this would make things worse 🙂

What makes you say that?

I remember putting this in because
- I had problem on a document (but did not document it, sorry about that)
- this is common best practices -> https://www.blog.datahut.co/post/web-scraping-best-practices-tips

martinratinaud (changes requested):
Let's discuss this then

Ndpnt:
See https://github.com/ambanum/OpenTermsArchive/issues/805#issuecomment-1082793637

martinratinaud (approved):
As discussed, I'm fine with that

Ndpnt:
Could we close this PR @MattiSG?

Ndpnt:
It's a long lived PR…

MattiSG:
Okay.

PR #824, opened 2022-04-07
Enable auto restart in production

### LINKED ISSUE #821: Delay restarts
The exponential backoff restart strategy currently implemented is capped at 15 seconds to restart instances, which makes it easy to saturate.

We will restart 3 times at 15 minutes interval. This implies switching `pm2` to a config file.

Ndpnt:
Fixes #821.

### DIFF
--- a/pm2.config.cjs
+++ b/pm2.config.cjs
@@ -1,0 +1,20 @@
+module.exports = {
+  apps: [
+    {
+      name: 'ota',
+      script: 'npm',
+      args: 'run start:scheduler',
+      max_restarts: 2,
+      min_uptime: '1h', // Set a relatively high duration (more than the longest run) so that restarts that occur before this duration has elapsed are considered unstable.
+      restart_delay: 15 * 60 * 1000, // 15min
+    },
+    {
+      name: 'ota-release',
+      script: 'npm',
+      args: 'run dataset:scheduler',
+      max_restarts: 2,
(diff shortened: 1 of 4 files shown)


pm2.config.cjs
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
martinratinaud, line 1: Can you please explain why you are using a .cjs file instead of a plain .js one?

I do not see anything relative to that in https://pm2.keymetrics.io/docs/usage/application-declaration/

Thanks

martinratinaud (approved):

pm2.config.cjs
Ndpnt: Because if we do not use a `cjs`file we have this following error:
```
[PM2][ERROR] File pm2.config.js malformated
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/ndpnt/Workspace/OTA/OpenTermsArchive/pm2.config.js
require() of ES modules is not supported.
require() of /Users/ndpnt/Workspace/OTA/OpenTermsArchive/pm2.config.js from /Users/ndpnt/.nvm/versions/node/v14.18.0/lib/node_modules/pm2/lib/Common.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename pm2.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/ndpnt/Workspace/OTA/OpenTermsArchive/package.json.
```

PR #827, opened 2022-04-13
Migrate sendinblue

martinratinaud:


MattiSG (approved):
Every ID matches what I find in the dashboard. Thank you for renaming the `france` list to `france-elections`!

For the API keys, I checked them with  `ansible-vault view`. I could only match the last 5 characters or so since the rest is hidden in SIB's dashboard, but unless there was a problem with copy-pasting it should be fine 🙂

Since the current setup still works, I suggest to merge only at a moment where several people are on duty to handle a potential production problem. @martinratinaud if you are off tomorrow, I suggest we wait until next Monday 😄

martinratinaud:
Will do this today

martinratinaud:
Just received emails of contrib

I also think a nice thing to add would be to build a digest on every run, instead of a mail per service update

martinratinaud:
I still receive emails from services-all
I suspect it is the experimentations website but can't confirm before I leave sorry about that

PR #834, opened 2022-04-28
Set Git username and email for repos handled by Git storage adapter

Ndpnt:
Ensure author and committer are the always the same


martinratinaud (approved):
Should we remove the `const options = { '--author': `${this.author.name} <${this.author.email}>` };` then.

I believe this acts as a duplicate

Issue #837, opened 2022-05-03
Automatically create GitHub issues creation labels


MattiSG:
The automated GitHub issue creation mechanism relies on tags being set on the automated issues. These tags should be created automatically as they lead otherwise to duplicated issues each time we create a new instance.

The default tag value should be changed to `bot-report` since `update` is duplicate information with the name of the issue.

martinratinaud:
@Ndpnt here is the PR https://github.com/ambanum/OpenTermsArchive/issues/837

I currently create the label at instantiation of the Tracker. As it is in a constructor, promise is not awaited (I believe this is not a problem as first error creation should occur way after this promise is resolved)

That leads to the following question:
What should we do if label creation fails ? There is today an error displayed in the logs but I'm wondering if it's sufficient.

Handling this would make the code slightly more complicated, here are 2 solutions:
1. keep track of the created label id and, if label creation did not work, retry within the `onVersionRecorded`, `onVersionNotChanged` and `onInaccessibleContent`. This would lead to more resilience as the odds of having a fail twice are very thin.
2. always create label in `onVersionRecorded`, `onVersionNotChanged` and `onInaccessibleContent` (Not recommended as we would have a lot of unnecessary calls to the API)

I think leaving it this way is fine as the odds of having an error from Github at setup are very thin and the worst that can happen is some duplicate issues. 

Please let me know what you think are have a chat about it

martinratinaud:
Here are the actions to take

- [ ] Deploy on all instances
- [ ] Update tag on PGA (add `bot-report` tag to all issues with `update` tag and remove `update` tag)
- [ ] Rename tag on contrib
- [ ] Rename tag on france
- [ ] Rename tag on france-elections
- [ ] Rename tag on dating
- [ ] Create `add` tag automatically on contribution tool

PR #840, opened 2022-05-04
Create GitHub labels

martinratinaud:
Automatically create required Github labels in declaration repository to prevent creation of duplicates

Fixes https://github.com/ambanum/OpenTermsArchive/issues/837

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -57,7 +57,7 @@
   "tracker": {
     "githubIssues": {
       "repository": "OpenTermsArchive/sandbox",
-      "label": "update"
+      "label": "bot-report",
     }
   },
   "dataset": {
--- a/config/default.json
+++ b/config/default.json
@@ -59,7 +59,7 @@
       "repository": "OpenTermsArchive/sandbox",
       "label": "bot-report",
       "color": "FEF2C0",
-      "description": "automatically created by Open Terms Archive when fetching error is encountered"
+      "description": "Automatically created by Open Terms Archive when fetching error is encountered"
     }
   },
   "dataset": {
--- a/config/default.json
+++ b/config/default.json
@@ -58,6 +58,8 @@
     "githubIssues": {
       "repository": "OpenTermsArchive/sandbox",
       "label": "bot-report",
+      "color": "FEF2C0",
+      "description": "automatically created by Open Terms Archive when fetching error is encountered"
     }
   },
   "dataset": {
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -93,15 +93,13 @@ export default class Tracker {
   }
 
   async createLabel(params) {
-    try {
-      await this.octokit.rest.issues.createLabel({ ...this.commonParams, ...params });
-    } catch (e) {
-      if (!e.toString().includes('"code":"already_exists"')) {
-        // log error when error is unknown
-        logger.error(`Could not create label ${params.name}`);
-        logger.error(e.toString());
-      }
-    }
+    return this.octokit.rest.issues.createLabel({ ...this.commonParams, ...params })
+      .catch(error => {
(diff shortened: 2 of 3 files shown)


src/tracker/index.js
```diff
@@ -40,6 +40,12 @@ export default class Tracker {
       repo,
       accept: 'application/vnd.github.v3+json',
     };
+
+    this.createLabel({
```
Ndpnt, line 44: Move this call inside an `initialize` function which should be executed by the caller.

```
async initialize() {
    await this.createLabel({
      name: config.get('tracker.githubIssues.label'),
      color: config.get('tracker.githubIssues.color'),
      description: config.get('tracker.githubIssues.description'),
    });
}
```

src/tracker/index.js
```diff
@@ -82,6 +88,18 @@ export default class Tracker {
     });
   }
 
+  async createLabel(params) {
+    try {
```
Ndpnt, line 92: As there is only one instruction inside the `try`I suggest to use a `.catch`.

src/tracker/index.js
```diff
@@ -82,6 +88,18 @@ export default class Tracker {
...
+      await this.octokit.rest.issues.createLabel({ ...this.commonParams, ...params });
+    } catch (e) {
+      if (!e.toString().includes('"code":"already_exists"')) {
+        // log error when error is unknown
+        logger.error(`Could not create label ${params.name}`);
+        logger.error(e.toString());
+      }
+    }
+  }
```
Ndpnt, line 101: Suggestion with all previous comments:

```suggestion
  async createLabel(params) {
    return this.octokit.rest.issues.createLabel({ ...this.commonParams, ...params }).catch(error => {
      if (error.toString().includes('"code":"already_exists"')) {
        return;
      }

      logger.error(`Could not create label "${params.name}": ${error.toString()}`);
    });
  }
```

src/tracker/index.js
```diff
@@ -82,6 +88,18 @@ export default class Tracker {
     });
   }
 
+  async createLabel(params) {
+    try {
+      await this.octokit.rest.issues.createLabel({ ...this.commonParams, ...params });
+    } catch (e) {
+      if (!e.toString().includes('"code":"already_exists"')) {
```
Ndpnt, line 95: Prefer early exit

src/tracker/index.js
```diff
@@ -82,6 +88,18 @@ export default class Tracker {
...
+        // log error when error is unknown
```
Ndpnt, line 96: Delete the comment that does not provide significant information.

src/tracker/index.js
```diff
@@ -82,6 +88,18 @@ export default class Tracker {
...
+        logger.error(`Could not create label ${params.name}`);
```
Ndpnt, line 97: Log in one line.

config/default.json
```diff
@@ -57,7 +57,9 @@
   "tracker": {
     "githubIssues": {
       "repository": "OpenTermsArchive/sandbox",
-      "label": "update"
+      "label": "bot-report",
+      "color": "FEF2C0",
+      "description": "automatically created by Open Terms Archive when fetching error is encountered"
```
Ndpnt, line 62: Uppercase the first letter

Ndpnt (changes requested):

src/tracker/index.js
```diff
@@ -40,6 +40,16 @@ export default class Tracker {
       repo,
       accept: 'application/vnd.github.v3+json',
     };
+
+    this.initialize();
```
Ndpnt, line 44: The initialization should be executed by the caller not the constructor. See https://github.com/ambanum/OpenTermsArchive/pull/847

Issue #833, opened 2022-04-22
Lint declaration files


MattiSG:
Declaration validation does not lint files.

### What happened

I tried to validate with `npm run validate Example` this misindented declaration:

```json
{
"name": "Example",
    "documents": {
    "Terms of Service":
    {
    "fetch": "https://domain.example/tos",
    "select": "body"
  }
}
}
```

And I got no error.

### What I expected

An error in the style of:

```
Indentation is incorrect: expected 2 spaces on line 2
```

martinratinaud:
After some investigations and a lot of testing, here is what I found

- Could not make https://github.com/azeemba/eslint-plugin-json#readme work as it does not have indentation rules
- Tried `jsonlint` with no success as detail of errors were not sufficient
- Same with `json-lint`
- Considered using `prettier`but did not want to ad another (not-eslint) dependency
- Found https://github.com/kuceb/eslint-plugin-json-format#readme which works as expected

Here is an example of what it looks like when launching `npm run test Aigle` from `france-declarations`
<img width="1266" alt="Screenshot 2022-05-05 at 14 46 00" src="https://user-images.githubusercontent.com/4191809/166908385-3b7f6a77-6c66-42d6-a482-a46dd32ebd5c.png">

I will try now to add an autofix feature.
My first idea is to use it with 

`npm run test Aigle --fix`

MattiSG:
Thanks @martinratinaud!

The error message depicted above does not seem very clear to me 😕

Please prioritise clear error messages over any auto fix feature 🙂

MattiSG:
The goal here is simply to prevent misformated files from being put in production. Let's not increase too much the API scope.

martinratinaud:
@MattiSG I'm not sure to understand why you think this is not clear enough

For example a normal eslint error is displayed like this
<img width="583" alt="Screenshot 2022-05-05 at 15 10 51" src="https://user-images.githubusercontent.com/4191809/166911910-473e50aa-4988-4738-8445-dfdca7c36dc9.png">

In My Opinion, the current error states
- the path of the file concerned
- the line(s) concerned
- the unexpected character on this particular line

Considering the above, can you please elaborate what you find not clear enough?

In the meantime, possible solutions to make it clearer:
- I'm not sure how to get the coloring with chai but I can look into that.
- I can also display the expecting json to copy paste in the file
- if the above is good enough, then no auto fixing is needed

martinratinaud:
Note also that 
[...]

PR #841, opened 2022-05-05
Json lint

martinratinaud:
Fixes https://github.com/ambanum/OpenTermsArchive/issues/833


package.json
```diff
@@ -59,6 +55,11 @@
     "cross-env": "^7.0.3",
     "datauri": "^4.1.0",
     "dotenv": "^10.0.0",
+    "eslint": "^8.2.0",
+    "eslint-config-airbnb-base": "^15.0.0",
+    "eslint-plugin-chai-friendly": "^0.7.2",
```
MattiSG, line 60: > Do we really need all the plugins to be `dependencies` as well? 🙂

martinratinaud: Yes

open-terms-archive is meant to be a dev dependency (or dependency) in declarations folders.
In order for eslint to be installed, it has to be a real dependency of the project.

I did not find any other solution. @Ndpnt maybe?

Ndpnt: See #842, I moved back eslint in `devDependencies`.

Ndpnt (changes requested):

martinratinaud:
Merged after discussion with @MattiSG 

Further discussion can be done on the format of the JSON and specifically the display of single item arrays, but it will be part of another PR, this one aiming at bring linting at once

Ndpnt:
Perhaps we should just avoid using an array for a single element, as the engine can handle both cases, with or without an array.

MattiSG:
We briefly had this conversation with @martinratinaud indeed, and concluded two things:

1. @martinratinaud wasn't such a huge fan of using commas as delimiter inside a string. My counter-argument to that was that we cannot prevent CSS selectors from supporting commas syntax (unless we validate it manually but I believe this is really too much), but we can remove support for multiple selectors in an array. I'm always in favour of having only one way to do things, so suggesting to use the CSS selector syntax over the array had my preference.
2. This was over the scope of this PR, so we postponed this to a later discussion 🙂 

One thing that I wasn't sure about was: can we really drop support for arrays? Isn't it needed to use the `before` / `after` syntax?

If it is technically doable, then we could open an RFC for it 🙂

Ndpnt:
I was not proposing to completely remove the support for arrays, just not to use arrays when you have a single selector, because that's when it's especially ugly.
And yes array is necessary for `range selector` (`before/after` syntax).

PR #845, opened 2022-05-11
Document GitHub issues label creation and improve associated config

martinratinaud:


README.md
```diff
@@ -198,7 +198,7 @@ The default configuration can be found in `config/default.json`.
   "tracker": { // Tracking mechanism to create GitHub issues when document content is inaccessible
     "githubIssues": {
       "repository": "GitHub repository where to create isssues",
-      "label": "GitHub issues label to attach to a newly created issue. This specific label has to exist in the repository",
+      "label": "GitHub issues label to attach to a newly created issue. This specific label will be created automatically in the target repository",
```
Ndpnt, line 201: Can you use this PR to change the configuration like this:
```
"tracker": {
    "githubIssues": {
      "repository": "OpenTermsArchive/sandbox",
      "label": {
        "name": "bot-report",
        "color": "FEF2C0",
        "description": "Automatically created by Open Terms Archive when fetching error is encountered"
      }
    }
  },
  ```

Comment (martinratinaud) [reply README.md:201]: Done @Ndpnt I also renamed the variables to be more coherent

Comment (Ndpnt) [reply README.md:201]: You forgot to update the README file.

File: config/default.json [hunk 2]
```diff
@@ -57,9 +57,11 @@
   "tracker": {
     "githubIssues": {
       "repository": "OpenTermsArchive/sandbox",
-      "label": "bot-report",
-      "color": "FEF2C0",
-      "description": "Automatically created by Open Terms Archive when fetching error is encountered"
+      "label": {
+        "name": "bot-report",
+        "color": "FEF2C0",
+        "description": "Automatically created by Open Terms Archive when fetching error is encountered"
```
MattiSG, line 63:
```suggestion
        "description": "Automatically created by Open Terms Archive when a document cannot be fetched"
```

src/tracker/index.js
```diff
@@ -11,24 +11,25 @@ const ISSUE_STATE_CLOSED = 'closed';
...
 const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/en/service';
 const CONTRIBUTE_URL = 'https://contribute.opentermsarchive.org/en/service';
-const GITHUB_REPO_URL = `https://github.com/${config.get('tracker.githubIssues.repository')}/blob/main/declarations`;
+const GITHUB_REPO_URL = `https://github.com/${TRACKER_REPOSITORY}/blob/main/declarations`;
 const GOOGLE_URL = 'https://www.google.com/search?q=';
 
 export default class Tracker {
   static isTokenValid() {
-    return (config.get('tracker.githubIssues.repository') || '').includes('/');
+    return (TRACKER_REPOSITORY || '').includes('/');
```
MattiSG, line 24: > This doesn't seem to do what the function name states it does 😅

MattiSG (approved):
This is a breaking change for all reusers having their own config file. Time to move to semver soon 😉

README.md
MattiSG: That was quite a huge scope creep. I agree with the change, and I don't think it will have major consequences from what we know of current reusers, but let's please keep in mind that any change to the config can break down instances that don't update their config accordingly and be careful about it 🙂

PR #847, opened 2022-05-11
Wait for the GitHub tracker to initialise before continuing

Ndpnt:


martinratinaud (changes requested):
I'm not really found of an `initialize` function that you may forget to include.

A better way I believe is to use "Static Async Factory Functions" (Detailed explanation https://dev.to/somedood/the-proper-way-to-write-async-constructors-in-javascript-1o8c)

This would lead to 

```
const tracker = await Tracker.initialize()
```
with initialize which would return the cconstructor.

What do you think?

Ndpnt:
After discussion, we agreed that calling `initialize` function it not totally satisfying but it is better than before. Static Async Factory Functions before has also drawbacks. 
Moreover if we find a better solution we should apply it to the whole codebase, not just this module.

martinratinaud (approved):

PR #849, opened 2022-05-12
Document bot SSH key usage

MattiSG:


clementbiron (approved):

ops/README.md
```diff
@@ -59,6 +59,10 @@ For example, you can update `Open Terms Archive` by running:
 ansible-playbook ops/app.yml --tags update
 ```
 
+### Allowed keys
+
+Some instances might have user-specific SSH keys enabled. But if they are automatically deployed in CI, they will all have the bot key enabled. If your access is denied on an instance, you can thus run any of the commands with the `--private-key` option, passing it the bot SSH private key. This key can be found in the passwords database.
+
```
martinratinaud, line 65: I'm not sure to understand the phrasing of this.

> Some instances might have user-specific SSH keys enabled. But if they are automatically deployed in CI, they will all have the bot key enabled

Maybe 

```
In case the instance you're targetting is hosted by the core team of OTA, you can use the deploy commands with the `--private-key` option, passing it the bot SSH private key. This key can be found in the passwords database.

````

martinratinaud (changes requested):

ops/README.md
MattiSG: Much better indeed, thanks @martinratinaud!

martinratinaud (approved):
perfect, thanks

PR #850, opened 2022-05-13
Fix typo

MattiSG:


martinratinaud (approved):

martinratinaud:
@MattiSG Careful that your tests do not pass though. One missing change.

martinratinaud:
@MattiSG I've fixed the tests and rebased.

Should be ok to merge now

MattiSG:
Thanks @martinratinaud for having fixed this! Unfortunately we're still stuck with a failing CI following #841.

For future times: I would have liked to see what I had missed by reading your own commit independently from the original one I had made, so I can avoid making the same sort of mistake ☺️

martinratinaud:
Sorry about the fixup :-) I wanted to keep the history clean but understand your point.

For reference, you forgot one occurence of `witout`, you can see it here https://github.com/ambanum/OpenTermsArchive/runs/6424782425?check_suite_focus=true

PR #852, opened 2022-05-16
Expose features

martinratinaud:
In order to use open-terms-archive core features as a package import, we need to expose those features.


martinratinaud:
Closed as not yet finished

PR #859, opened 2022-05-18
Automatically set the file extension to use with the git adapter

Ndpnt:


src/storage-adapters/git/index.js
```diff
@@ -13,12 +13,13 @@ import Git from './git.js';
...
 export default class GitAdapter {
-  constructor({ path, fileExtension, author, publish, prefixMessageToSnapshotId }) {
+  constructor({ path, author, publish, prefixMessageToSnapshotId }) {
     this.path = path;
     this.author = author;
-    this.fileExtension = fileExtension;
     this.needsPublication = publish;
     this.prefixMessageToSnapshotId = prefixMessageToSnapshotId;
+
+    mime.define({ 'text/markdown': ['md'] }, true); // ensure extension for markdown files is `.md` and not `.markdown`
```
martinratinaud, line 22: Does this has to be in the constructor?
Seems to me it's more global and should be below imports

src/storage-adapters/git/index.test.js
```diff
@@ -45,7 +45,6 @@ describe('GitAdapter', () => {
     subject = new GitAdapter({
```
martinratinaud, line 45: subject has been a bit confusing for me to understand.

What about `adapter`?

martinratinaud (approved):
LGTM

src/storage-adapters/git/index.test.js
Ndpnt: It's a common practice used through the whole codebase so I won't modify it here but we could discuss about this if we want to change this globally.

PR #860, opened 2022-05-18
Return long hash on commit

Ndpnt:
Ensure to always handle ids in the same format to facilitate comparison.

### DIFF
--- a/src/storage-adapters/git/git.js
+++ b/src/storage-adapters/git/git.js
@@ -54,7 +54,7 @@ export default class Git {
     const shortHash = summary.commit.replace('HEAD ', '').replace('(root-commit) ', '');
     const longHash = (await this.git.show([ shortHash, '--pretty=%H', '-s' ])).trim();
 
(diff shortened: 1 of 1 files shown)


src/storage-adapters/git/git.js
```diff
@@ -47,7 +47,14 @@ export default class Git {
...
 
-    return summary.commit.replace('HEAD ', '').replace('(root-commit) ', '');
+    if (!summary.commit) { // Nothing committed, no hash to return
+      return;
+    }
+
+    const shortHash = summary.commit.replace('HEAD ', '').replace('(root-commit) ', '');
+    const longHash = (await this.git.show([ shortHash, '--pretty=%H', '-s' ])).trim();
+
+    return longHash; // Return a long commit hash to always handle ids in the same format and facilitate comparison as other commands return a long hash
```
martinratinaud, line 57: I'd remove ` as other commands return a long hash` which does not bring much more value.
Goal is to handle ids in same format and this function does it

martinratinaud (approved):

PR #862, opened 2022-05-18
Adapt urls to new default branch on contrib versions

martinratinaud:


martinratinaud:
@Ndpnt I believe having `master` still in ops for snapshots is misleading.

I suggest to deploy a snapshot repository ONLY if `snapshot_repository` is set using 

```
- include_tasks: database.yml
  when: snapshot_repository is defined
  vars:
    name: snapshots
    repository: '{{ snapshots_repository }}'
    branch: '{{ snapshots_branch }}'
    directory: '{{ snapshots_directory }}'
  tags:
    - setup
    - update
 ```
 
 And do the same with versions as we could have an only mongo solution.
 
 If you agree I will do this in another PR

Ndpnt (approved):

martinratinaud:
Here it is https://github.com/ambanum/OpenTermsArchive/pull/864

PR #865, opened 2022-05-19
Add custom vagrant config

martinratinaud:
When deploying on vagrant, as no `config_file_name` is specified, the `config/production.json` file is used.

Today `production.json` holds the fact that an email will be sent on error.

When `config_file_name` is set, a symbolic link to `config/local.json` is made. The `local.json` file  overrides all othe configs 
`local.json` > `production.json` > `default.json`

2 options to not receive emails when an error occurs in vagrant:

1. keep no `vagrant.json` custom file but move email config in every instance
2. create a `vagrant.json` file which disables the email sending

This PR implements the latter.

### DIFF
--- a/config/vagrant.json
+++ b/config/vagrant.json
@@ -1,0 +1,5 @@
+{
+  "logger": {
+    "sendMailOnError": false
+  }
+}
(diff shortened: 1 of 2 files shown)


MattiSG (commented):
Thanks @martinratinaud! I'm in favour of this implementation but I remember @ndpnt insisted that Vagrant should mimic production as much as possible because its only use case is to test deployment, not to test features. I'll thus let him approve or request changes on this 🙂

Ndpnt (approved):

PR #866, opened 2022-05-19
Improve documentation

MattiSG:
Fix typos and grammar, unify phrasings, add details, fix some translation hiccups, fix table of contents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,11 +239,10 @@ The default configuration can be found in `config/default.json`. The full refere
 }
 ```
 
-The default configuration is merged with (and overridden by) environment-specific configuration that can be specified at startup with the `NODE_ENV` environment variable.
+The default configuration is merged with (and overridden by) environment-specific configuration that can be specified at startup with the `NODE_ENV` environment variable. For example, you would run `NODE_ENV=development npm start` to load the `development.json` configuration file.
 
-If you want to change your local configuration, we suggest you create a `config/development.json` file with overridden values.
+If you want to change your local configuration, we suggest you create a `config/development.json` file with overridden values. An example of a production configuration file can be found in `config/production.json`.
 
-An example of a production configuration file can be found in `config/production.json`.
 
 ##### Storage adapters configuration
 
--- a/README.md
+++ b/README.md
@@ -351,7 +351,6 @@ To release a dataset:
 npm run dataset:release
 ```
 
-
 To weekly release a dataset:
 
 ```
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-<img src="https://disinfo.quaidorsay.fr/assets/img/logo.png" width="140">
-
 # Open Terms Archive
 
 **Services** have **terms** that can change over time. _Open Terms Archive_ enables users rights advocates, regulatory bodies and any interested citizen to follow the **changes** to these **terms** by being **notified** whenever a new **version** is published, and exploring their entire **history**.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -111,43 +108,24 @@ To find out the address of the RSS feed you want to subscribe to:
...
-- When refering to the base folder, it means the folder where you will be `git pull`ing everything.
+This module is built with [Node](https://nodejs.org/en/) and is tested on macOS, UNIX and Windows. You will need to [install Node >= v16.x](https://nodejs.org/en/download/) to run it.
 
 #### Declarations repository
 
-1. Locally clone your declarations repository, e.g., `git@github.com:OpenTermsArchive/contrib-declarations.git`.
-2. Go into your folder and initialize it, e.g., `cd contrib-declarations; npm install`.
-3. You can now modify your declarations in the `./declarations/` folder, following [these instructions](https://github.com/OpenTermsArchive/contrib-declarations/blob/main/CONTRIBUTING.md).
+1. Locally clone your declarations repository, e.g., `git@github.com:OpenTermsArchive/template-declarations.git`.
+2. Go into your folder and initialize it, e.g., `cd template-declarations; npm install`.
```
martinratinaud, line 127: 1. Create a declarations repository from the [declarations template repository](https://github.com/OpenTermsArchive/template-declarations)


What you describe in the current doc will not work as you need to launch the `init.sh` script

README.md
```diff
@@ -111,43 +108,24 @@ To find out the address of the RSS feed you want to subscribe to:
...
+3. You can now modify your declarations in the `./declarations/` folder, following [these instructions](https://github.com/OpenTermsArchive/template-declarations/blob/main/CONTRIBUTING.md).
```
martinratinaud, line 128: You mentioned it was possible to not have the `blob/main` part of the url but I couldn't find how to.
If it's possible, please remove this part so that it works even if we switch branch (again)

README.md
```diff
@@ -352,13 +332,29 @@ To release a dataset:
...
-
 To weekly release a dataset:
 
 ```
 npm run dataset:scheduler
 ```
 
+## Contributing
+
+Thanks for wanting to contribute! There are different ways to contribute to Open Terms Archive. We describe the most common below. If you want to explore other venues for contributing, please contact us over email (contact@[our domain name]) or [Twitter](https://twitter.com/OpenTerms).
```
martinratinaud, line 343: contact@[our domain name] ?

it it a mistake or a, hidden functionality?

martinratinaud (approved):
I suggest to anyone reviewing this to use the split view :-).



Good job @MattiSG

README.md
MattiSG: Unfortunately, I don't know how to do this either :(

README.md
MattiSG: I just didn't want to give our email to spambots 😉

MattiSG: Do you think it's hard to understand?

martinratinaud: Well, I did not understand it when I reviewed :-)

So I guess `contact [at] opentermsarchive [dot] org` would be more readable (but I do not know how smart spam bots are ;-))

(And if does not work, it's already in the comments anyway 🤣 )

README.md
```diff
@@ -255,17 +237,15 @@ The default configuration can be found in `config/default.json`.
 }
 ```
 
-The default configuration is merged with (and overridden by) environment-specific configuration that can be specified at startup with the [`NODE_ENV` environment variable](#node-env).
-
-If you want to change your local configuration, we suggest you create a `config/development.json` file with overridden values.
+The default configuration is merged with (and overridden by) environment-specific configuration that can be specified at startup with the `NODE_ENV` environment variable. For example, you would run `NODE_ENV=development npm start` to load the `development.json` configuration file.
```
Ndpnt, line 240: This example is not so good. Because if it's defined, `development.json` will already be loaded when running `npm start` even if no `NODE_ENV` are specified.

README.md
```diff
@@ -302,25 +282,25 @@ Two storage adapters are currently supported: Git and MongoDB. Each one can be u
 
 #### Environment variables
 
-Environment variables can be provided in a `.env` file at the root of the repository. See `.env.example` for an example of such a file.
+Environment variables can be passed in the command-line or provided in a `.env` file at the root of the repository. See `.env.example` for an example of such a file.
```
Ndpnt, line 285: Perhaps add a note to explain that if we add `NODE_ENV` to the `.env`, it will not work. It will only work for the variables defined below, not for all environment variables.

PR #869, opened 2022-05-19
Apply docopts and long options policies

### LINKED ISSUE #848: Unify CLI options documentation
Our way of documenting CLI options varies: sometimes using `$` for variables, sometimes `<>`, sometimes using a comma for separating repeating arguments, sometimes not. This leads to recurring discussions between contributors and lack of clarity for users.

Considering [there is no agreed-upon standard](https://stackoverflow.com/questions/9725675/is-there-a-standard-format-for-command-line-shell-help-text):

- Use http://docopt.org syntax.
- Reference http://docopt.org in the CONTRIBUTING.

MattiSG:
Fixes #848.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,8 +65,8 @@ For command-line examples and documentation, we follow the [docopt usage pattern
 - mutually exclusive elements are given between `(` and `)` and separated by `|`.
 
 ```diff
-- npm start $service_id [, $service_id, ...]
-+ npm start <service_id>...
+- npm start [ $service_id ] [, $service_id, ...]
++ npm start [service_id]...
 ```
 
 ##### Long options
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,8 @@ We add this additional rule:
 
 #### CLI
 
+##### docopt
+
 For command-line examples and documentation, we follow the [docopt usage patterns](http://docopt.org) syntax. Quick recap of the main points:
 
 - mandatory arguments are given between `<` and `>`;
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -48,6 +50,35 @@ We add this additional rule:
...
+
+#### CLI
+
+##### docopt
+
+For command-line examples and documentation, we follow the [docopt usage patterns](http://docopt.org) syntax. Quick recap of the main points:
+
+- mandatory arguments are given between `<` and `>`;
+- optional elements are given between `[` and `]`;
+- mutually exclusive elements are given between `(` and `)` and separated by `|`.
```
martinratinaud, line 65: is there also a `default` value we should document?

like `[argument=defaultvalue]`

martinratinaud (approved):

CONTRIBUTING.md
```diff
@@ -48,6 +50,35 @@ We add this additional rule:
...
+
+```diff
+- npm start $service_id [, $service_id, ...]
++ npm start <service_id>...
```
Ndpnt, line 69: `service_id` is optional, shouldn't it be written like that `[service_id]` instead?

CONTRIBUTING.md
MattiSG: This is just a recap, I don't think repeating everything from the full doc is mandatory. We can also just remove these examples if they sound limitative.

CONTRIBUTING.md
MattiSG: I voluntarily gave an example that did not reflect accurately the current CLI but simply gave examples of how to rewrite. I will adjust it to reflect the CLI.

PR #871, opened 2022-05-24
Crash on EAI_AGAIN (DNS lookup failure) errors

### LINKED ISSUE #857: Do not consider DNS lookup errors as FetchDocumentError
A large number of declarations in France failed [because of `EAI_AGAIN`](https://github.com/OpenTermsArchive/france-declarations/issues?q=is%3Aissue+EAI_AGAIN) errors. These seem to be linked to [DNS lookup](https://stackoverflow.com/questions/40182121/whats-the-cause-of-the-error-getaddrinfo-eai-again) failing. These should not be considered as document fetch failure but like network errors instead, crashing the app and generating an email to the administrators.

martinratinaud:
Fixes #857

### DIFF
--- a/src/archivist/fetcher/fullDomFetcher.js
+++ b/src/archivist/fetcher/fullDomFetcher.js
@@ -1,9 +1,10 @@
-import puppeteer from 'puppeteer-extra';
+import puppeteer from 'puppeteer';
+import puppeteerExtra from 'puppeteer-extra';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
 
 import { FetchDocumentError } from './errors.js';
 
-puppeteer.use(stealthPlugin());
+puppeteerExtra.use(stealthPlugin());
 
 let browser;
 
@@ -52,6 +53,9 @@ export default async function fetch(url, cssSelectors, configuration) {
       content: await page.content(),
     };
   } catch (error) {
+    if (error instanceof puppeteer.errors.TimeoutError) {
+      throw new FetchDocumentError(`Timed out after ${configuration.navigationTimeout / 1000} seconds when trying to fetch '${url}'`);
+    }
     throw new FetchDocumentError(error.message);
   } finally {
     if (page) {
@@ -65,7 +69,7 @@ export async function launchHeadlessBrowser() {
     return browser;
   }
 
-  browser = await puppeteer.launch({ headless: true });
+  browser = await puppeteerExtra.launch({ headless: true });
 
   return browser;
 }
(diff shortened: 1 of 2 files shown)


src/archivist/fetcher/fullDomFetcher.js
```diff
@@ -1,9 +1,10 @@
-import puppeteer from 'puppeteer-extra';
+import puppeteer from 'puppeteer';
```
Ndpnt, line 1: Why did you import `puppeteer`?

From the doc of `puppeteer-extra`, it seems not necessary: 
```
// puppeteer-extra is a drop-in replacement for puppeteer,
// it augments the installed puppeteer with plugin functionality.
```

src/archivist/index.js
```diff
@@ -130,7 +130,15 @@ export default class Archivist extends events.EventEmitter {
       }));
     } catch (error) {
       if (error instanceof FetchDocumentError) {
-        throw new InaccessibleContentError(error.message);
+        if (error.message.includes('EAI_AGAIN')) {
+          // DNS error is a network connectivity error or proxy related error. It does
+          // not necesarily mean the document is not accesible to all users but
+          // only to the Open Terms Archive engine at this specific time.
+          // As this is transient most of the time, consider it as a system error
```
Ndpnt, line 137: The problem with the sentence `It does not necesarily mean the document is not accesible to all users but only to the Open Terms Archive engine at this specific time.` is that it's true for some others errors (like 403, timed out, …) So I think we cannot use this as an argument to manage it this way.
I suggest :

```suggestion
          // EAI_AGAIN is a DNS lookup timed out error, which means it is a network connectivity error or proxy related error.
          // This operational error is mostly transient and should be handled by retrying the operation.
          // As there is no retry mechanism in the engine yet, crash the engine and leave it to the process manager to handle the retries and the delay between them.
```

Ndpnt (changes requested):

src/archivist/fetcher/fullDomFetcher.js
martinratinaud: puppeteer-extra does not export `errors` and especially the TimeoutError.
Could not find another way to do so

src/archivist/index.js
martinratinaud: I agree your version is clearer, thanks

src/archivist/fetcher/fullDomFetcher.js
Ndpnt: ok

Ndpnt (approved):

MattiSG (approved):
Really good comment! 👏

martinratinaud:
Review after rebase on `expose_features`

Ndpnt (approved):

Ndpnt:
Fixes #857

PR #872, opened 2022-05-24
Improve fetcher default config

Ndpnt:
It allows to define default configuration values at only one place. It also allows users who import the `fetcher` module and which already use `node-config` to override the default values for the `fetcher` module in their already existing config files.

It also make me think that we should wrap all our configuration inside a `open-terms-archive` key to avoid conflict with such a generic terms as `fetcher`.

### DIFF
--- a/src/archivist/fetcher/index.js
+++ b/src/archivist/fetcher/index.js
@@ -1,23 +1,41 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import config from 'config';
+
 import fetchFullDom from './fullDomFetcher.js';
 import fetchHtmlOnly from './htmlOnlyFetcher.js';
 
 export { launchHeadlessBrowser, stopHeadlessBrowser } from './fullDomFetcher.js';
 export { FetchDocumentError } from './errors.js';
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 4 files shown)


src/archivist/fetcher/index.js
```diff
@@ -1,23 +1,41 @@
...
 import fetchFullDom from './fullDomFetcher.js';
 import fetchHtmlOnly from './htmlOnlyFetcher.js';
 
 export { launchHeadlessBrowser, stopHeadlessBrowser } from './fullDomFetcher.js';
 export { FetchDocumentError } from './errors.js';
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const defaultConfigs = JSON.parse(fs.readFileSync(path.resolve(__dirname, '../../../config/default.json')));
+
+config.util.setModuleDefaults('fetcher', defaultConfigs.fetcher);
```
MattiSG, line 16: > This is great! But maybe a bit heavy loading hidden deep in a module `index`. Shouldn't we set this module default in the main config setup?

MattiSG (approved):
👌

src/archivist/fetcher/index.js
Ndpnt: I don't see how we can do this, because it has to be loaded even when a reuser import only this module.

martinratinaud (approved):
I've added some documentation about how `fetch` can be configured with node config.

PR #873, opened 2022-05-25
Migrate services

Ndpnt:


martinratinaud (changes requested):
Very good job again. I believe the end user will need some documentation to use this.

Still have some parts where I need to pair though.

Ndpnt:
I close this PR as I'm changing the exposed API.

PR #875, opened 2022-06-08
Define an API to query records history

Ndpnt:
- Define and use an API to query records history through `GitRepository` and `MongoRespository`
- Add a migration script as an example of use
- Refactor `storage-adapters` to `repositories` to properly the Repository adn Data Mapper patterns
- Significantly improve performance of git-related queries


scripts/history/utils/index.js
```diff
@@ -0,0 +1,23 @@
+import fsApi from 'fs';
+
+const fs = fsApi.promises;
+
+export async function importReadme({ from: sourceRepository, to: targetRepository }) {
+  const sourceRepositoryReadmePath = `${sourceRepository.path}/README.md`;
+  const targetRepositoryReadmePath = `${targetRepository.path}/README.md`;
+
+  const [readmeCommit] = await sourceRepository.git.log(['README.md']);
```
MattiSG, line 9: This expects that only a single commit contains the README, which does not sound right to me. I see that you then copy the file contents and get just the message and date of the first commit, which could be understood —if I am right— as squashing all README commits to the first one. This is acceptable behaviour to me but would benefit from being made more explicit.
```suggestion
  const [firstReadmeCommit] = await sourceRepository.git.log(['README.md']);
```

scripts/rewrite/initializer/index.js
```diff
@@ -21,7 +21,10 @@ export async function initReadmeAndLicense(targetRepo, targetPath, authorDate) {
   await fs.copyFile(LICENSE_PATH, targetLicenseFilePath);
   await targetRepo.add(targetReadmeFilePath);
   await targetRepo.add(targetLicenseFilePath);
-  await targetRepo.commit(null, 'Add Readme and License', authorDate);
+  await targetRepo.commit({
+    message: 'Add Readme and License',
```
MattiSG, line 25:
```suggestion
    message: 'Add readme and license',
```

src/archivist/index.js
```diff
@@ -41,9 +41,9 @@ export default class Archivist extends events.EventEmitter {
     return Object.keys(this.services);
   }
 
-  constructor({ storage: { versions, snapshots } }) {
+  constructor(config) {
```
MattiSG, line 44: > Throughout the codebase, this is used as the `config` module, making it here ambiguous.
```suggestion
  constructor(properties) {
```

```suggestion
  constructor(props) {
```

```suggestion
  constructor(options) {
```

src/archivist/index.test.js
```diff
@@ -27,14 +26,12 @@ const VERSIONS_PATH = path.resolve(ROOT_PATH, config.get('recorder.versions.stor
...
-let snapshotsStorageAdapter;
-let versionsStorageAdapter;
-
 async function resetGitRepositories() {
-  return Promise.all([ snapshotsStorageAdapter._removeAllRecords(), versionsStorageAdapter._removeAllRecords() ]);
+  return Promise.all([ app.recorder.snapshotsRepository.removeAll(), app.recorder.versionsRepository.removeAll() ]);
 }
 
 let gitVersion;
+let app;
```
MattiSG, line 34: There is a forward usage of `app` that has not been declared yet. Move these lines right after the `const` declarations.

src/archivist/recorder/index.js
```diff
@@ -1,23 +1,22 @@
-export default class Recorder {
-  constructor({ versionsStorageAdapter, snapshotsStorageAdapter }) {
-    if (!versionsStorageAdapter || !snapshotsStorageAdapter) {
-      throw new RangeError('Storage adapters should be defined both for versions and snapshots');
```
MattiSG, line 4: > This validation does not seem to have been reproduced in this class nor in the Factory. It was not tested. I'm okay with losing it and considering behaviour undefined in case storage is not defined for both cases, but this is less defensive as it was.

src/archivist/recorder/record.test.js
```diff
@@ -0,0 +1,43 @@
+import chai from 'chai';
+import config from 'config';
+
+import Record from './record.js';
+import RepositoryFactory from './repositories/factory.js';
+
+const { expect } = chai;
+
+describe('Record', () => {
```
MattiSG, line 9: We are missing a constructor / validator test.

src/archivist/recorder/repositories/git/dataMapper.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const COMMIT_MESSAGE_PREFIXES_REGEXP = new RegExp(`^(${COMMIT_MESSAGE_PREFIX.startTracking}|${COMMIT_MESSAGE_PREFIX.refilter}|${COMMIT_MESSAGE_PREFIX.update})`);
+export default class GitDataMapper {
+  static toPersistence(record, prefixMessageToSnapshotId) {
+    const { serviceId, documentType, isRefilter, snapshotId, mimeType, isFirstRecord } = record;
+
+    let prefix = isRefilter ? COMMIT_MESSAGE_PREFIX.refilter : COMMIT_MESSAGE_PREFIX.update;
+
+    prefix = isFirstRecord ? COMMIT_MESSAGE_PREFIX.startTracking : prefix;
```
MattiSG, line 22:
```suggestion
    let prefix = isFirstRecord ? COMMIT_MESSAGE_PREFIX.startTracking :
    isRefilter ? COMMIT_MESSAGE_PREFIX.refilter :
    COMMIT_MESSAGE_PREFIX.update;
```

src/archivist/recorder/repositories/git/dataMapper.js
```diff
@@ -0,0 +1,62 @@
...
+mime.define({ 'text/markdown': ['md'] }, true); // ensure extension for markdown files is `.md` and not `.markdown`
+
+export const COMMIT_MESSAGE_PREFIX = {
+  startTracking: 'Start tracking',
+  refilter: 'Refilter',
+  update: 'Update',
+};
+
+export const COMMIT_MESSAGE_PREFIXES_REGEXP = new RegExp(`^(${COMMIT_MESSAGE_PREFIX.startTracking}|${COMMIT_MESSAGE_PREFIX.refilter}|${COMMIT_MESSAGE_PREFIX.update})`);
+export default class GitDataMapper {
```
MattiSG, line 16: There are only static members: why is this a class?

src/archivist/recorder/repositories/git/git.js
```diff
@@ -114,6 +104,18 @@ export default class Git {
     return this.git.raw(options);
   }
 
+  async show(options) {
+    return this.git.show(options);
+  }
+
+  async getFullHash(shortHash) {
+    return (await this.git.show([ shortHash, '--pretty=%H', '-s' ])).trim();
```
MattiSG, line 112: I don't understand how that would work in cases where the short hash is ambiguous 🤔

src/archivist/recorder/repositories/git/index.js
```diff
@@ -0,0 +1,203 @@
...
+
+import RepositoryInterface from '../interface.js';
+
+import DataMapper, { COMMIT_MESSAGE_PREFIXES_REGEXP, COMMIT_MESSAGE_PREFIX } from './dataMapper.js';
+import Git from './git.js';
+
+const fs = fsApi.promises;
+const PDF_MIME_TYPE = 'application/pdf';
+
+mime.define({ 'text/markdown': ['md'] }, true); // ensure extension for markdown files is `.md` and not `.markdown`
```
MattiSG, line 19: This is repeated in https://github.com/ambanum/OpenTermsArchive/pull/875/files#diff-6673716d44e6e04b6852f1537f050fa7ebf0bf5c2c7f55d27a6f2de7f627ddadR7.

src/archivist/recorder/repositories/git/index.js
```diff
@@ -0,0 +1,203 @@
...
+
+    return record;
+  }
+
+  finalize() {
+    if (!this.needsPublication) {
+      return;
+    }
+
+    return this.git.pushChanges();
```
MattiSG, line 61: That method is `async`, how can this one not be?

scripts/dataset/export/index.js
```diff
@@ -16,7 +17,7 @@ const fs = fsApi.promises;
 const ARCHIVE_FORMAT = 'zip'; // for supported formats, see https://www.archiverjs.com/docs/archive-formats
 
 export default async function generate({ archivePath, releaseDate }) {
-  const versionsStorageAdapter = await (instantiateVersionsStorageAdapter()).initialize();
+  const versionsRepository = await (RepositoryFactory.create(config.get('recorder.versions.storage'))).initialize();
```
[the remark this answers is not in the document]
Ndpnt: It's an error.

src/archivist/recorder/record.test.js
```diff
@@ -0,0 +1,43 @@
...
+  let repository;
+  let record;
```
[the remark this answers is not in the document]
Ndpnt: I will change it here for consistency, but we should have a discussion about this, I am not as convinced as before.

src/archivist/recorder/repositories/git/dataMapper.js
Ndpnt: Nested ternary expression can be hard to read so we choose to disallow them with https://eslint.org/docs/rules/no-nested-ternary

src/archivist/recorder/repositories/git/git.js
```diff
@@ -32,35 +32,44 @@ export default class Git {
...
-  async commit(filepath, message, authorDate) {
-    const options = {};
+  async commit({ filepath, message, date = new Date() }) {
+    let summary;
 
-    if (authorDate) {
-      options['--date'] = new Date(authorDate).toISOString();
-    }
+    try {
+      const commitDate = new Date(date).toISOString();
```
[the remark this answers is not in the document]
Ndpnt: I not really agree. As the finally is here to rollback env setup even in case of error, I prefer to have the `try` wrapping the content related to the `finally`.

src/archivist/recorder/repositories/git/git.js
Ndpnt: It won't… and it can't. That's why I wanted to switch to full hash.

MattiSG (commented):
Almost done.

src/archivist/recorder/repositories/git/index.js
Ndpnt: Yep I knew, currently both are needed. So what do you suggest?

src/archivist/recorder/repositories/git/index.js
Ndpnt: When marking method `async` it will just wrap it in a promise. In this case I return a promise, so there is no need to wrap it.

src/archivist/recorder/repositories/mongo/index.js
```diff
@@ -0,0 +1,123 @@
...
+import { MongoClient, ObjectId, Binary } from 'mongodb';
+
+import RepositoryInterface from '../interface.js';
+
+import DataMapper from './dataMapper.js';
+
+export default class MongoRepository extends RepositoryInterface {
+  constructor({ database: databaseName, collection: collectionName, connectionURI }) {
+    super();
+    const client = new MongoClient(connectionURI);
```
MattiSG, line 15: Shouldn't the client be a private variable?
```suggestion
    this.#client = new MongoClient(connectionURI);
```

MattiSG (commented):
First pass done.

src/archivist/recorder/repositories/git/git.js
MattiSG: I get that, but the way we get the full hash is by getting the source commit through its short hash, so won't a collision always prevent `getFullHash` from working?

scripts/dataset/export/index.js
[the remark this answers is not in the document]
MattiSG: > What do you suggest instead?

Shouldn't `create` always call `initialise`? 🤔

MattiSG (dismissed):
👍

scripts/history/migrate-services.js
```diff
@@ -0,0 +1,212 @@
...
+  console.timeEnd('Total time');
+
+  await finalize(migration);
+}());
+
+async function rewriteSnapshots(repository, records, idsMapping, logger) {
+  let i = 1;
+
+  for (const record of records) {
+    const { id: recordId } = await repository.record(record); // eslint-disable-line no-await-in-loop
```
Ndpnt, line 143: Use .save instead of .record

scripts/history/logger/index.js
```diff
@@ -0,0 +1,39 @@
...
+      if (type.length > 50) {
+        type = `${type.substring(0, 48)}…`;
+      }
+
+      prefix += `${type}`.padEnd(50);
+    }
+
+    if (id) {
+      prefix += `${id}`.padEnd(42);
+    }
```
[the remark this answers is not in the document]
Ndpnt: Yes but it's not very important. This logger is only used in history scripts and either we have all the parameters or we only have the message.

scripts/history/migrate-services.js
```diff
@@ -0,0 +1,212 @@
...
+    prefixMessageToSnapshotId: 'This version was recorded after filtering snapshot https://github.com/OpenTermsArchive/france-snapshots/commit/',
+  },
+  to: {
+    snapshots: 'france-elections-snapshots',
+    versions: 'france-elections-versions-hash-updated',
+    prefixMessageToSnapshotId: 'This version was recorded after filtering snapshot https://github.com/OpenTermsArchive/france-elections-snapshots/commit/',
+  },
+};
+
+const COUNTERS = {
```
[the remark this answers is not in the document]
MattiSG: Yes, I agree with that understanding. I'm not saying it's a problem, I just find it weird: the hint given to my mental model, both with `const` and all caps, is that the intent of this variable is to represent a constant, and I find it hard to think of a constant as just a reference and then all its members are intended to be mutable.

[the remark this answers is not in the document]
Ndpnt: I understand and I agree. I think we should keep all caps to effectively constant values, I will rename `COUNTERS` to `counters`.

scripts/history/migrate-services.js
```diff
@@ -0,0 +1,212 @@
+import fsApi from 'fs';
```
martinratinaud, line 1: This implementation seems very clear to me but have the downside that every migration created will have commitable changes.

I think this may not be part of this PR but it would be great to have the config loaded from a gitignored file `.migration.json`
Only change needed would be to add the type of repository to it so that the right repository class is used

src/archivist/recorder/repositories/git/git.js
MattiSG: We agreed on a synchronous call to have a better implementation in a separate PR.

src/archivist/recorder/index.test.js
```diff
@@ -51,20 +23,22 @@ describe('Recorder', () => {
         let record;
 
         before(async () => {
-          recorder = new Recorder({
-            versionsStorageAdapter: versionsAdapter,
-            snapshotsStorageAdapter: snapshotsAdapter,
-          });
+          const options = config.util.cloneDeep(config.recorder);
```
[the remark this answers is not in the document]
Ndpnt: Because the object returned by `node-config` is immutable and I need to modify it.

src/archivist/recorder/repositories/git/git.js
```diff
@@ -32,35 +32,44 @@ export default class Git {
...
 
-    let summary;
+      process.env.GIT_AUTHOR_DATE = commitDate;
```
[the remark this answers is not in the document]
Ndpnt: Because it is the only way to set the git committer date.

src/archivist/recorder/repositories/interface.js
```diff
@@ -0,0 +1,102 @@
...
+  * @param {string} recordId - Record ID of the record to find
+  * @returns {Promise<Record>} Promise that will be resolved with the found record or an empty object if none match the given ID
+  */
+  async findById(recordId) {
+    throw new Error(`#findById method is not yet implemented in ${this.constructor.name}`);
+  }
+
+  /**
+  * Find all records
+  * For performance reasons, the content of the records will not be loaded by default. Use #loadRecordContent to load the content of individual records
```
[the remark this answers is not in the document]
Ndpnt: I think the previous wording is clearer as the caller will have to load the content of each record found

src/archivist/recorder/repositories/mongo/index.js
```diff
@@ -0,0 +1,123 @@
...
+      return record;
+    }
+
+    await this.loadRecordContent(record);
+
+    return record;
+  }
+
+  async #toPersistence(record) {
+    if (record.content === undefined || record.content === null) {
```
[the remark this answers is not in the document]
Ndpnt: Content could be defined manually so it could be undefined or null

src/archivist/index.js
```diff
@@ -41,9 +41,9 @@ export default class Archivist extends events.EventEmitter {
     return Object.keys(this.services);
   }
 
-  constructor(config) {
+  constructor({ recorder }) {
```
MattiSG, line 44: > I find this less clear than knowing the parameter is called `config`. I expect an instance of `Recorder` in the `recorder` field, not a config object for a new Recorder.

src/archivist/recorder/index.test.js
```diff
@@ -32,10 +32,7 @@ describe('Recorder', () => {
           await recorder.initialize();
         });
 
-        after(async () => {
-          await recorder.snapshotsRepository.removeAll();
-          await recorder.finalize();
-        });
+        after(async () => recorder.finalize());
```
MattiSG, line 35:
```suggestion
        after(recorder.finalize.bind(recorder));
```

MattiSG (commented):
I'm impressed that we can drop `src/index.js` altogether. With these incremental reviews, I can't really be sure how appropriate it is.

src/archivist/recorder/repositories/mongo/dataMapper.js
```diff
@@ -2,45 +2,30 @@ import { ObjectId } from 'mongodb';
...
-    documentFields.created_at = new Date();
-
-    return documentFields;
+  if (documentFields.snapshotId) {
+    documentFields.snapshotId = new ObjectId(record.snapshotId);
   }
 
-  static toDomain(document) {
-    const { _id, serviceId, documentType, fetchDate, mimeType, isRefilter, isFirstRecord, snapshotId } = document;
+  documentFields.content = record.content;
```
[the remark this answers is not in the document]
Ndpnt: No because `content` is defined with a getter and it's not enumerable, so it is not defined in `documentFields`

PR #883, opened 2022-06-22
Add timeout to cleaning step in case of fatal errors

Ndpnt:


martinratinaud (approved):
Very nice!

Good job as it has been a real pain.

Let's see how it behaves next time it crashes

PR #885, opened 2022-06-28
Add support for Ubuntu in ops

Ndpnt:


ops/inventories/production.yml
```diff
@@ -9,41 +9,28 @@ all:
     app_directory: ota
     ota_repository: https://github.com/ambanum/OpenTermsArchive.git
     ota_branch: main
+    app_config: "{{ lookup('file','../config/{{config_file_name}}.json') | from_json }}"
```
martinratinaud, line 12: Very nice!

ops/roles/infra/tasks/main.yml
```diff
@@ -45,17 +49,28 @@
...
   # See https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#recommended-enable-user-namespace-cloning
 - name: Enable user namespace cloning to allow running Chromium in a sandbox
   command: sysctl -w kernel.unprivileged_userns_clone=1
   when: ansible_facts['architecture'] != "aarch64"
 
 - include_tasks: mongo.yml
-  when: ansible_facts['architecture'] != "aarch64"
+  when:
+    - mongo_required
+    - ansible_facts['architecture'] != "aarch64"
```
martinratinaud, line 76: Can you please add a comment on why it should not run on aarch64.
Also, if this is a mongo version problem, can't we use mongo 4 on ubuntu?

martinratinaud (changes requested):

ops/roles/infra/tasks/main.yml
Ndpnt: Not really, I do not know exactly why we cannot use mongo on aarch64. By the way, it was you that added this condition, so If you remember why, please add the comment yourself :)

martinratinaud: In order to test, uncomment the lines and try to deploy on vagrant: `ansible-playbook ops/site.yml -i ops/inventories/dev.yml -vvv`

PR #886, opened 2022-06-28
Clear git repository on initialization

Ndpnt:
Prevent uncommitted changes added to the staged area to be committed

Should avoid this kind of issue: https://github.com/OpenTermsArchive/contrib-versions/commit/e79cbdb65f98fab9180a3fe4fea150cfd2ee71b7


martinratinaud (approved):
👍

Also,
> Should avoid this kind of issue: [OpenTermsArchive/contrib-versions@e79cbdb](https://github.com/OpenTermsArchive/contrib-versions/commit/e79cbdb65f98fab9180a3fe4fea150cfd2ee71b7)

Just to make it clear, it will avoid having 2 files commited at once in the same commit.

Ndpnt:
> Just to make it clear, it will avoid having 2 files commited at once in the same commit.

Not exactly. It will avoid to have leftovers files in the staging area that may be committed for another service.

PR #887, opened 2022-06-29
Ensure only specified file is committed in git repository

Ndpnt:


src/archivist/recorder/repositories/git/git.js
```diff
@@ -120,6 +119,12 @@ export default class Git {
     return this.git.raw([ 'restore', '-s', commit, '--', path ]);
   }
 
+  async cleanUp() {
```
martinratinaud, line 122: it now seems weird to me that

`cleanUp` function is using `this.git.clear`
`clear` function is using `this.git.clean`

Maybe we should harmonize this

martinratinaud (changes requested):

src/archivist/recorder/repositories/git/git.js
```diff
@@ -120,8 +110,13 @@ export default class Git {
     return this.git.raw([ 'restore', '-s', commit, '--', path ]);
   }
 
+  async destroyHistory() {
```
martinratinaud, line 113: 👍

martinratinaud (approved):

PR #895, opened 2022-07-21
Add command to track documents from declarations folder

martinratinaud:
Up until now, if the user wants to generate the snapshots and versions, he is obliged to launch `npm start Service` from the OpenTermsArchive project and setup a `config/development.json` file pointing to the declarations folder he wants to use. Something like:

```json
{
  "services": {
    "declarationsPath": "../sandbox-declarations/declarations"
  },
}
```
This would result in files created in `OpenTermsArchive/data` folder.

While this works well, it obliges user to clone both OpenTermsArchive and declaration repo.

In most of use cases, a user will work on one instance, checkout the corresponding declarations folder and work from there.

This PR aims at providing a new command `ota` that will be exposed as a `bin` command and that will generate the snapshots and versions directly in the folder it is called from. 

Check also the [corresponding PR](https://github.com/OpenTermsArchive/sandbox-declarations/pull/116) in declaration repository


src/start.js
```diff
@@ -0,0 +1,61 @@
+import config from 'config';
+import cron from 'croner';
+
+import Archivist from './archivist/index.js';
+import logger from './logger/index.js';
+import Notifier from './notifier/index.js';
+import Tracker from './tracker/index.js';
+
+async function startOpenTermsArchive({ services, refilterOnly, schedule }) {
```
MattiSG, line 9: In the context of the Open Terms Archive project, I am not in favour of adding the project name to that function name.

Since this is the exportable main loop, convention is to call this file `index.js`, and to export the default as the main function, which I suggest we call `track` 🙂

src/main.js
```diff
@@ -1,64 +1,8 @@
-import config from 'config';
-import cron from 'croner';
-
-import Archivist from './archivist/index.js';
-import logger from './logger/index.js';
-import Notifier from './notifier/index.js';
-import Tracker from './tracker/index.js';
+import startOpenTermsArchive from './start.js';
```
MattiSG, line 1:
```suggestion
import track from './index.js';
```

package.json
```diff
@@ -19,7 +19,8 @@
   },
   "bin": {
     "ota-lint-declarations": "./bin/lint-declarations.js",
-    "ota-validate-declarations": "./bin/validate-declarations.js"
+    "ota-validate-declarations": "./bin/validate-declarations.js",
+    "ota": "./bin/ota.js"
```
MattiSG, line 23: I foresee that we'll want to expose options rather than several executables… maybe it would be more future-proof to not take the `ota` name in those conditions, and instead keep the same convention as the other names?
```suggestion
    "ota-track": "./bin/track.js"
```

README.md
```diff
@@ -128,6 +128,7 @@ The following commands are available where the package is installed:
 
 - `./node_modules/.bin/ota-lint-declarations [service_id]...` to lint declarations
 - `./node_modules/.bin/ota-validate-declarations [service_id]...` to test declarations
+- `./node_modules/.bin/ota [service_id]...` to test services and generate artifacts
```
MattiSG, line 131: “test services and generate artifacts” is not clear to me. Do we mean “to track all declared documents and store their snapshots and versions in the `./data` folder”? We probably want to mention that Git is mandatory in this case, and add guidance for when the user might prefer to run from source rather than using the CLI (i.e. for config), and information on where the declarations will be read from 🙂

MattiSG (changes requested):
Nice! I really like the refactor of the tracker.

I find the suggested command appealing, but it's not entirely clear to me if now the main usage we recommend is to use the CLI or not. This should be disambiguated in the doc 🙂

martinratinaud:
@Ndpnt can you please review before I make changes so that I come back to this only once. Thanks

README.md
Ndpnt: It's not only to test services, some users may want to use it to really track services. So I suggest:
```
The following commands are available where the package is installed:

- `./node_modules/.bin/ota-lint-declarations [service_id]`: Lint declarations by auto formatting content.
- `./node_modules/.bin/ota-validate-declarations [service_id]` Validate declarations.
- `./node_modules/.bin/ota-track [service_id]` Track services. Recorded snapshots and versions will be store in the `./data` folder at the root of the project where the package is installed.
```

Ndpnt:
Avoid commit message mentioning contributor's review message and explain the intention instead: 
`Add Nico's suggestion on doc` -> `Improve documentation`

martinratinaud:
I plan to merge this tomorrow.

README.md
```diff
@@ -126,8 +126,9 @@ npm install "ambanum/OpenTermsArchive#main"
 
 The following commands are available where the package is installed:
 
-- `./node_modules/.bin/ota-lint-declarations [service_id]...` to lint declarations
-- `./node_modules/.bin/ota-validate-declarations [service_id]...` to test declarations
+- `./node_modules/.bin/ota-lint-declarations [service_id]...` Lint declarations by auto formatting content.
```
MattiSG, line 129:
```suggestion
- `./node_modules/.bin/ota-lint-declarations [service_id]...`: check and normalise the format of declarations.
```

README.md
```diff
@@ -126,8 +126,9 @@ npm install "ambanum/OpenTermsArchive#main"
...
+- `./node_modules/.bin/ota-validate-declarations [service_id]...` Validate declarations.
```
MattiSG, line 130:
```suggestion
- `./node_modules/.bin/ota-validate-declarations [service_id]...`: validate declarations.
```

README.md
```diff
@@ -126,8 +126,9 @@ npm install "ambanum/OpenTermsArchive#main"
...
+- `./node_modules/.bin/ota-track [service_id]...` Track services. Recorded snapshots and versions will be store in the `./data` folder at the root of the project where the package is installed.
```
MattiSG, line 131:
```suggestion
- `./node_modules/.bin/ota-track [service_id]...`: track services. Recorded snapshots and versions will be stored in the `data` folder at the root of the module where the package is installed.
```

bin/track.js
```diff
@@ -0,0 +1,32 @@
...
+  versions: { storage: { git: { path: path.resolve(process.cwd(), './data/versions') } } },
+  snapshots: { storage: { git: { path: path.resolve(process.cwd(), './data/snapshots') } } },
+}));
+config.util.setModuleDefaults('logger', defaultConfigs.logger);
+// we do not want any tracker when launching through this command line
+config.util.setModuleDefaults('tracker', {});
+
+const servicesToValidate = process.argv.slice(2);
+
+(async () => {
```
MattiSG, line 28: > Top-level await is available [since Node v14.8](https://nodejs.org/en/blog/release/v14.8.0/) and we already require Node `>= 16`, so I don't think we need this wrapper 🙂

MattiSG (changes requested):

MattiSG (approved):
Nice!

PR #898, opened 2022-08-10
Suggest a more universal install command

guillett:
The `#` has a [specific meaning](https://stackoverflow.com/questions/12303805/oh-my-zsh-hash-pound-symbol-bad-pattern-or-match-not-found#13783754) with `zsh`.

The `"` add some complexity and allow the command to be directly used in more contexts.


martinratinaud (approved):
Thanks a lot @guillett

PR #899, opened 2022-08-13
Get target server fingerprints from inventory instead of from secret

MattiSG:
**This changeset [brings CD back](https://github.com/ambanum/OpenTermsArchive/actions/runs/2848281295) and removes one secret to maintain.**

## Problem overview

CD fails since #894, because deployment relies on a secret called `SERVER_FINGERPRINT` to fill in the SSH `known_hosts` we use to connect to the production servers.

This variable is misnamed, since its singular implies that there is only one fingerprint. This is not accurate: each server added to the production inventory needs to have its fingerprint added to the `known_hosts` for CD to accept connecting to it. The reason for the failure was that this new fingerprint was not added in the secrets.

In GitHub Actions, secrets are write-only, making it cumbersome to append content to them, and impossible to diagnose such an error.

## Solutions considered

I first considered [disabling](https://github.com/shimataro/ssh-key-action#i-want-to-omit-known_hosts) [host key checking](https://docs.ansible.com/ansible/latest/user_guide/connection_details.html#managing-host-key-checking) altogether, since we only have public data (declarations are pulled, and snapshots and declarations are pushed) on the servers. However, we do have one private piece of information we would not want to give to a compromised server: the bot private key, which has write access to repositories. It is mandatory that this information stays secret, as it otherwise enables an attacker to tamper with the data while pretending to be a legitimate serve. In the current deployment setup, host authentification is useful.

I then considered writing a `known_hosts` as a cleartext file maintained as an Ansible asset and applying at once. This implementation can be seen in fa86ac48443ccb150157191bd02e8b27f97b1912. There is indeed [no security risk](https://serverfault.com/a/705073/406386) in disclosing this file, and we thus make read possible, write version-controlled, and improve discovery. However, I felt that discoverability was low and maintenance costs high: with no direct mapping between instance name and host fingerprint, we were extremely likely to systematically forget updating that file when adding a new instance.

## Solution implemented

This changeset switches from getting known hosts from a GitHub Actions secret to entries in the Ansible inventory list. There is [no security risk](https://serverfault.com/a/705073/406386) in disclosing this information, and we thus make read possible, write version-controlled, and improve discovery and maintenance.

### Additional benefits

Deployment is now handled through a matrix, so failures and successes can be assessed by instance.

Screen Shot 2022-08-12 at 19 21 58

### Using a third-party action to read the contents of the file

I tried to implement it [manually](https://stackoverflow.com/questions/58913512/how-to-give-github-action-the-content-of-a-file-as-input) but newlines need to be escaped, making the steps pretty heavy.
I first went with a pure YAML reader, but it did not support dynamic path construction (see 5bb96f7d69f7a4d801f569b1e4db499a5d22946b), so I ended up loading `jq`.

### Storing only the ED-25519 fingerprint

Only one cypher fingerprint is needed, and ED-25519 is both the shortest and the [most secure](https://twitter.com/matti_sg/status/788673647720423424) available.


martinratinaud (approved):
Good job, it seems there has been a lot of trial and errors (I feel you as I had some of them when working on a way to have [instances dynamically loaded from the yaml](https://github.com/ambanum/OpenTermsArchive/pull/901)

Feels a lot easier to add a new instance

MattiSG:
> it seems there has been a lot of trial and errors

I cleaned the history, there was a lot more, I just kept trace of the most tempting optimisations that won't work :wink:

MattiSG:
Just waiting for #901 to be merged (or not) before merging this one 🙂

MattiSG:
Tests failed on Windows because of reasons detailed in #904. Hopefully, a rebase after #904 should enable merging this PR.

PR #901, opened 2022-08-17
Retrieve instance names directly from YAML

martinratinaud:
As it would be a hassle to update this list everytime, I used [this article](https://www.cynkra.com/blog/2020-12-23-dynamic-gha/) to generate this list automatically from reading the yaml file

I also [tested the whole process](https://github.com/ambanum/OpenTermsArchive/actions/runs/2875351373) (without the last step of actual deployment) before making the PR

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,20 +7,27 @@ on:
 jobs:
   test:
     uses: "ambanum/OpenTermsArchive/.github/workflows/test.yml@main"
+
+  get_instances:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set matrix with instance names
+        id: get_instance_names
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq ".all.children.production.children | keys" ops/inventories/production.yml -o=json | tr '\n' ' '
+      - id: set-matrix
+        run: echo "::set-output name=matrix::{\"instance\":${{ toJson(steps.get_instance_names.outputs.result) }}}"
+
   deploy:
-    needs: [ test ]
+    needs: [ test, get_instances ]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false  # deploy other instances even if one fails to deploy
-      matrix:
-        instance:
-          - contrib
-          - dating
-          - france
-          - france_elections
-          - pga
-          - p2b_compliance
-
+      matrix: ${{ fromJSON(needs.get_instances.outputs.matrix) }}
     steps:
       - uses: actions/checkout@v2
       - name: Get target server hostname
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ jobs:
         id: get_production_instances_names
         uses: mikefarah/yq@master
         with:
-          cmd: yq ".all.children.production.children | keys" ops/inventories/production.yml -o=json
+          cmd: yq ".all.children.production.children | keys" ops/inventories/production.yml --output-format=json
 
   deploy:
     needs: [ test, get_instances ]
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -7,20 +7,27 @@ on:
...
   test:
     uses: "ambanum/OpenTermsArchive/.github/workflows/test.yml@main"
+
+  get_instances:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set matrix with instance names
```
MattiSG, line 17: This step does not actually set any matrix :wink:

```suggestion
      - name: Get all declared instances
```

.github/workflows/deploy.yml
```diff
@@ -7,20 +7,27 @@ on:
...
+        id: get_instance_names
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq ".all.children.production.children | keys" ops/inventories/production.yml -o=json | tr '\n' ' '
+      - id: set-matrix
```
MattiSG, line 22: Use consistent formatting; in this case, snake case.

Also, this step does not actually set any matrix.

```suggestion
      - id: get_instances_matrix_list
```

MattiSG, line 22: Use consistent formatting; in this case, snake case.

Also, this step does not actually set any matrix.

```suggestion
      - id: get_instances_matrix_list
```

.github/workflows/deploy.yml
```diff
@@ -7,20 +7,27 @@ on:
...
-      matrix:
-        instance:
-          - contrib
-          - dating
-          - france
-          - france_elections
-          - pga
-          - p2b_compliance
-
+      matrix: ${{ fromJSON(needs.get_instances.outputs.matrix) }}
```
MattiSG, line 30: Do we really need the additional wrapping? Why not:

```suggestion
      matrix:
        instance: ${{ fromJSON(needs.get_instances.outputs.instances) }}

MattiSG (changes requested):
Thanks for having implemented that! I also saw that possibility but decided against it because I had already invested significant time in making #899, and because I wondered if it did not make sense to have an instance defined without necessarily deploying it continuously. However, since we fetch the list of instances from the `production` inventory, I guess we can agree that servers listed there should be in CD 👍

.github/workflows/deploy.yml
martinratinaud: I did not manage to make it work like that, and I made many attempts. But maybe I missed something.

Anyway, I'd rather leave it like this now as it works

MattiSG:
@martinratinaud I am not sure how to interpret the fact that you marked the other conversations as “resolved” with no comment or change, I would be happy to get a clarification 🙂

.github/workflows/deploy.yml
MattiSG: https://github.com/ambanum/OpenTermsArchive/pull/901/commits/a7924981b23937eee0d6dc6febfb476f3a4b690f [seems to work](https://github.com/ambanum/OpenTermsArchive/actions/runs/2875810065) 🙂

MattiSG (approved):
I allowed myself to push the refactor straight to this branch after [a test](https://github.com/ambanum/OpenTermsArchive/actions/runs/2875810065) showed it worked, considering that your comment showed you agreed with the intention 🙂

I'll let you confirm that interpretation by doing the merge yourself. Feel free to revert if you wanted something else!

martinratinaud:
> @martinratinaud I am not sure how to interpret the fact that you marked the other conversations as “resolved” with no comment or change, I would be happy to get a clarification 🙂

My bad, I forgot to push the changes and I see you made yourself in the meantime.

.github/workflows/deploy.yml
martinratinaud: Well thanks a lot, it's a lot clearer and I though that was one of the first thing I tried (but surely wrongly)

martinratinaud:
So I let you merge that when you want then

PR #905, opened 2022-08-23
Improve CI

MattiSG:
Following exploration of the problem (hopefully) fixed by #904, I found several inconsistencies or venues for improvement in our current CI setup. This changeset collects them and:

- Adds logs for MongoDB server on CI.
- Adds macOS testing in CI.
- Enables testing MongoDB on IPv6-enabled hosts. On my machine, for example, my hosts file maps `localhost` to the IPv6 `::1` loopback interface instead of only the IPv4, and that seems to prevent the Mongo client from connecting to the database.
- Actually tests all `Recorder` operations on MongoDB. The initialisation step was not factored properly and we ended up testing version recording and refiltering twice with Git, not with MongoDB.
- Updates the `mongodb` driver version to bring compatibility with MongoDB v6, which came out late July.
- Removes obsolete `git config --global` for user name and email in CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,8 +19,6 @@ jobs:
 
     steps:
       - run: |
-          git config --global user.name 'Open Terms Archive Bot'
-          git config --global user.email 'bot@opentermsarchive.org'
           git config --global core.autocrlf false
       - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


martinratinaud (approved):
Thanks a lot for all the fixing and good catch on the missing tests

PR #907, opened 2022-08-24
Fix range selectors puppeteer

### LINKED ISSUE #906: Snapshot always fail when `executeClientScripts` and `startBefore`
When creating a snapshot for a document using puppeteer, we wait for the needed selectors to be present on the page before recording the snapshot.

When using an object instead of a string as a selector like

```
 {
          "startBefore": "[role=\"separator\"]",
          "endAfter": "body"
        }
```

The snapshot will not be retrieved and will fail with an error as code is currently not supporting this possibility

martinratinaud:
This closes #906

I decided to keep the logic of css selectors flattening in the fetcher itself as if not, the contribution tool would also need to be updated.
Also I think we must have an open discussion about the coherence between a DocumentDeclaration and the representation of this declaration in the JSON file. 
From what I see, we could benefit a lot about making them converge.

This PR is blocking for adding Airbnb documents on P2B.

### DIFF
--- a/src/archivist/fetcher/index.js
+++ b/src/archivist/fetcher/index.js
@@ -1,5 +1,7 @@
 import config from 'config';
 
+import DocumentDeclaration from '../services/documentDeclaration.js';
+
 import fetchFullDom from './fullDomFetcher.js';
 import fetchHtmlOnly from './htmlOnlyFetcher.js';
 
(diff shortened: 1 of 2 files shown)


martinratinaud:
As fetcher should only have a list of CSS selectors as an array of string, this PR is obsolete and thus replaced by https://github.com/ambanum/OpenTermsArchive/pull/912

PR #908, opened 2022-08-24
Unify configuration filename and instance name

MattiSG:
This changeset replaces declaring the name of the configuration file with getting the name of the instance straight from the inventory. It also simplifies the inventory layout by removing one level of nesting.

### Consequences

- I am not sure if hostnames on production instances should be updated.
- For consistency, repository names containing dashes (currently `p2b-compliance` and `france-elections`) should be renamed with underscores.

### Implementation choices

A check for the existence of the configuration file has been removed. This seems appropriate since an instance with no config should not exist, and this results in an Ansible error complaining that the target file does not exist and that a `force` option should be set if the link should really be created under these circumstances.


martinratinaud (approved):
This looks very good to me and easier to maintain. Thanks

martinratinaud:
I think hostname do not have to be changed as we are now creating a local.json file within the directory that overrides all other config.
It was neede before as we used a behaviour (which I found too obscure) where, if a config file exists with the same name of the machine, then it takes predescendance over all other files.

MattiSG (commented):
After discussion with @Ndpnt: overriding the config file name is important for multi-tenancy. This is not currently used but has been has some point, and removing support in this PR is not necessary. We will implement this by defaulting to the instance name, but supporting override.

MattiSG:
- File name override added.
- As seen with @Ndpnt, postponed migrating `p2b-compliance` and `france-elections` names by taking advantage of that feature.
- Further simplified inventory layout.
- Fixed some typos and whitespace issues along the way.
- Added support for `declarations_branch` variable, to unify with other database types.
- Documented defaults.
- Rebased over #911.

MattiSG:
@Ndpnt I can't seem to re-request review through the GitHub interface, asking through this comment 😉

Ndpnt (approved):

PR #909, opened 2022-08-24
Use third-party action to set up MongoDB in CI

MattiSG:
This changeset simplifies cross-platform setup, at the expenses of adding a dependency. I am not certain this is worth it, but wanted to consolidate some of my discoveries from #904.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,26 +21,14 @@ jobs:
       - run: |
           git config --global core.autocrlf false
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


martinratinaud (approved):
Although I don't mind having a dependency for that kind of need, [this one](https://github.com/ankane/setup-mongodb) seems to not have that many stars and only one maintainer, so I guess it's the risk.

Checking further down this repo, it seems this user is very proficient in writing github actions as he's done the same for all databases :-)

So I agree to take the risk

Ndpnt (approved):

PR #910, opened 2022-08-24
Use third-party action to deploy with Ansible

MattiSG:
This changeset speeds up deployment by ~1 minute by avoiding installing Ansible on the runner by using Docker to avoid reinstalling every time, greatly simplifies the CI workflow declaration, and simplifies the inventory maintenance by removing entirely the need to update hosts fingerprints (yes, that makes most of #899 obsolete :wink:).

This [has been tested](https://github.com/ambanum/OpenTermsArchive/actions/runs/2918678773) and proven to work.

The only drawback from this solution is that we depend on a third party to update Ansible, and this is unlikely to happen (cf. https://github.com/saubermacherag/ansible-playbook-docker-action/issues/11). We also open up more of a security risk. We could mitigate both issues by forking and maintaining that action ourselves. I am not sure which of the three options (status quo, third-party, fork) is preferable.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,22 +29,12 @@ jobs:
         instance: ${{ fromJSON(needs.get_instances.outputs.names) }}
     steps:
       - uses: actions/checkout@v2
-      - name: Get target server hostname
-        id: get_hostname
-        uses: mikefarah/yq@master
-        with:
-          cmd: yq ".all.children.production.children.${{ matrix.instance }}.hosts.* | key" ops/inventories/production.yml
-      - name: Get target server fingerprint
-        id: get_fingerprint
-        uses: mikefarah/yq@master
-        with:
-          cmd: yq '.all.children.production.children.${{ matrix.instance }}.hosts.["${{ steps.get_hostname.outputs.result }}"].ed25519_fingerprint' ops/inventories/production.yml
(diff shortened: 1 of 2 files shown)


martinratinaud (approved):
Neater, there will soon be nothing to update in this file when deploying a new version 🤣.

Although I'd like to see how it looks like after merge of https://github.com/ambanum/OpenTermsArchive/pull/908

MattiSG:
As discussed with @Ndpnt, this third-party action does not seem to be actively maintained. This is not a major problem at this stage.

However, what is a blocking problem and increases significantly the workload in forking and maintaining it ourself, is that the action itself depends on a [prebuilt Docker image](https://github.com/saubermacherag/docker-ansible-alpine/blob/master/Dockerfile) by @saubermacherag that is not versioned and that [disables host key checking](https://github.com/saubermacherag/docker-ansible-alpine/blob/master/Dockerfile#L43). This opens an immediate security risk for us as we upload a private key to our target servers. Without host key checking, we cannot authenticate recipients. Moreover, the reliance on a third-party Docker image prevents us from having sovereignty over the deployment process we hand our vault key to.

The CD speedup could probably be obtained with proper caching of `ansible`, and is not currently a major problem.

The conclusion is thus to drop this changeset, for its added value is too low and its added risk is too high.

Issue #897, opened 2022-08-09
`select` acts different  with array and comma


martinratinaud:
While adding the Oculus Commercial terms I stumbled on the following problem when creating the version

Using `h1, h1 ~ *` (See on [contribution tool](https://contribute.opentermsarchive.org/en/service?destination=OpenTermsArchive%2Fp2b-compliance-declarations&documentType=Commercial%20Terms&executeClientScripts=true&expertMode=true&hiddenCss[]=%5Bdata-testid%3D%22cookie-policy-dialog%22%5D&name=Oculus&selectedCss[]=h1%2C%20h1%20~%20%2A&url=https%3A%2F%2Fdeveloper.oculus.com%2Fpolicy%2Fp2b%2F))

gives a different version than using `['h1', 'h1 ~ *']` (See on [contribution tool](https://contribute.opentermsarchive.org/en/service?destination=OpenTermsArchive%2Fp2b-compliance-declarations&documentType=Commercial%20Terms&executeClientScripts=true&expertMode=true&hiddenCss[]=%5Bdata-testid%3D%22cookie-policy-dialog%22%5D&name=Oculus&selectedCss[]=h1&selectedCss[]=h1%20~%20%2A&url=https%3A%2F%2Fdeveloper.oculus.com%2Fpolicy%2Fp2b%2F))

Click on `Verify version` and you'll see on the second link it gets only the `h1` and not the siblings

PR #913, opened 2022-08-29
Ensure to not remove nodes when selecting them

Ndpnt:
Fixes https://github.com/ambanum/OpenTermsArchive/issues/897


martinratinaud (approved):
Very good, have you understood why the only fact to use it with appendChild removed it from the whole document?

Ndpnt:
> Very good, have you understood why the only fact to use it with appendChild removed it from the whole document?

`appendChild` moves the original node. So, as we move nodes from one fragment to another, it is removed from the first one.

PR #919, opened 2022-08-31
Fix continuous deployment

MattiSG:
Following #908 improvements after initial review, the continuous deployment workflow was not tested and failed. The root cause was the YAML selectors that were not up to date. Once fixed, another issue was identified: the `site` playbook did not benefit from the default `ansible_user` that was set only for the `ota` role, preventing SSH connection when played.


MattiSG:
This has been tested to work: https://github.com/ambanum/OpenTermsArchive/actions/runs/2963539659

Ndpnt (approved):

PR #920, opened 2022-08-31
Revert deletion of mandatory production file

Ndpnt:
Config production file allows factorizing some configuration keys which are the same for all instances when ran in the production environment.

Deleted in [cab3ea2ff47bb115e64c21528d60faf3935d054b](https://github.com/ambanum/OpenTermsArchive/pull/908/commits/cab3ea2ff47bb115e64c21528d60faf3935d054b).

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -1,0 +1,27 @@
+{
+  "recorder": {
+    "versions": {
+      "storage": {
+        "git": {
+          "path": "../versions",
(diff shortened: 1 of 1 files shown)


martinratinaud (approved):
This also fixes the fact that all commits on all instances were done in `data/versions` instead of `../versions` and same for snapshots.

PR #927, opened 2022-09-04
Add Ranking Parameters Description document type

MattiSG:
Follows https://github.com/ambanum/OpenTermsArchive/discussions/902.

This changeset also introduces the `references` key in the document types, to give more context and legal background to a document type.

### DIFF
--- a/src/archivist/services/documentTypes.json
+++ b/src/archivist/services/documentTypes.json
@@ -333,6 +333,11 @@
       "writer": "search engine or intermediation service provider",
       "audience": "users presented with a list of ranked results, or contributors of results that are ranked",
       "object": "parameters determining ranking and the reasons for their relative importance"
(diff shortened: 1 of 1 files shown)


martinratinaud (approved):

martinratinaud:
Please do not forget to update the excel file after merge.
Also, note that contribution tool is relying on the [api url](https://api.opentermsarchive.org/data/api/list_documentTypes/v1/) to load the corresponding document types. This means that dataset will need to be republished to make the new document types available in the contribution tool.

The other possibility would be to not rely on the legacy API anymore and use the [raw url from github](https://raw.githubusercontent.com/ambanum/OpenTermsArchive/main/src/archivist/services/documentTypes.json)

Ndpnt (approved):

PR #930, opened 2022-09-05
Shorten long GitHub comments to prevent errors

### LINKED ISSUE #925: Trim long error messages in tracker
When an error arises with a long message, the generated GitHub issue cannot be created, with the following error:

> Validation Failed: {"resource":"Issue","code":"custom","field":"body","message":"body is too long (maximum is 65536 characters)"}

This happens for the Privacy Policy of TIME in the `contrib` instance.

martinratinaud:
Fixes #925.

I removed the link to the local contribution tool as it doubled the number of characters.

Thus I implemented the trimming of number of characters but I'm not sure it is necessary. And maybe we want the error to be not created instead of truncated, I'd like your thoughts on that.

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -14,7 +14,6 @@ const ISSUE_STATE_ALL = 'all';
 const TRACKER_LABEL = config.get('tracker.githubIssues.label.name');
 const TRACKER_REPOSITORY = config.get('tracker.githubIssues.repository');
 
-const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/en/service';
 const CONTRIBUTE_URL = 'https://contribute.opentermsarchive.org/en/service';
 const GITHUB_REPO_URL = `https://github.com/${TRACKER_REPOSITORY}/blob/main/declarations`;
 const GOOGLE_URL = 'https://www.google.com/search?q=';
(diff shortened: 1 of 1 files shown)


src/tracker/index.js
```diff
@@ -14,10 +14,10 @@ const ISSUE_STATE_ALL = 'all';
 const TRACKER_LABEL = config.get('tracker.githubIssues.label.name');
 const TRACKER_REPOSITORY = config.get('tracker.githubIssues.repository');
 
-const LOCAL_CONTRIBUTE_URL = 'http://localhost:3000/en/service';
 const CONTRIBUTE_URL = 'https://contribute.opentermsarchive.org/en/service';
 const GITHUB_REPO_URL = `https://github.com/${TRACKER_REPOSITORY}/blob/main/declarations`;
 const GOOGLE_URL = 'https://www.google.com/search?q=';
+const MAX_ISSUE_BODY_LENGTH = 65536;
```
MattiSG, line 20:
```suggestion
const GITHUB_API_MAX_ISSUE_BODY_LENGTH = 65536;
```

MattiSG:
> maybe we want the error to be not created instead of truncated

My preference in terms of implementation would be that if the body size is still too long, we do not include a link to the contribution interface but the message is complete. This trim is indeed a bit brutal 😉

MattiSG:
We could also base64-encode the contribution tool payload, I believe this is the way forward anyway as declarations get more complex and we will also hit the URL maximum size.

Ndpnt (approved):

martinratinaud:
@MattiSG I suggest to pur the base64 encoding in another PR as, as explained, I may try to passe the whole json encoded.

For the time being, I think should remove the trim and rename the PR as it is very unlikely for now that we reach this number of characters as we removed one url

MattiSG:
Agreed! 👍

PR #931, opened 2022-09-07
Change page ID tag in filesystem to avoid collision with document types

Ndpnt:


martinratinaud:
I was not aware that you could use `#` in a filesystem name but ok.

Also, you did not put a space at the end while there was one before

I'm not sure to understand why there was conflict with ` - `, can you please explain a bit more

Ndpnt:
> Also, you did not put a space at the end while there was one before

Yes it is voluntary, before it was a dash `-` separator so it made sense to have spaces before and after, now with hashtag `#` I find it more common to have it without space.

> I'm not sure to understand why there was conflict with `-`, can you please explain a bit more

Because on France-elections we created specific document types like `Community Guidelines - Self-harm` so `Self-harm` was interpreted as a page ID.

martinratinaud (approved):

PR #936, opened 2022-09-14
Fix different charset

### LINKED ISSUE #935: Documents with non latin characters fail if encoding is not properly set
Example with [this bulgarian website](https://contribute.opentermsarchive.org/en/service?destination=OpenTermsArchive%2Fp2b-compliance-declarations&documentType=Acceptable%20Use%20Policy&expertMode=true&mode=4&name=Mobile.bg&selectedCss[]=body&url=https%3A%2F%2Fwww.mobile.bg%2Fpcgi%2Fmobile.cgi%3Fact%3D9)

martinratinaud:
Fixes #935 

The `.text()` function is not trying to detect encoding anymore [since version 2](https://github.com/node-fetch/node-fetch/blob/main/docs/v2-UPGRADE-GUIDE.md#text-no-longer-tries-to-detect-encoding)

The new way to handle that kind of encoding is [here for v3](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#dropped-the-restextconverted-function)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "http-proxy-agent": "^5.0.0",
         "https": "^1.0.0",
         "https-proxy-agent": "^5.0.0",
+        "iconv-lite": "^0.6.3",
         "joplin-turndown-plugin-gfm": "^1.0.12",
         "jsdom": "^18.1.0",
         "json-source-map": "^0.6.1",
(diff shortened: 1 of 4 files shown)


martinratinaud:
@Ndpnt do you have any idea why there is a lint problem?

I have the same on my local but can not figure out why as package is correctly installed

martinratinaud:
As discussed with @Ndpnt yesterday, I will merge this without former approval but with oral approval.

I had to bypass the eslint error by disabling `import/no-unresolved`.

As I managed to create a [reproducible environment](https://github.com/OpenTermsArchive/error-reproduction-eslint-fetch-charset-detection), I created [an issue](https://github.com/node-fetch/fetch-charset-detection/issues/247) on their repo.

PR #942, opened 2022-09-20
Update `fetch-charset-detection` to 1.0.1

Ndpnt:
Currently hosted on the @opentermsarchive account and awaiting publication by the original author.

See https://github.com/node-fetch/fetch-charset-detection/issues/249

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "@accordproject/markdown-cicero": "^0.15.2",
     "@accordproject/markdown-pdf": "^0.15.2",
+    "@opentermsarchive/fetch-charset-detection": "^1.0.1",
     "abort-controller": "^3.0.0",
     "ajv": "^6.12.6",
     "archiver": "^5.3.0",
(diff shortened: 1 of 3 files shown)


src/archivist/fetcher/htmlOnlyFetcher.js
```diff
@@ -1,6 +1,6 @@
+import convertBody from '@opentermsarchive/fetch-charset-detection'; // eslint-disable-line import/no-unresolved
```
martinratinaud, line 1: Do we still need the no-unresolved?

martinratinaud (approved):

martinratinaud:
@Ndpnt in fact, why don't you link directly to github instead of forking? which needs more maintenance


"fetch-charset-detection": "git+https://git@github.com/node-fetch/fetch-charset-detection"

src/archivist/fetcher/htmlOnlyFetcher.js
Ndpnt: I guess, there are no changes on that side.

Ndpnt:
> @Ndpnt in fact, why don't you link directly to github instead of forking? which needs more maintenance
> 
> "fetch-charset-detection": "git+https://git@github.com/node-fetch/fetch-charset-detection"

Because the github repository does not contain the built version, so it does not work (I tried).

martinratinaud (approved):

package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "@accordproject/markdown-cicero": "^0.15.2",
     "@accordproject/markdown-pdf": "^0.15.2",
+    "@opentermsarchive/fetch-charset-detection": "^1.0.1",
```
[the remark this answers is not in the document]
Ndpnt: See reply above

PR #946, opened 2022-09-26
Add converter from page and document declaration to JSON

martinratinaud:
To easily pass from in-memory declaration to useable JSON representation to be able to pass it to GitHub issue creation


src/archivist/services/pageDeclaration.test.js
```diff
@@ -158,4 +158,66 @@ describe('PageDeclaration', () => {
       });
     });
   });
+
+  describe('toJSON', () => {
+    it('convert noise and content semectors to JSON', async () => {
```
Ndpnt, line 163:
```suggestion
    it('converts noise and content selectors to JSON', async () => {
```

Ndpnt: By the way, I think to JSON is misnamed because it not only converts the representation to JSON but also renames its attributes to be the same as those found in service declarations.

src/archivist/services/documentDeclaration.test.js
```diff
@@ -0,0 +1,85 @@
+import chai from 'chai';
+
+import DocumentDeclaration from './documentDeclaration.js';
+import PageDeclaration from './pageDeclaration.js';
+
+const { expect } = chai;
+
+describe('PageDeclaration', () => {
```
Ndpnt, line 8:
```suggestion
describe('DocumentDeclaration', () => {
```

src/archivist/services/pageDeclaration.js
```diff
@@ -38,4 +38,14 @@ export default class PageDeclaration {
 
     return [selector];
   }
+
+  toJSON() {
```
Ndpnt, line 42: I think `toJSON` is misnamed because it not only converts the representation to JSON but also renames its attributes to be the same as those found in service declarations.
What about toPersistenceFormat?

src/archivist/services/documentDeclaration.js
```diff
@@ -12,4 +12,15 @@ export default class DocumentDeclaration {
   get isMultiPage() {
     return this.pages.length > 1;
   }
+
+  toJSON() {
```
Ndpnt, line 16: I think `toJSON` is misnamed because it not only converts the representation to JSON but also renames its attributes to be the same as those found in service declarations.
What about toPersistenceFormat?

src/archivist/services/pageDeclaration.test.js
martinratinaud: I check if it works well in different page declaration configurations
- select and remove
- filter and select

martinratinaud:
I left on purpose `page1AsJSON` as `page1AsPersistence` did not look good to me nor `page1AsPersisted` and it is still a JSON afterall

src/archivist/services/documentDeclaration.js
```diff
@@ -12,4 +12,15 @@ export default class DocumentDeclaration {
   get isMultiPage() {
     return this.pages.length > 1;
   }
+
+  toPersistence() {
+    return {
+      name: this.service.name,
+      documents: {
+        [this.type]: this.pages.length === 1
```
Ndpnt, line 20: Use `this.isMultiPage()` instead

src/archivist/services/documentDeclaration.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+      const expectedResult = {
+        name: service.name,
+        documents: { [type]: page1AsJSON },
+      };
+
+      expect(result).to.deep.equal(expectedResult);
+    });
+
+    it('converts multi page document to JSON representation', async () => {
```
Ndpnt, line 74:
```suggestion
    it('converts multipage document to JSON representation', async () => {
```

src/archivist/services/pageDeclaration.test.js
Ndpnt: I do not see the added value of splitting this. Why not `select` and `executeClientScript`? Grouping `select` with `remove` and `select` with `filter` seems arbitrary to me, I do not see the domain constraint on this. 

I'm in favor of a single test that checks that all attributes are properly converted, or else make a test for each attribute, but not an arbitrary grouping.

martinratinaud: I thus made two tests
- one with simple declaration (fetch and select) to make sure other fields are undefined
- one with all fields

src/archivist/services/pageDeclaration.test.js
```diff
@@ -158,4 +158,67 @@ describe('PageDeclaration', () => {
       });
     });
   });
+
+  describe('toPersistence', () => {
```
Ndpnt, line 162:
```suggestion
  describe('#toPersistence', () => {
```

src/archivist/services/documentDeclaration.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+  const page2AsJSON = {
+    fetch: URL,
+    select: 'body',
+    remove: undefined,
+    filter: undefined,
+    executeClientScripts: undefined,
+  };
+
+  describe('toPersistence', () => {
```
Ndpnt, line 62:
```suggestion
  describe('#toPersistence', () => {
```

Ndpnt (approved):

PR #951, opened 2022-10-10
Improve CLI

Ndpnt:
Improve CLI and add possibility to specify document types

Capture d’écran 2022-10-10 à 15 19 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ npm start -- --services <service_id>
 
 > The service ID is the case sensitive name of the service declaration file without the extension. For example, for `Twitter.json`, the service ID is `Twitter`.
 
-To disaply help:
+To display help:
 
 ```
 npm start -- --help
--- a/README.md
+++ b/README.md
@@ -371,10 +371,16 @@ npm run start:scheduler
 To get the latest version of a specific service's terms:
 
 ```
-npm start <service_id>
+npm start -- --services <service_id>
 ```
 
 > The service ID is the case sensitive name of the service declaration file without the extension. For example, for `Twitter.json`, the service ID is `Twitter`.
+
+To disaply help:
+
+```
+npm start -- --help
+```
 
 ## Deploying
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -371,11 +371,17 @@ npm run start:scheduler
 To get the latest version of a specific service's terms:
 
 ```
-npm start <service_id>
+npm start -- --services <service_id>
 ```
 
 > The service ID is the case sensitive name of the service declaration file without the extension. For example, for `Twitter.json`, the service ID is `Twitter`.
 
+To disaply help:
```
martinratinaud, line 379:
```suggestion
To display help:
```

martinratinaud (changes requested):
In order to keep consistency, all the commander code should also be used in the `bin/track.js` so that this feature is available to all declaration repos.

The above change will need to change all declarations package script `test:modified` and `track:modified`

This is a very nice addition, thanks.

Ndpnt:
> A nice plus (or other PR) would be to

?

README.md
```diff
@@ -371,11 +371,17 @@ npm run start:scheduler
 To get the latest version of a specific service's terms:
 
 ```
-npm start <service_id>
+npm start -- --services <service_id>
 ```
 
 > The service ID is the case sensitive name of the service declaration file without the extension. For example, for `Twitter.json`, the service ID is `Twitter`.
```
clementbiron, line 378:
```suggestion
To get the latest version of a specific service's terms and document type:

npm start -- --services <service_id> -d <document_type>
```

clementbiron (approved):

clementbiron:
I feel like it's Christmas, thanks!

Ndpnt:
Declarations template and other declarations repositories should be updated with:
`"track:modified": "git diff --name-only HEAD $(git merge-base HEAD origin/main) -- ./declarations/ | grep -v .history | grep -v .filters | sed 's/declarations\\///g' | sed 's/.json//g' | sed 's/.*/\"&\"/' | xargs -r npm run track -- -s"`

martinratinaud (approved):
I am not a big fan of the double dash but could not find a way out.

For reference, here is [an article](https://medium.com/fhinkel/the-curious-case-of-double-dashes-b5e7711698f) that explains it.

In all declarations repo, users will now need to use `npm run track -- -s <service_id> and tests scripts should be modified in all repos

PR #957, opened 2022-10-17
Automate package release with Semantic versioning

Ndpnt:
This PR trigger a package release on GitHub and NPM with an appropriate [version bump](https://docs.npmjs.com/cli/v8/commands/npm-version) when merged on the `main` branch.

Changelog file is based on [Common Changelog](https://common-changelog.org).

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,4 +1,4 @@
-name: Changelog check
+name: Check changelog
 
 on:
   push:
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,33 @@
+name: Changelog check
+
+on:
+  push:
+    branches-ignore:
+      - main
+      - push-action/** # Temporary branches created by CasperWA/push-protected@v2 action on release workflow
+  pull_request:
+    types: [ opened, reopened ]
+  workflow_call:
+    outputs:
+      release_type:
+        description: The release type extracted from changelog
+        value: ${{ jobs.check_changelog.outputs.release_type }}
+
+jobs:
+  check_changelog:
+    runs-on: [ ubuntu-latest ]
+    outputs:
+      release_type: ${{ steps.set_release_type_output.outputs.release_type }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get release type in changelog
+        run: echo "RELEASE_TYPE=$(cat CHANGELOG.md | grep -E '^## Unreleased \[(patch|minor|major)\]$' | grep -E -w -o "patch|minor|major" | tr -d '\n')" >> $GITHUB_ENV
+      - name: Make release type available to subsequent jobs
+        if: env.RELEASE_TYPE
+        id: set_release_type_output
+        run: |
+          echo "Found release type '${{ env.RELEASE_TYPE }}'"
+          echo "release_type=${{ env.RELEASE_TYPE }}" >> $GITHUB_OUTPUT
+      - name: Fail and display error if no proper release type is found in changelog
+        if: env.RELEASE_TYPE == ''
+        run: echo "No valid release type found in changelog. The title of the 'Unreleased' section must contain one of the following tags '[patch]', '[minor]', '[major]'. For example, '## Unreleased [minor]'."; exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,20 +1,9 @@
 name: Release
 
 on:
-  workflow_dispatch:
-    inputs:
-      release-type:
-        type: choice
-        description: Release type
-        options:
-        - patch
-        - minor
-        - major
-        - prepatch
-        - preminor
-        - premajor
-        - prerelease
-        
+  push:
+    branches: [ main ]
+
 jobs:
   test:
     uses: "ambanum/OpenTermsArchive/.github/workflows/test.yml@main"
@@ -28,24 +17,16 @@ jobs:
       - name: Git configuration
         run: |
           git config --global user.name "GitHub Actions"
-          git config --global user.email "github-action@users.noreply.github.com"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: Get release type
+        run: echo "RELEASE_TYPE=$(cat CHANGELOG.md | grep '## Unreleased' | sed -r 's/^## Unreleased \[(patch|minor|major)\]/\1/g')" >> $GITHUB_ENV
+          
 
       - name: Bump package version
-        if: startsWith(github.event.inputs.release-type, 'pre') != true
         run: |
-          echo "NEW_VERSION=$(npm --no-git-tag-version version $RELEASE_TYPE)" >> $GITHUB_ENV
+          echo "NEW_VERSION=$(npm --no-git-tag-version version ${{ env.RELEASE_TYPE }})" >> $GITHUB_ENV
           echo "RELEASE_TAG=latest" >> $GITHUB_ENV
-        env:
-          RELEASE_TYPE: ${{ github.event.inputs.release-type }}
-
-      - name: Bump package pre-release version
-        # Use tag "beta" for pre-release versions
-        if: startsWith(github.event.inputs.release-type, 'pre')
-        run: |
-          echo "NEW_VERSION=$(npm --no-git-tag-version --preid=beta version $RELEASE_TYPE)" >> $GITHUB_ENV
-          echo "RELEASE_TAG=beta" >> $GITHUB_ENV
-        env:
-          RELEASE_TYPE: ${{ github.event.inputs.release-type }}
 
       - name: Update changelog unreleased section with new version
         uses: superfaceai/release-changelog-action@v1
@@ -64,16 +45,6 @@ jobs:
 
       - name: Add authentication token for NPM
         run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPMJS_ACCESS_TOKEN }}" > ~/.npmrc
-      
-      - name: Publish version to public repository
-        run: npm publish --verbose --access public --tag ${{ env.RELEASE_TAG }}
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}
-
-      - name: Push changes to repository
-        run: git push origin && git push --tags
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Read version changelog
         uses: superfaceai/release-changelog-action@v1
@@ -83,11 +54,20 @@ jobs:
           version: ${{ env.NEW_VERSION }}
           operation: read
 
+      - name: Push changes to repository
+        run: git push origin && git push --tags
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Update GitHub release with changelog
         uses: softprops/action-gh-release@v1
         with:
           tag_name: ${{ env.NEW_VERSION }}
           body: ${{ steps.get-changelog.outputs.changelog }}
-          prerelease: ${{ startsWith(github.event.inputs.release-type, 'pre') }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish version to public repository
+        run: npm publish --verbose --access public --tag ${{ env.RELEASE_TAG }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,9 +43,6 @@ jobs:
           git commit -m "Release ${{ env.NEW_VERSION }}"
           git tag ${{ env.NEW_VERSION }}
 
-      - name: Add authentication token for NPM
-        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPMJS_ACCESS_TOKEN }}" > ~/.npmrc
-
       - name: Read version changelog
         uses: superfaceai/release-changelog-action@v1
         id: get-changelog
@@ -67,7 +64,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish version to public repository
+      - name: Setup Node to allow publishing to NPM
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Publish version to NPM public repository
         run: npm publish --verbose --access public --tag ${{ env.RELEASE_TAG }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,9 +19,8 @@ jobs:
           git config --global user.name "GitHub Actions"
           git config --global user.email "github-actions@users.noreply.github.com"
 
-      - name: Get release type
-        run: echo "RELEASE_TYPE=$(cat CHANGELOG.md | grep '## Unreleased' | sed -r 's/^## Unreleased \[(patch|minor|major)\]/\1/g')" >> $GITHUB_ENV
-          
+      - name: Get release type from the changelog
+        run: echo "RELEASE_TYPE=$(cat CHANGELOG.md | grep -E '^## Unreleased \[(patch|minor|major)\]$' | grep -E -w -o "patch|minor|major")" >> $GITHUB_ENV
 
       - name: Bump package version
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,20 @@
 name: Release
 
 on:
-  push:
-    branches: [ main ]
+  pull_request:
+    branches: 
+      - main
+    types: [ closed ]
+
 
 jobs:
+  changelog:
+    uses: "ambanum/OpenTermsArchive/.github/workflows/changelog.yml@main"
   test:
     uses: "ambanum/OpenTermsArchive/.github/workflows/test.yml@main"
   release:
-    needs: [ test ]
+    if: github.event.pull_request.merged == true
+    needs: [ changelog, test ]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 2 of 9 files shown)


Ndpnt:
Note: the release workflow will be available on the `Actions` tab only when this PR is merged on the `main` branch.

martinratinaud (approved):
Very nice

I'm wondering if we should add a 0.0.0 version in the changelog with what Open Terms Archive already do or at least put a date when we start tracking the changes
Something like
```
## v0.0.0 2022-10-18

### Added

Start tracking changes
```

martinratinaud:
I approved but could not test of course.

When this is done, we should update all instances with the npm version

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "open-terms-archive",
-  "version": "0.15.0",
-  "description": "Tracks and makes visible all changes to the Terms Of Service of online service providers.",
+  "version": "0.0.0",
+  "description": "Tracks and makes visible changes to the Terms of Service of online service",
```
martinratinaud, line 4: Is `Terms of Service` the new vocabulary to talk about service documents?

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+    needs: [ test ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Git configuration
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "github-action@users.noreply.github.com"
```
martinratinaud, line 31: Why not use OTA bot for this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          git add "package.json"
+          git add "package-lock.json"
+          git add "CHANGELOG.md"
+          git commit -m "Release ${{ env.NEW_VERSION }}"
+          git tag ${{ env.NEW_VERSION }}
+
+      - name: Add authentication token for NPM
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPMJS_ACCESS_TOKEN }}" > ~/.npmrc
+      
+      - name: Publish version to public repository
```
martinratinaud, line 68: Is it normal to push after publishing? I would have done the opposite

martinratinaud (changes requested):

MattiSG:
As seen in a synchronous call with @ndpnt, we want to:

- Make deployment automated and continuous.
- Use [Common Changelog](https://common-changelog.org) instead of Keep a Changelog.
- Document this in the `CONTRIBUTING`.
- Obtain version number bump from a tag in the title: `## Unreleased [patch] [minor] [major]`.
  - This will enable learning. Over time, when we have consistent interpretation, we might automate that from other signals from the Changelog.
- Remove dist-tags management since they are irrelevant in a CD context.

.github/workflows/release.yml
MattiSG: Indeed, it is easier to recover from an unpublished but pushed state than from an un-pushed but published state which might need yanking from the package manager.

.github/workflows/release.yml
MattiSG: I believe it is good to properly assign changes to the actor that is responsible for it. In this case, it is not @OTA-Bot but an unauthenticated GitHub Actions process that will create this commit 🙂 

However, let's unify name and email to the official GitHub Actions name:
```suggestion
          git config --global user.email "github-actions@users.noreply.github.com"
```

package.json
MattiSG:
```suggestion
  "description": "Tracks and makes visible changes to the contractual documents of online services",
```

MattiSG:
I don't find it relevant to downgrade the version to `0.0.0`. This will lead to multiple versions in the VCS having the same version number. I don't think it's a problem if the first published version is `0.16.0` 🙂

MattiSG:
We want to publish on NPM under the name `open-terms-archive`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+  test:
+    uses: "ambanum/OpenTermsArchive/.github/workflows/test.yml@main"
+  release:
+    needs: [ test ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Git configuration
```
MattiSG, line 17: Let's use active verbs for all steps names 🙂 

```suggestion
      - name: Configure Git author
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: Get release type
+        run: echo "RELEASE_TYPE=$(cat CHANGELOG.md | grep '## Unreleased' | sed -r 's/^## Unreleased \[(patch|minor|major)\]/\1/g')" >> $GITHUB_ENV
```
MattiSG, line 23: Considering how critical this parsing is, let's add a failsafe: we should test if the `Unreleased` line contains one of `patch|minor|major` before attempting to publish.
It might actually make sense to test this continuously, in order to get used to always filling in the Changelog as we work.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Commit CHANGELOG.md and package.json changes and create tag
+        run: |
+          git add "package.json"
+          git add "package-lock.json"
+          git add "CHANGELOG.md"
+          git commit -m "Release ${{ env.NEW_VERSION }}"
+          git tag ${{ env.NEW_VERSION }}
+
+      - name: Add authentication token for NPM
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPMJS_ACCESS_TOKEN }}" > ~/.npmrc
```
MattiSG, line 47: Wouldn't it be simpler to use https://github.com/marketplace/actions/npm-publish? 🙂

In any case, it looks like we pass the token as an env variable in the only further step where we use NPM, so why is this instruction necessary?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+          path-to-changelog: CHANGELOG.md
+          version: ${{ env.NEW_VERSION }}
+          operation: read
+
+      - name: Push changes to repository
+        run: git push origin && git push --tags
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update GitHub release with changelog
```
MattiSG, line 62:
```suggestion
      - name: Create GitHub release
```

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format is based on [Common Changelog](https://common-changelog.org).\
+Unlike Common Changelog, the `unreleased` section of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) is preserved with the addition of a tag to specify which type of release should be published and to foster discussions about it inside pull requests. This tag should be one of the following: `[patch]`, `[minor]`, `[major]`. Ex `Unreleased [minor]`.
+
+## [Unreleased][minor]
```
MattiSG, line 8: ? 🤔 
```suggestion
## Unreleased [minor]
```

This seems to be a good demonstration we need to check in CI that the format is correct 😄

CONTRIBUTING.md
```diff
@@ -52,6 +53,10 @@ We add this additional rule:
 
 - Do not rely on GitHub issue reference numbers in commit messages, as we have no guarantee the host system and its autolinking will be stable in time. Make sure the context is self-explanatory. If an external reference is given, use its full URL.
 
+### Changelog
+
+All notable changes to this project must be documented in the [`CHANGELOG.md`](./CHANGELOG.md) file.
```
MattiSG, line 58:
```suggestion
All changes to the codebase that impact users must be documented in the [`CHANGELOG.md`](./CHANGELOG.md) file. The format to use is documented in the file itself.
```

package.json
```diff
@@ -14,6 +14,7 @@
   "author": "ambanum",
   "type": "module",
   "exports": {
+    ".": "./src/exports.js",
     "./fetch": "./src/archivist/fetcher/exports.js",
     "./filter": "./src/archivist/filter/exports.js",
     "./page-declaration": "./src/archivist/services/pageDeclaration.js"
```
MattiSG, line 20: Is this still needed when we now list exports in `exports.js`?

MattiSG (changes requested):

CHANGELOG.md
Ndpnt: Both formats are valid and supported by keep-a-changelog

MattiSG: Yes, but we don't parse that format with our `sed` 😉

package.json
Ndpnt: Yes, it allows to import in two different ways:
- `import { Fetch } from 'opentermsarchive';`
- `import Fetch from 'opentermsarchive/fetch';`

.github/workflows/release.yml
Ndpnt: We can't use `npm-publish` as its trigger is only a change in the version number in `package.json`.

MattiSG: Isn't this perfect if we use it after `npm version`? 🙂

Ndpnt: You're right, it could be done like that.

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format is based on [Common Changelog](https://common-changelog.org).\
+Unlike Common Changelog, the `unreleased` section of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) is preserved with the addition of a tag to specify which type of release should be published and to foster discussions about it inside pull requests. This tag should be one of the following: `[patch]`, `[minor]`, `[major]`. Ex `Unreleased [minor]`.
+
+## Unreleased [minor]
+
+### Added
+
+- Automate release with GitHub Actions
```
MattiSG, line 12: This is of little interest to reusers. I believe it would be more useful to document what impacts reusers 🙂 
```suggestion
- Publish package on NPM under name `opentermsarchive`.
- Export `filter`, `pageDeclaration` and `fetch` in NPM module.
- Add changelog.
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node to allow publishing to NPM
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Publish version to NPM public repository
+        run: npm publish --verbose --access public --tag ${{ env.RELEASE_TAG }}
```
MattiSG: Public access and `latest` tag are [default](https://docs.npmjs.com/cli/v9/commands/npm-publish?v=true#access), why specify them? 🙂 
```suggestion
        run: npm publish --verbose
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          tag_name: ${{ env.NEW_VERSION }}
+          body: ${{ steps.get-changelog.outputs.changelog }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node to allow publishing to NPM
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          registry-url: 'https://registry.npmjs.org'
```
MattiSG: Inconsistent quote style.

MattiSG (changes requested):

.github/workflows/release.yml
Ndpnt: Not relevant anymore as it uses `npm-publish` now.

.github/workflows/release.yml
Ndpnt: Not relevant anymore as it uses `npm-publish` now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Bump package version
+        run: |
+          echo "NEW_VERSION=$(npm --no-git-tag-version version ${{ env.RELEASE_TYPE }})" >> $GITHUB_ENV
+          echo "RELEASE_TAG=latest" >> $GITHUB_ENV
+
+      - name: Update changelog unreleased section with new version
+        uses: superfaceai/release-changelog-action@v1
+        with:
+          path-to-changelog: CHANGELOG.md
```
MattiSG, line 34: https://github.com/superfaceai/release-changelog-action/pull/248 has been merged 🙂 
```suggestion
```

martinratinaud (approved):
This looks really good to me and very valuable.

Congrats to both of you

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format is based on [Common Changelog](https://common-changelog.org).\
```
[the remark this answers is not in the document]
Ndpnt: Line break

.github/workflows/test.yml
```diff
@@ -9,6 +9,13 @@ on:
   workflow_call:
 
 jobs:
+  check_changelog:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check for proper release type in changelog
+        run: $(cat CHANGELOG.md | grep '## Unreleased \[.*\]' | grep -E -w -o -q "patch|minor|major") || (echo "No valid release type found in changelog. The title of the 'Unreleased' section must contain one of the following tag [patch], [minor], [major]. Ex Unreleased [minor]."; exit 1)
```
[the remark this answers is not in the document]
Ndpnt: I have not succeeded in sharing an output variable between workflows. From my research it seems that this is not possible at the moment, it is only possible between tasks within the same workflow.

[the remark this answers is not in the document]
Ndpnt: Finally, it's done

MattiSG (changes requested):
In order to approve, I would need to see a proof of this code working 🙂 
For example, we could create another branch from this one, or add a `[TOREMOVE]` commit, with the trigger branch being that one instead of `main`. Ideally, we would also amend the NPM publish option to publish to another package name, or with a `release-candidate` tag instead of `latest`, but since nobody depends on this package yet, I would also be fine with publishing as it is. In any case, I could find no run of the `release` workflow in our [Actions log](https://github.com/ambanum/OpenTermsArchive/actions/runs/3532937266) and thus cannot guarantee that this will not break when merged into `main`.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    branches-ignore:
+      - simulated-main-for-testing-packaging
+      - push-action/** # Temporary branches created by CasperWA/push-protected@v2 action on release workflow
+  pull_request:
+    types: [ opened, reopened ]
+  workflow_call:
+    outputs:
+      release_type:
+        description: "The release type extracted from changelog"
```
MattiSG, line 13:
```suggestion
        description: The release type extracted from changelog
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
+name: Release
+
+on:
+  pull_request:
+    branches: 
+      - simulated-main-for-testing-packaging
+    types: [closed]
```
MattiSG, line 7:
```suggestion
    types: [ closed ]
```

src/archivist/recorder/repositories/git/index.test.js
```diff
@@ -471,7 +471,9 @@ describe('GitRepository', () => {
     let records;
     const expectedIds = [];
 
-    before(async () => {
+    before(async function () {
+      this.timeout(5000);
```
MattiSG, line 475: 🥷 Please move this to a separate PR.

.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   test:
     strategy:
       matrix:
-        operating_system: [ ubuntu-latest, windows-latest, macos-latest ]
+        operating_system: [ ubuntu-20.04, windows-latest, macos-latest ]
```
MattiSG, line 15: 🥷 Please move this to a separate PR.

.github/workflows/test.yml
```diff
@@ -3,7 +3,7 @@ name: Test
 on:
   push:
     branches-ignore:
-      - main # tests will be launched by workflow_call from the deploy workflow
+      - simulated-main-for-testing-packaging # tests will be launched by workflow_call from the deploy workflow
```
MattiSG, line 6: Okay, proof well established, should be reverted!

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+          operation: read
+
+      - name: Create GitHub release
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ env.NEW_VERSION }}
+          body: ${{ steps.get-changelog.outputs.changelog }}
+          token: ${{ secrets.RELEASE_BOT_GITHUB_TOKEN }}
+
+      - name: Publish version to NPM public repository
```
MattiSG, line 73:
```suggestion
      - name: Publish to NPM public repository
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+        uses: JS-DevTools/npm-publish@v1
+        with:
+          token: ${{ secrets.NPMJS_ACCESS_TOKEN }}
+          dry-run: true
```
MattiSG, line 77: Okay, proof well established, should be removed.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,33 @@
+name: Changelog check
```
MattiSG, line 1: Always use active verbs in workflow names as they will be displayed in the UI 🙂 

```suggestion
name: Check changelog
```

MattiSG (changes requested):

MattiSG (approved):
Approving even with the ninja commits since now that #974 is merged the conflict will have to be handled anyway 😉

MattiSG:
I expect a conflict will happen on the `v0.17.0` tag that has already been added to the repo, it should probably be removed before merging  🙂

Ndpnt:
> I expect a conflict will happen on the `v0.17.0` tag that has already been added to the repo, it should probably be removed before merging 🙂

Already removed ;)

PR #958, opened 2022-10-24
Better OTA scripts to handle modified files

martinratinaud:
For now there is a very complicated bash command in all declaration repositories to run `lint` `test` and `track`

All this complexity can be handled directly in the scripts and this is waht this PR does


martinratinaud:
@Ndpnt please validate the concept of this PR so that I can go on with it tomorrow for `track` and `test` command.

Idea here is to
- use `commander` instead of args parsing
- add a `-m` option to every script in order to launch the script only on modified files

scripts/utils/git/index.js
```diff
@@ -0,0 +1,47 @@
...
+  const servicesDocumentTypes = modifiedServiceIds.reduce((acc, val) => ({ ...acc, [val]: [] }), {});
+
+  await Promise.all(modifiedFilePaths.map(async modifiedFilePath => {
+    try {
+      const serviceId = filePathToServiceId(modifiedFilePath);
+      const fileInMain = JSON.parse(await git.show([`main:${modifiedFilePath}`]));
+      const fileModified = JSON.parse(await git.show([`HEAD:${modifiedFilePath}`]));
+      const modifiedDocumentTypes = diff.diff(fileInMain, fileModified).map(({ path }) => (path[0] === 'documents' ? path[1] : path[0]));
+
+      servicesDocumentTypes[serviceId] = uniq(modifiedDocumentTypes);
```
Ndpnt, line 32: Does not work when service file is modified outside the `documents` key. (For example when changing service name)

```suggestion
const modifiedDocumentTypes = DeepDiff.diff(fileInMain, fileModified).reduce((acc, { path }) => {
        if (modifiedFilePath.includes('.history')) {
          acc.add(path[0]);
        } else if (path[0] == 'documents') {
          acc.add(path[1]);
        }

        return acc;
      }, new Set());

      servicesDocumentTypes[serviceId] = Array.from(modifiedDocumentTypes);
```

martinratinaud:
This PR has been revamped after discussion with @Ndpnt 

I created a branch in sandbox-declarations in order to test this behavior https://github.com/OpenTermsArchive/sandbox-declarations/pull/149

It handles cases like
- deleted service
- modified history service
- modified service
- modified filter

When this PR is accepted and the one on sandbox also, a PR like [this one](https://github.com/OpenTermsArchive/contrib-declarations/pull/694) should be done on all instances.

Let's note here that the process of updating multiple instances everytime a script changes is somehow cumbersome and we should definitely begin thinking about a new repo like `declarations-tools` which could be installed globally.

This would remove the need for a package.json in the declarations repo (except in those which need filtering)

bin/validate-declarations.js
```diff
@@ -5,29 +5,42 @@ import fs from 'fs';
...
   // Re-throw them so that the validation command fails in these cases (for example, if there is a syntax error when parsing JSON declaration files)
   throw reason;
 });
 
-// Initialise configs to allow clients of this module to use it without requiring node-config in their own application.
-// see https://github.com/lorenwest/node-config/wiki/Sub-Module-Configuration
-config.util.setModuleDefaults('services', { declarationsPath: path.resolve(process.cwd(), './declarations') });
-config.util.setModuleDefaults('fetcher', defaultConfigs.fetcher);
+program
+  .name('validate declaration files')
```
Ndpnt, line 32: Use the exported name of the command
```suggestion
  .name('ota-validate-declarations')
```

bin/validate-declarations.js
```diff
@@ -5,29 +5,42 @@ import fs from 'fs';
...
+  .description('check if document declarations are valid and linted properly')
```
Ndpnt, line 33: Start description with a capital.

Ndpnt, line 33:
```suggestion
  .description('Run a series of tests to check the validity of document declarations')
```

bin/validate-declarations.js
```diff
@@ -39,6 +52,10 @@ const VALIDATE_PATH = path.resolve(__dirname, '../scripts/validation/validate.js
 
   let hasFailedTests = false;
 
+  const addValidationTestSuite = (await import('../scripts/declarations/validate/index.mocha.js')).default;
+
+  addValidationTestSuite(program.parse(process.argv).opts());
```
Ndpnt, line 57: `process.argv` is optional here.

```suggestion
  addValidationTestSuite(program.parse().opts());
```

bin/lint-declarations.js
```diff
@@ -1,45 +1,28 @@
...
+  .version(version)
+  .option('-s, --services [serviceId...]', 'service IDs of services to handle')
+  .option('-m, --modified', 'to only lint modified services already commited to git');
 
-    await ESLint.outputFixes(lintResults);
-    console.log(lintResults.map(lintResult => `${path.basename(lintResult.filePath)} linted`).join('\n'));
-    /* eslint-enable no-await-in-loop */
-  }
-})();
+lintDeclarations(program.parse(process.argv).opts());
```
Ndpnt, line 28: `process.argv` is optional here.

```suggestion
lintDeclarations(program.parse().opts());
```

bin/validate-declarations.js
Ndpnt, line 57: Maybe it's better `generateValidationTestSuite`?

scripts/declarations/lint/index.js
```diff
@@ -0,0 +1,34 @@
+import path from 'path';
```
Ndpnt, line 1: Bugs with deleted files when using -m option. 

Maybe `getModifiedServiceDocumentTypes` should not return deleted services?

scripts/declarations/utils/index.js
```diff
@@ -0,0 +1,74 @@
...
+export const getModifiedServiceDocumentTypes = async instancePath => {
+  const { git, modifiedFilePaths, modifiedServiceIds } = await getModifiedData(instancePath);
+  const defaultBranch = 'main'; // could be retrieved from git
+  const servicesDocumentTypes = {};
+
+  await Promise.all(modifiedFilePaths.map(async modifiedFilePath => {
+    const serviceId = filePathToServiceId(modifiedFilePath);
+
+    if (!modifiedFilePath.endsWith('.json')) {
+      // here we should compare AST of both files to detect on which function
```
Ndpnt, line 41: Always start comment with a capital.

Ndpnt (changes requested):
For consistency, also move `track` command description in `./bin/track.js`, remove `src/main.js` and update npm start script in `package.json`.

Ndpnt:
Nice addition, it will simplify declarations repositories! 👍

scripts/declarations/lint/index.js
martinratinaud: Agreed

martinratinaud:
> For consistency, also move `track` command description in `./bin/track.js`, remove `src/main.js` and update npm start script in `package.json`.

For this I'd like to postpone this until you come back @Ndpnt.

Indeed, I'm still not quite sure I understand correctly how config.util.setModuleDefaults works, but it seems that if I do so, it will override the corresponding defaults from `config/default.json` and thus make all instance repositories to not work anymore as they rely for example on `services/declarationsPath` which would be not accurate anymore.

Ndpnt (approved):

PR #959, opened 2022-11-01
Handle all cases of email protection to prevent blinking and ease maintenance

martinratinaud:
Here is an implementation of a global filter that aims at removing blink on several documents in an harmonized way.

For now, there has been an implementation that does not cover all cases and this lack of universality in the process is causing blinks on several instances and forcing contributors to [create a filter](https://github.com/OpenTermsArchive/p2b-compliance-declarations/blob/main/declarations/Fairbnb.coop.filters.js) for which we already have at least 3 versions.

Here are some affected documents

- [BlaBlaCar](https://github.com/OpenTermsArchive/france-versions/commit/724461ac41466ec152eb2501a7b0eb93b7821cb9)
- Cityscoot on France
- FairCoop on P2B
- and many others

Thanks to this PR, blink will not occur anymore on any instance and will remove the need of custom filters, thus improcing stability and maintainability

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format is based on [Common Changelog](https://common-changelog.org).\
 Unlike Common Changelog, the `unreleased` section of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) is preserved with the addition of a tag to specify which type of release should be published and to foster discussions about it inside pull requests. This tag should be one of the names mandated by SemVer, within brackets: `[patch]`, `[minor]` or `[major]`. For example: `## Unreleased [minor]`.
 
-## Unreleased
+## Unreleased [minor]
+
+Full changeset and discussions: [#959](https://github.com/ambanum/OpenTermsArchive/pull/959)._
+
+### Changed
+
+- Improved reliability and expanded coverage of email protection global filter
 
 ## 0.19.1 - 2022-12-13
 ### Fixed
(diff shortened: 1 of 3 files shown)


MattiSG (changes requested):
Thanks @martinratinaud for this improvement suggestion! Indeed, these email protection practices are often a source of blinks and since they seem very similar, it would make sense to consolidate one “best” way to handle them.

However, the notion of “global filters” has not been properly defined yet, and I don't think they are a very good idea. I would prefer that we, for example, offer a standard library of common filters that users can then activate if needed (see this [🔒 draft issue](https://github.com/orgs/ambanum/projects/1/views/4?filterQuery=windstream)), or that we add a verb in declarations. In any case, having hidden, invisible transformations made on content is dangerous and can lead to very hard to diagnose issues on specific documents.

In any case, the test cases updated here do not demonstrate the improvements proposed.

Let's discuss this with @Ndpnt  🙂

martinratinaud:
Noted.

I hope we find a way to handle this easily as if it was to be removed, it would create many blinks that would make maintenance harder.

Concerning the test case, it has been replaced by the new way of handling this kind of links so I'm not sure what you mean by 

> the test cases updated here do not demonstrate the improvements proposed

Can you please be more specific?

Ndpnt (approved):

martinratinaud:
@MattiSG asked to rewrite the test cases but I have a workshop tomorrow and have no time to take care of that now.

Feel free to change it if you find the time.

martinratinaud:
@Ndpnt on [this snapshot](https://github.com/OpenTermsArchive/france-snapshots/commit/c7bfe571ad431032e3bb3c028d3a8304962a27ec), line 616

`<a href="/cdn-cgi/l/email-protection#2d4e4243594c4e596d4e4459545e4e424259034858">conta<span class="__cf_email__" data-cfemail="87e4f3c7e4eef3fef4e4e8e8f3a9e2f2">[email&#160;protected]</span></a>`, 

You can see the `conta<span` 

In the previous version, the code was only replacing the `href` whereas this version replaces the whole content of the `a`.

It is true though that it will not work on `BlaBlaCar` as the link is http://dataprotection@blablacar.com.
Maybe we should add a test for `dataprotection@*`

I pushed a test for this behaviour

let me know if it's alright

MattiSG (approved):

PR #962, opened 2022-11-15
Document namespaces naming

MattiSG:
This changeset describes a policy for usernames and namespaces reservation across universes where the Open Terms Archive brand is to be found under a user ID (instead of full text).


martinratinaud (approved):
👍

Ndpnt (approved):

PR #970, opened 2022-11-30
Fix GitHub issue link formatting

### LINKED ISSUE #967: Bot Report Tickets have faulty description syntax
*I did not find a bot repository, and this repo has the dom fetchers throw fetch errors, so I assume the bot repo is private. So I am creating this ticket here.*

1. The inline code syntax is used across two lines and list

image

Presumably the error type was supposed to be code-marked, so the closing backtick is missing after it, and the one after the URL is wrong.

2. Using `'` around the URL erroneously links it as part of the href

image

In Markdown `<url>` can be used for urls as-is - so that should be used for the URL.

Kissaki:
Previously, errors were reported on a single line, where the inline code-fencing on message worked.

The current logic is:
The html fetcher throws FetchDocumentErrors, with the URL fenced with apostrophe `'`. Archivist collects these errors in InaccessibleContentError, which produces a multiline error message (generic text + foreach(\n- error)).

Because the message is multiline now, the inline code fencing with backticks no longer worked.

As a side effect, the GitHub URL identification included the closing \'\` in the URL. Without the \` the \' is not included in the URL href.

So dropping the now invalid backticks resolves both the invalid backticks and wrong URL linking.

Fixes #967

### DIFF
--- a/src/tracker/index.js
+++ b/src/tracker/index.js
@@ -197,7 +197,7 @@ export default class Tracker {
     const body = `
 This document is no longer properly tracked.
 
-\`${message}\`
+${message}
 
 Check what's wrong by:
 - Using the [online contribution tool](${CONTRIBUTE_URL}?${urlQueryParams}).


MattiSG (approved):
👌

PR #992, opened 2023-02-07
Do not launch tracking directly when scheduling

### LINKED ISSUE #986: Rate Limit GitHub on deployment
When a deployment happens, it can happen that the [rate limit for GitHub](https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps) is reached.

```
Default server-to-server rate limits for GitHub.com
GitHub Apps making server-to-server requests use the installation's minimum rate limit of 5,000 requests per hour. If an application is installed on an organization with more than 20 users, the application receives another 50 requests per hour for each user. Installations that have more than 20 repositories receive another 50 requests per hour for each repository. The maximum rate limit for an installation is 12,500 requests per hour.

Server-to-server rate limits for GitHub Enterprise Cloud
GitHub Apps that are installed on an organization within an enterprise on GitHub.com are subject to a limit of 15,000 requests per hour per organization that has installed the app.
```

So several options:
- rework the error workflow to use fewer api calls
- Subscribe to GitHub Enterprise Cloud
- Ask the GitHub community if the rate can be increased on a project like Open Terms Archive

martinratinaud:
Fixes #986


MattiSG (approved):
Most impactful line change yet in this codebase 😄

PR #996, opened 2023-02-27
Apply vocabulary changes [1/3]

Ndpnt:
_In order to facilitate the reviews related to the application of the new vocabulary on the code base and the associated conceptual changes, this work will be divided into several sequential PRs._

_This one is the first in this series._

_Also, the changelog will be updated at the end, before merging this PR in the main branch._

PR #999, opened 2023-03-07
Apply vocabulary changes [2/3]

Ndpnt:
_This PR follows the PR #996._

It applies vocabulary changes on the data layer.
Thus, the data will be stored in a new way on the data layer, but the backwards compatibility is preserved to allow mixing the old and new way of storing data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ _Full changeset and discussions: [#996](https://github.com/OpenTermsArchive/engi
 - **Breaking:** Rename CLI option `--refilter-only`, `-r` to `--extract-only`, `-e` in API; simply rename accordingly in your own codebase
 - **Breaking:** Rename function and its parameters `filter({ content, mimeType, pageDeclaration })` to `extract({ content, mimeType, sourceDocument })` in API; simply rename accordingly in your own codebase
 - **Breaking:** Rename class `PageDeclaration` to `SourceDocument` and its atribute `noiseSelectors` to `insignificantContentSelectors` in API; simply rename accordingly in your own codebase
+- **Breaking:** Rephrase commit messages in Git data layer but keep backwards compatibility for existing data; `Start tracking` to `Initial record of`, `Refilter` to `Apply technical or declaration upgrade on` and `Update` to `Record new changes of`; if you have scripts that parse commit messages directly, update them accordingly
+- **Breaking:** Rename document attribute `isRefilter` to `isExtractOnly` in MongoDB data layer but keep backwards compatibility for existing data; if you have scripts that query the Mongo database directly, update them accordingly
 - Make vocabulary consistent throughout the codebase ([#971](https://github.com/OpenTermsArchive/engine/pull/971))
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ _Full changeset and discussions: [#996](https://github.com/OpenTermsArchive/engi
 - **Breaking:** Rename CLI option `--refilter-only`, `-r` to `--extract-only`, `-e` in API; simply rename accordingly in your own codebase
 - **Breaking:** Rename function and its parameters `filter({ content, mimeType, pageDeclaration })` to `extract({ content, mimeType, sourceDocument })` in API; simply rename accordingly in your own codebase
 - **Breaking:** Rename class `PageDeclaration` to `SourceDocument` and its atribute `noiseSelectors` to `insignificantContentSelectors` in API; simply rename accordingly in your own codebase
-- **Breaking:** Rephrase commit messages in Git data layer but keep backwards compatibility for existing data; `Start tracking` to `Initial record of`, `Refilter` to `Apply technical or declaration upgrade on` and `Update` to `Record new changes of`; if you have scripts that parse commit messages directly, update them accordingly
+- **Breaking:** Rephrase commit messages in Git data layer but keep backwards compatibility for existing data; `Start tracking` to `First record of`, `Refilter` to `Apply technical or declaration upgrade on` and `Update` to `Record new changes of`; if you have scripts that parse commit messages directly, update them accordingly
 - **Breaking:** Rename document attribute `isRefilter` to `isExtractOnly` in MongoDB data layer but keep backwards compatibility for existing data; if you have scripts that query the Mongo database directly, update them accordingly
 - Make vocabulary consistent throughout the codebase ([#971](https://github.com/OpenTermsArchive/engine/pull/971))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All changes that impact users of this module are documented in this file, in the
 
 ## Unreleased [minor]
 
-_Full changeset and discussions: [#996](https://github.com/OpenTermsArchive/engine/pull/996)._
+_Full changeset and discussions: [#996](https://github.com/OpenTermsArchive/engine/pull/996), [#999](https://github.com/OpenTermsArchive/engine/pull/999)._
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -11,7 +11,12 @@ All changes that impact users of this module are documented in this file, in the
 - **Breaking:** In exposed CLI `--terms-types` option is now called `--types`. Short option is stil `-t`.
 - **Breaking:** In exposed API `filter` function is now called `extract`.
 - **Breaking:** In exposed API `PageDeclaration` class is now called `SourceDocument`.
-
+- **Breaking:** Change commit messages in Git data layer but keep backwards compatibility:
```
MattiSG, line 14: If backwards compatibility is maintained, then this is not a breaking change 🙂 
In this case, I believe we should be explicit that the “backwards compatibility” is about reading from existing data repositories.

src/archivist/recorder/repositories/git/dataMapper.js
```diff
@@ -7,27 +7,30 @@ import Record from '../../record.js';
 mime.define({ 'text/markdown': ['md'] }, true); // ensure extension for markdown files is `.md` and not `.markdown`
 
 export const COMMIT_MESSAGE_PREFIX = {
-  startTracking: 'Start tracking',
-  refilter: 'Refilter',
-  update: 'Update',
+  startTracking: 'Initial record of',
```
MattiSG, line 10: > I wonder if “initial” makes it sound as if it was a special business term.
```suggestion
  startTracking: 'First record of',
```

src/archivist/recorder/repositories/git/dataMapper.js
```diff
@@ -7,27 +7,30 @@ import Record from '../../record.js';
...
+  deprecated_startTracking: 'Start tracking',
+  deprecated_refilter: 'Refilter',
+  deprecated_update: 'Update',
 };
 
 export const TERMS_TYPE_AND_DOCUMENT_ID_SEPARATOR = ' #';
 export const SNAPSHOT_ID_MARKER = '%SNAPSHOT_ID';
-const SINGLE_SNAPSHOT_PREFIX = 'This version was recorded after filtering snapshot';
-const MULTIPLE_SNAPSHOT_PREFIX = 'This version was recorded after filtering and assembling the following snapshots from %NUMBER source documents:';
+const SINGLE_SNAPSHOT_PREFIX = 'This version was recorded after an extraction from the snapshot';
```
MattiSG, line 20: > Should we unify with wording from #996?
```suggestion
const SINGLE_SOURCE_DOCUMENT_PREFIX = 'This version was recorded after an extraction from the snapshot';
```

src/archivist/recorder/repositories/git/dataMapper.js
```diff
@@ -7,27 +7,30 @@ import Record from '../../record.js';
 mime.define({ 'text/markdown': ['md'] }, true); // ensure extension for markdown files is `.md` and not `.markdown`
 
 export const COMMIT_MESSAGE_PREFIX = {
```
MattiSG, line 9:
```suggestion
export const COMMIT_MESSAGE_PREFIXES = {
```

src/archivist/recorder/repositories/git/index.test.js
```diff
@@ -8,15 +8,15 @@ import mime from 'mime';
...
+import { TERMS_TYPE_AND_DOCUMENT_ID_SEPARATOR, SNAPSHOT_ID_MARKER, COMMIT_MESSAGE_PREFIX } from './dataMapper.js';
 import Git from './git.js';
 
 import GitRepository from './index.js';
 
 const { expect } = chai;
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const RECORDER_PATH = path.resolve(__dirname, '../../../', config.get('recorder.versions.storage.git.path'));
+const RECORDER_PATH = path.resolve(__dirname, '../../../../..', config.get('recorder.versions.storage.git.path'));
```
MattiSG, line 19: I see no file rename. Why is this changed?

src/archivist/recorder/repositories/git/index.test.js
```diff
@@ -712,4 +712,188 @@ describe('GitRepository', () => {
...
+          it('returns the proper value', async () => {
+            expect((await subject.findById(commits.deprecatedRefilter.id)).isExtractOnly).to.be.true;
+          });
+
+          it('returns the proper value', async () => {
+            expect((await subject.findById(commits.deprecatedFirstRecord.id)).isExtractOnly).to.be.false;
+          });
+        });
+
+        context('record with current message', () => {
```
MattiSG, line 809: This context conflicts with the broader context: we are actually testing both the “current” and “deprecated” versions. I recommend to isolate the compatibility tests.

src/archivist/recorder/repositories/mongo/dataMapper.js
```diff
@@ -16,7 +16,7 @@ export function toPersistence(record) {
 }
 
 export function toDomain(mongoDocument) {
-  const { _id, serviceId, termsType, documentId, fetchDate, mimeType, isRefilter, isFirstRecord, snapshotIds } = mongoDocument;
+  const { _id, serviceId, termsType, documentId, fetchDate, mimeType, isExtractOnly, isRefilter, isFirstRecord, snapshotIds } = mongoDocument;
```
MattiSG, line 19: Shouldn't ESLint catch that one? 🤔 

```suggestion
  const { _id, serviceId, termsType, documentId, fetchDate, mimeType, isExtractOnly, isFirstRecord, snapshotIds } = mongoDocument;
```

src/archivist/recorder/repositories/git/index.test.js
```diff
@@ -712,4 +712,188 @@ describe('GitRepository', () => {
...
+      await subject.initialize();
+
+      /* eslint-disable no-await-in-loop */
+      for (const commit of Object.values(commits)) {
+        const { path: relativeFilePath, date, content, message } = commit;
+        const filePath = path.join(RECORDER_PATH, relativeFilePath);
+
+        await GitRepository.writeFile({ filePath, content });
+
+        const sha = await subject.commit({ filePath, message, date });
```
MattiSG, line 786: > I understand the only reason we now expose the `commit` method is to test it. Intrinsically, that's a bad reason, and our use of intermediary `exports.js` files to expose only subsets of features is not useful here because that method is a member of a class that is exposed or not as a whole.
>
> I would prefer finding an alternative way to test so we limit the API surface.

MattiSG (changes requested):
Congratulations for the implementation of the deprecation! I love how this made the code more future-proof by avoiding enumerating known keys :smiley: Keep up the good work!

CHANGELOG.md
Ndpnt: If think it's a breaking change in the way that newly created data will have different commit messages and if the user created a script that parses commit messages it won't work anymore (it's the same for mongo document attributes). The backwards compatibility is only ensure inside the engine but data structure will change.

src/archivist/recorder/repositories/git/index.test.js
MattiSG: Thank you for this clarification!

CHANGELOG.md
MattiSG: Yes, I agree with you. I am sorry, I was not clear. I meant that the current wording sounds contradictory. This detail that you add is important to add to the changelog.

src/archivist/recorder/repositories/mongo/dataMapper.js
Ndpnt: It's not an error, it is used for backwards compatibility. See line 29. (that's why Eslint did not cry)

src/archivist/recorder/repositories/git/index.test.js
Ndpnt: I totally agree and I tried to create a fixture repository with already saved commits but I had trouble to had them in the history without having to add a sub-module.

Ndpnt: I have an idea, I'm going to try something else.

MattiSG (changes requested):
Almost there! I'll handle the text improvement.

PR #1006, opened 2023-04-19
Clean up README

Ndpnt:


clementbiron (approved):

README.md
```diff
@@ -1,5 +1,3 @@
-_The document you are reading now is targeted at developers wanting to use or contribute to the engine of [Open Terms Archive](https://opentermsarchive.org). For a high-level overview of Open Terms Archive’s wider goals and processes, please read its [public homepage](https://opentermsarchive.org)._
```
clementbiron, line 1: I am not opposed to this change, but this piece of text does not seem obsolete to me

Ndpnt: What do you think @MattiSG?

MattiSG: That a pull request should always have a description, and that if that one had one we could understand what was your motivation with this change 🙂 

On the change itself, that “The document you are reading now” is pretty empty and thus it is borderline ridiculous to call I a “document”. But, on the other hand, that the second sentence linking to the public homepage is still relevant. I would be in favour of keeping it but moving it to the main paragraph.

MattiSG: (and that this should have been part of https://github.com/OpenTermsArchive/engine/pull/1005)

Ndpnt: > (and that this should have been part of #1005)

I agree, that's why I opened it right away.

Ndpnt: > On the change itself, that “The document you are reading now” is pretty empty and thus it is borderline ridiculous to call I a “document”. But, on the other hand, that the second sentence linking to the public homepage is still relevant. I would be in favour of keeping it but moving it to the main paragraph.

The link to the public web page is already in the "About" section of the engine, so I was thinking it was enough.

Ndpnt: I think this introduction paragraph is obsolete (and as you said a little bit ridiculous) as the document is almost empty. I think the idea is that users can quickly find out if the module is of interest to them or that they are quickly redirected to the main documentation rather than reading an unhelpful paragraph.

Ndpnt: > I don't find this super discoverable. Anyway, fine by me :)

I agree, I moved the second sentence to the main paragraph as you suggest.

PR #1010, opened 2023-04-24
Include collection name in dataset name

### LINKED ISSUE #954: Include collection name in dataset name
Currently, all generated datasets have the same name `dataset-${year}-${month}-${day}` and it does not make it easy to distinguish between files when using several of them.

We could add the instance name, for exemple `contrib-dataset-${year}-${month}-${day}` to improve that.

Ndpnt:
Fixes #954.


clementbiron (approved):
🥳

PR #1013, opened 2023-06-26
Fix issue creation on GitHub tracker

### LINKED ISSUE #1012: Cannot read properties of `undefined` (reading 'service')
As described in https://github.com/OpenTermsArchive/pga-declarations/issues/157: following an upgrade from v0.24 to v0.29, in the PGA collection, the module that updates issues on GitHub fails on every run with the following error:

```
unhandledRejection: Cannot read properties of undefined (reading 'service')
TypeError: Cannot read properties of undefined (reading 'service')
at Function.formatIssueTitleAndBody (…/pga/node_modules/@opentermsarchive/engine/src/tracker/index.js:183:41)
at Tracker.onInaccessibleContent (…/pga/node_modules/@opentermsarchive/engine/src/tracker/index.js:73:37)
at Archivist.emit (node:events:525:35)
```

This does not seem to block runs since [snapshots are still updated](https://github.com/OpenTermsArchive/pga-snapshots/tree/main). However, I see no new version since then, so it could be that versions are not properly extracted anymore (it could also just be coincidence). 

I had never seen that error before. The faulting line (`src/tracker/index.js:183`) has not been changed in the last three months. I see no related breaking changes in the [Changelog](https://github.com/OpenTermsArchive/engine/blob/02ae930c52122d8a893cb9fdf4288e6b4bfc0bb2/CHANGELOG.md).

It seems that one instance of `terms` cannot be destructured because it does not contain the `service` property.
I checked and every service declared in PGA has a `name`.

Ndpnt:
Fixes #1012, and will enable fixing https://github.com/OpenTermsArchive/pga-declarations/issues/157 when deployed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## Unreleased [patch]
+
+_Full changeset and discussions: [#1013](https://github.com/OpenTermsArchive/engine/pull/1013)._
+
+### Fixed
+
+- Fix issue creation on GitHub's tracker ([#1012](https://github.com/OpenTermsArchive/engine/issues/1012))
 
 ## 0.29.0 - 2023-06-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ _Full changeset and discussions: [#1013](https://github.com/OpenTermsArchive/eng
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
...
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## Unreleased [patch]
+
+_Full changeset and discussions: [#1013](https://github.com/OpenTermsArchive/engine/pull/1013)._
+
+### Fixed
+
+- Fix issue creation on GitHub's tracker ([#1012](https://github.com/OpenTermsArchive/engine/issues/1012))
```
MattiSG, line 11: The tracker does not belong to GitHub 🙂 
```suggestion
- Fix issue creation on GitHub tracker ([#1012](https://github.com/OpenTermsArchive/engine/issues/1012))
```

MattiSG (approved):

PR #1022, opened 2023-09-11
Added files to run OTA engine into Docker container

fabianospinelli:
Here the files needed to create a Docker container with the OTA engine

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,56 @@
+FROM ubuntu:latest
+
+WORKDIR /root
+
+RUN apt-get update \ 
+    && apt-get upgrade -y \
+    && apt install -y curl \
+    && apt install -y git \ 
+    && apt-get install -y chromium-browser
+
+RUN curl -fsSL https://deb.nodesource.com/setup_19.x | bash - \
+    && apt install -y nodejs
+
+# Add user so we don't need --no-sandbox.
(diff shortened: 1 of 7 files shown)


MattiSG:
Thanks @fabianospinelli for this contribution! We will need a bit of time to review it properly 🙂 

I can already mention that we will probably need the config file to be in the `config` directory, and that we don't want the engine to contain specific declarations. Are you embarking one for testing purposes? 🙂

fabianospinelli:
Hi! With the latest version of "npm" (v10.1.0) the build fails, so you need to fix a version (I use the version 9.8.0).
So you need to substitute "npm install -g npm@latest" with "npm install -g npm@9.8.0"

MattiSG:
Thanks for the heads up @fabianospinelli!
This has been fixed by @Ndpnt in our Ansible-based deployment system in [v0.0.11](https://github.com/OpenTermsArchive/deployment/blob/main/CHANGELOG.md#0011---2023-09-06).

Since we do not ourselves use Docker, and do not intend to, we expect it to be your responsibility to update the Docker files if they are to be made available publicly. Could you please update this PR accordingly? 🙂

I once more recommend the use of a tool such as [Packer](https://developer.hashicorp.com/packer/integrations/hashicorp/ansible/latest/components/provisioner/ansible) to provision Docker images through Ansible rather than duplicating the deployment instructions specifically for Docker.

PR #1026, opened 2023-10-18
Validate reporter configuration

Ndpnt:


clementbiron (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,18 @@
...
+_Full changeset and discussions: [#1026](https://github.com/OpenTermsArchive/engine/pull/1026)._
+
+### Added
+
+- Ensure `Reporter` configuration is defined before instantiating it
+- Ensure to add links to `versions` and `snapshots` in issues only if repositories are defined in configuration
+
+### Removed
+
+- No longer define default configuration for `Reporter` module
```
MattiSG, line 16: I believe this was a breaking change: if the reuser relied on the default configuration and did not define theirs, the behaviour would be significantly altered as they would not have any issues updates anymore.

config/default.json
```diff
@@ -54,15 +54,6 @@
...
   },
-  "reporter": {
-    "githubIssues": {
-      "repositories": {
-        "declarations": "OpenTermsArchive/sandbox-declarations",
-        "versions": "OpenTermsArchive/sandbox-versions",
-        "snapshots": "OpenTermsArchive/sandbox-snapshots"
-      }
-    }
-  },
```
MattiSG, line 65: Shouldn't this be moved to `config/test.json`?

src/reporter/index.js
```diff
@@ -132,8 +132,8 @@ If the source documents are accessible in a browser but fetching them always fai
 ### References
 
 - ${latestDeclarationLink}
-- ${latestVersionLink}
-- ${latestSnapshotsLink}
+${this.repositories.versions ? `- ${latestVersionLink}` : ''}
+${this.repositories.snapshots ? `- ${latestSnapshotsLink}` : ''}
```
MattiSG, line 136: I am not sure I understand: what would it mean that only declarations are defined, but not snapshots nor versions? Do we want to support such behaviour? 🤔

PR #1028, opened 2023-10-25
Define how to expose version through the collection metadata API

Ndpnt:
This PR is an experiment to create RFC that will lead to a decision record delivery.

The idea is to add comments right onto the document, much like you would with code. I will update the document for each consensus reached.

The deadline for this RFC is 27/11 end of day AoE (Anywhere on Earth).


decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+Note about the date:
+- It is not required that this date match exactly the fetch date of a version. As versions are fetched at periodic interval the version that will be returned will corresond to the version valid for this date range.
+- To have the latest version available, simple use the current date time as parameter.
+
+##### Returns
+
+- A raw markdown content when a version was found for this date
+- A HTTP 404 for a date that is anterior of the first available version
+- A HTTP 400 Bad Request, if the date is in the futur
```
MattiSG, line 45:
```suggestion
- HTTP 400 Bad Request with JSON content `{"error": "Requested version is in the future"}` if the date is in the future.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+##### Parameters
+
+| Parameter | Type   | Description            |
+| --------- | ------ | ---------------------- |
+| serviceId | URL-encoded string | The ID of the service |
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO date string | The date for which you want the version. This date should not be in the futur |
+
+Note about the date:
```
MattiSG, line 37:
```suggestion
Note about the date:
- A full date and time is required, and not a simple date (such as `2023-10-24`), to avoid ambiguities on days where a version changed, and timezone differences between client and server.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+- A HTTP 404 for a date that is anterior of the first available version
+- A HTTP 400 Bad Request, if the date is in the futur
+
+##### Example
+
+```
+GET /version/Open%20Terms%20Archive/Privacy%20Policy/2023-10-24T14%3A41%3A42.381Z.md
+```
+```markdown
+Privacy Policy
```
MattiSG, line 53:
```suggestion
HTTP 200
- - -
Privacy Policy
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Base URL
+
+`<collection host>/api/:version`
+
+#### Endpoint
+
+##### `GET /version/:serviceId/:termsType/:date`
+
+Return a JSON object containing of the version content valid for the given date along some metadata
```
MattiSG, line 73:
```suggestion
Returns a JSON object containing the version content valid for the given date along with metadata.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+##### Parameters
+
+| Parameter | Type   | Description            |
+| --------- | ------ | ---------------------- |
+| serviceId | URL-encoded string | The ID of the service |
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO date string | The date for which you want the version. This date should not be in the futur |
```
MattiSG, line 81: | date | URL-encoded ISO 8601 datetime string | The date and time for which the version is requested |

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+Note about the date:
+- It is not required that this date match exactly the fetch date of a version. As versions are fetched at periodic interval the version that will be returned will corresond to the version valid for this date range.
```
MattiSG, line 84:
```suggestion
- It is not required that this date matches exactly the fetch date of a version. As versions are fetched at a periodic interval, the version that will be returned will be the one that was applicable at the provided date.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+- To have the latest version available, simple use the current date time as parameter.
```
MattiSG, line 85:
```suggestion
- To get the latest version available, simply use the current date as parameter.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+
+##### Parameters
+
+| Parameter | Type   | Description            |
+| --------- | ------ | ---------------------- |
+| serviceId | URL-encoded string | The ID of the service |
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO date string | The date for which you want the version. This date should not be in the futur |
+
+Note about the date:
```
MattiSG, line 83:
```suggestion
Note about the date:
- A full date and time is required, and not a simple date (such as `2023-10-24`), to avoid ambiguities on days where a version changed, and timezone differences between client and server.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+- It is not required that this date match exactly the fetch date of a version. As versions are fetched at periodic interval the version that will be returned will corresond to the version valid for this date range.
+- To have the latest version available, simple use the current date time as parameter.
+
+##### Returns
+
+- A raw markdown content when a version was found for this date
+- A HTTP 404 for a date that is anterior of the first available version
+- A HTTP 400 Bad Request, if the date is in the futur
```
MattiSG, line 91:
```suggestion
- HTTP 400 Bad Request with JSON content `{"error": "Requested version is in the future"}` if the date is in the future.
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+| date | URL-encoded ISO date string | The date for which you want the version. This date should not be in the futur |
+
+Note about the date:
+- It is not required that this date match exactly the fetch date of a version. As versions are fetched at periodic interval the version that will be returned will corresond to the version valid for this date range.
+- To have the latest version available, simple use the current date time as parameter.
+
+##### Returns
+
+- A raw markdown content when a version was found for this date
+- A HTTP 404 for a date that is anterior of the first available version
```
MattiSG, line 90:
```suggestion
- HTTP 404 Not Found with JSON content `{"error": "No version found for date <requested-date>"}` for a date that is anterior to the first available version
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO date string | The date for which you want the version. This date should not be in the futur |
+
+Note about the date:
+- It is not required that this date match exactly the fetch date of a version. As versions are fetched at periodic interval the version that will be returned will corresond to the version valid for this date range.
+- To have the latest version available, simple use the current date time as parameter.
+
+##### Returns
+
+- A raw markdown content when a version was found for this date
```
MattiSG, line 89:
```suggestion
- HTTP 200 with a JSON object containing the Markdown content as body when a version is found for this date
```

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,113 @@
...
+  "content": "
+Privacy Policy
+==============
+
+Last updated: September 26, 2023
+
+This privacy policy explains how, why and under what conditions the Open Terms Archive (OTA) site collects personal information and how it is used. Our privacy policy will change over time. And of course, we also record the changes of [our own documents](https://github.com/OpenTermsArchive/demo-versions/tree/main/Open%20Terms%20Archive) 😉
+
+…
+"
```
MattiSG, line 111: This multi-line example is invalid. We would need to transform newlines to `\n`. The example should also provide an example with JSON data, to demonstrate how it would be encoded.

Ndpnt:
Solutions A and C are interesting, but I find that while it's still possible to pass metadata via HTTP headers, this approach is less common and less discoverable than solution B. 
Furthermore, in the case of solution C, the need to make a new request due to a redirect for each call does not seem desirable to me. Especially as the caching and transmission of metadata can be achieved in another way and are also doable in solution B.

So, I vote for the solution B as it aligns with the common and anticipated conventions of API responses. It provides a straightforward and highly extensible method for passing metadata in a standardized manner.

@MattiSG @clementbiron @mad @michielbdejong What do you think ?

_Note: The deadline elapsed a week ago, and I am starting the implementation this afternoon. The decision on the chosen solution will be confirmed tomorrow morning on the basis of the votes taken between now and then._

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,167 @@
+# Define how to expose version through the collection metadata API
```
MattiSG, line 1:
```suggestion
# Expose versions data through the collection API
```

MattiSG:
Thanks @Ndpnt!

Reading through solutions A, B and C, I came up with solution D that aims at keeping the best parts of each:

- A for ease of content access
- B for metadata extensibility
- C for performance capabilities

It has the added benefit of being possibly split in implementation in two parts, which are made clear in the proposal. Indeed, I feel that we have two orthogonal choices to make rather than 3 full-fledged solutions:

1. Do we redirect to leverage caching?
2. Do we serve JSON or Markdown?

…and that choice 2 can be delayed: for sure, there is relevance in JSON extensibility. Let's start with implementing that one and, while we're at it, defining in this RFC how to implement the Markdown version, without committing to implementing it at this stage as we first collect feedback from the JSOn version 🙂 


I thus vote for solution D, with only implementation of the JSON version of the endpoint!

Ndpnt:
Thanks @MattiSG for this relevant contribution!

I really like the idea of mixing solutions to keep the best parts of each 👍. At this stage I still find that using redirections to leverage caching is a premature optimization. And I prefer to wait until the need arises before choosing how to manage the cache. In addition, this caching optimisation also has the cost of introducing an extra round trip between client and server, which I still don't find desirable.

So for now I thus vote for solution D without the redirection mechanism. It means where solution D returns HTTP 302, I suggest to returns HTTP 200 with the version (either in markdown format or in JSON format according to the Content-Type)

clementbiron:
I'm in favour of solution E, which I feel is sufficient for the uses identified.

Well done Nico for your work!

Ndpnt:
Solution E received the most votes and will thus be selected.

Thanks for your participation :).

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,406 @@
...
+##### Parameters
+
+| Parameter | Type   | Description            |
+| --------- | ------ | ---------------------- |
+| serviceId | URL-encoded string | The ID of the service |
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO 8601 datetime string | The date and time for which the version is requested |
+
+Note about the date:
+- A full date and time is required, and not a simple date (such as `2023-10-24`), to avoid ambiguities on days where a version changed, and timezone differences between client and server.
```
michielbdejong, line 38: Maybe pin the timezone to UTC? So "A full UTC data and time"?

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Return the Markdown content of the version of the terms that was applicable at the given date
+
+##### Parameters
+
+| Parameter | Type   | Description            |
+| --------- | ------ | ---------------------- |
+| serviceId | URL-encoded string | The ID of the service |
+| termsType | URL-encoded string | The name of terms type |
+| date | URL-encoded ISO 8601 datetime string | The date and time for which the version is requested |
```
michielbdejong, line 35: And require that the timezone be UTC?

decision-records/0004-data-api.md
Ndpnt: If the timezone is defined as an offset from UTC (ex: `2023-12-06T12:34:56+02:00`) is still ISO 8601 valid. 
So, why forcing the timezone to be UTC?

michielbdejong:
Ah, looks like I came to this discussion just a day after it finished :)
OK, cool! You can take my suggestion to require UTC as the only allowed timezone or not, we can also simplify our code on our code to always use UTC even if your API supports more.

I think one way we could use this API is by setting the date far into the future, to get the latest known version, right? For instance:
```
GET /version/Open%20Terms%20Archive/Privacy%20Policy/9999-01-201T00%3A00%3A00.000Z`
```

And that would be cacheable! :)

Ndpnt:
> Ah, looks like I came to this discussion just a day after it finished :) OK, cool! You can take my suggestion to require UTC as the only allowed timezone or not, we can also simplify our code on our code to always use UTC even if your API supports more.

:+1:

> I think one way we could use this API is by setting the date far into the future, to get the latest known version, right? For instance:
> 
> ```
> GET /version/Open%20Terms%20Archive/Privacy%20Policy/9999-01-201T00%3A00%3A00.000Z`
> ```
> 
> And that would be cacheable! :)

We could also use a more elegant alias (`latest` for example) which would also allow caching:

```
GET /version/Open%20Terms%20Archive/Privacy%20Policy/latest
```

MattiSG:
> I think one way we could use this API is by setting the date far into the future, to get the latest known version, right?

In the current proposal, this would not work and would reply with a `416 Range Not Satisfiable`: the idea is that you're asking for the legal text that was applicable at the given date, and we can't guess what text will be applicable in the future.

I believe that this spontaneous usage, instead of using `Date.now()`, justifies further the relevance of a `/latest` endpoint.

decision-records/0004-data-api.md
```diff
@@ -0,0 +1,406 @@
...
+
+Note about the date:
+- A full date and time is required, and not a simple date (such as `2023-10-24`), to avoid ambiguities on days where a version changed, and timezone differences between client and server.
+- To get the latest version available, simply use the current date as parameter.
+
+##### Returns
+
+- HTTP 200 with the Markdown content as body when a version is applicable at this date
+- HTTP 404 Not Found with Markdown content `# Error\n\n_No version found for date <requested-date>_` for a date that is anterior to the first available version
+- HTTP 400 Bad Request with JSON content `# Error\n\n_Requested date <requested-date> is in the future, no version can exist there_` if the date is in the future.
```
MattiSG, line 380: There is an inconsistency with the Markdown / JSON content in this proposal here.
```suggestion
- HTTP 400 Bad Request with Markdown content `# Error\n\n_Requested date <requested-date> is in the future, no version can exist there_` if the date is in the future.
```

MattiSG, line 380: I suggest to use 416 instead of 400 for dates in the future, and to handle invalid dates with 400.

```suggestion
- HTTP 400 Bad Request with Markdown content `# Error\n\n_Requested date <requested-date> is not a valid ISO 8601 date time_` if the date is not a valid ISO 8601 date time.
- HTTP 416 Range Not Satisfiable with Markdown content `# Error\n\n_Requested date <requested-date> is in the future, no version can exist there_` if the date is in the future.
```

MattiSG (approved):
Thanks everyone! That works for me 🙂 👍 

My only concern is the lack of a `/latest` route for implementers, but since we introduced a 416 for far-future dates, the `/9999` trick suggested will not work and will thus enforce the use of `Date.now`. This might not be comfortable but is at least forward-compatible, and we could thus introduce the `/latest` route as an improvement after shipping the first version.

PR #1034, opened 2023-12-12
Provide a dedicated route in the Version API to obtain the latest version of some terms

Ndpnt:


src/api/routes/versions.js
```diff
@@ -97,14 +148,18 @@ router.get('/version/:serviceId/:termsType/:date', async (req, res) => {
   const version = await versionsRepository.findByDate(serviceId, termsType, requestedDate);
 
   if (!version) {
-    return res.status(404).json({ error: `No version found for date ${date}` });
+    return res.status(404).json({ error: `No version found the specified combination of service ID "${serviceId}", terms type "${termsType}" and date ${date}` });
```
MattiSG, line 151:
```suggestion
    return res.status(404).json({ error: `No version found for the specified combination of service ID "${serviceId}", terms type "${termsType}" and date ${date}` });
```

MattiSG (changes requested):
I like the refactor of tests and the opening of this route, but it is missing some important design choices. In particular, what happens if a service has been terminated, or if the chosen terms types have been phased out by the service? The `latest` version exists, but is not valid anymore. Do we really want to expose `latest`, or do we want to expose the `current` version?

If the service has been terminated (or the terms types do not exist anymore), do we reply with `404`, `416` or [`410 Gone`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410)?

Those questions should be discussed in a separate place than the PR implementing it.

Ndpnt:
Close this long-lived PR as it is not a priority right now.

PR #1044, opened 2024-02-08
Improve `test` and `changelog` workflows triggers

Ndpnt:
This change will :
- Avoid duplicate runs previously initiated by `push` and `pull_request` events
- Ensure test workflow is triggered on updates to pull requests originating from forks
- Enable workflows to be triggered manually to run CI tests on a branch without having to open a pull request


clementbiron (approved):
🚀

PR #1052, opened 2024-02-15
Improve optional modules loading

Ndpnt:


clementbiron (approved):
👍

src/index.js
```diff
@@ -38,8 +38,10 @@ export default async function track({ services, types, extractOnly, schedule })
     return;
   }
 
-  if (process.env.NODE_ENV === 'production') {
+  if (process.env.SENDINBLUE_API_KEY) {
     archivist.attach(new Notifier(archivist.services));
+  } else {
+    logger.warn('Environment variable "SENDINBLUE_API_KEY" was not found; the Notifier module will be ignored');
```
MattiSG, line 44: In logs, aim for descriptions that explain the impact to the end user: “the Notifier module will be ignored” is not a clear way to understand the consequences of an action and to assess whether it can be the source of an unexpected behaviour 🙂

PR #1059, opened 2024-03-07
Lint commit messages to enforce consistency

Ndpnt:
An example result can be seen here: https://github.com/Ndpnt/engine/actions/runs/8186516043/job/22385045460


.github/workflows/commit.yml
```diff
@@ -0,0 +1,15 @@
...
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: wagoid/commitlint-github-action@v5
```
MattiSG, line 15: Why use an action instead of running `npm ci && npm run commit:lint`? 🤔 This creates a difference between the local and CI workflows.

commitlint.config.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  rules: {
+    'header-case': [ 2, 'always', ['sentence-case']],
+    'header-max-length': [ 2, 'always', 50 ],
+    'header-full-stop': [ 2, 'never', '.' ],
+    'body-case': [ 2, 'always', ['sentence-case']],
+    'body-leading-blank': [ 2, 'always' ],
+    'body-max-line-length': [ 2, 'always', 72 ],
```
MattiSG, line 8: I am not a big fan of limiting line length for the body. For example, it forces to break URLs. What is the reasoning behind this limitation? It's been a long time since terminals were only 80 characters wide 😉

commitlint.config.js
```diff
@@ -0,0 +1,11 @@
...
+    'type-empty': [ 2, 'always' ],
```
MattiSG, line 9: > What does that mean? 🙂

commitlint.config.js
```diff
@@ -0,0 +1,11 @@
+export default {
```
MattiSG, line 1: > Cannot this be JSON?

package.json
```diff
@@ -30,6 +30,7 @@
     ".eslintrc.yaml"
   ],
   "scripts": {
+    "commit:lint": "commitlint --from=main --to=HEAD",
```
MattiSG, line 33: I find `commit` ambiguous in a script name, as it looks like a verb.
```suggestion
    "commit-messages:lint": "commitlint --from=main --to=HEAD",
```

MattiSG (changes requested):
These rules need also be documented in the CONTRIBUTING 🙂

MattiSG:
04d5f54 seems to be a test commit 😉

.github/workflows/commit.yml
Ndpnt: Yep, I tried quickly but it [did not work as expected](https://github.com/Ndpnt/engine/actions/runs/8186461328/job/22384893953) and as this feature was not planned I did not take time to investigate.

commitlint.config.js
Ndpnt: I can remove this rule, I implemented it because it is explicitly written [in our contributing file](https://github.com/OpenTermsArchive/engine/blob/main/CONTRIBUTING.md#commit-messages). If we decide to remove this rule, I'll ensure our contributing guidelines are updated accordingly.

MattiSG: Very good point! I'm personally in favor of not limiting, but that can wait until my next commit with frustratingly too short body lines 😉 I'll let you decide if you want to implement that here or not.

CONTRIBUTING.md
```diff
@@ -48,12 +48,12 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
 - [Capitalize the subject line](https://chris.beams.io/posts/git-commit/#capitalize).
 - [Do not end the subject line with a period](https://chris.beams.io/posts/git-commit/#end).
 - [Use the imperative mood in the subject line](https://chris.beams.io/posts/git-commit/#imperative).
-- [Wrap the body at 72 characters](https://chris.beams.io/posts/git-commit/#wrap-72).
 - [Use the body to explain what and why vs. how](https://chris.beams.io/posts/git-commit/#why-not-how).
 
 We add this additional rule:
 
 - Do not rely on GitHub issue reference numbers in commit messages, as we have no guarantee the host system and its autolinking will be stable in time. Make sure the context is self-explanatory. If an external reference is given, use its full URL.
+- Do not use [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/#summary) `<type>[optional scope]: <description>`, it will not be allowed, only describe what and why changes were made in one sentence.
```
MattiSG, line 56: I agree with that, but I feel a rationale why we don't allow it would be welcome 🙂

```suggestion
- Do not use the so-called “[conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)” format (`<type>[optional scope]: <description>`). There is no workflow in place for extracting metadata from such a format and it is thus needlessly cumbersome. 
```

Or maybe don't mention this at all since we already give a reference above?

CONTRIBUTING.md
```diff
@@ -48,12 +48,12 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
 - [Capitalize the subject line](https://chris.beams.io/posts/git-commit/#capitalize).
 - [Do not end the subject line with a period](https://chris.beams.io/posts/git-commit/#end).
 - [Use the imperative mood in the subject line](https://chris.beams.io/posts/git-commit/#imperative).
-- [Wrap the body at 72 characters](https://chris.beams.io/posts/git-commit/#wrap-72).
 - [Use the body to explain what and why vs. how](https://chris.beams.io/posts/git-commit/#why-not-how).
 
 We add this additional rule:
```
MattiSG, line 53: Should be "these rules" if we now have more than one :)

CONTRIBUTING.md
```diff
@@ -48,12 +48,12 @@ We strive to follow this [recommendation](https://chris.beams.io/posts/git-commi
 - [Capitalize the subject line](https://chris.beams.io/posts/git-commit/#capitalize).
 - [Do not end the subject line with a period](https://chris.beams.io/posts/git-commit/#end).
 - [Use the imperative mood in the subject line](https://chris.beams.io/posts/git-commit/#imperative).
-- [Wrap the body at 72 characters](https://chris.beams.io/posts/git-commit/#wrap-72).
```
MattiSG, line 51: We should explain why we don't follow this rule even though it is part of the reference we "strive to follow":

"As we might use URLs for references in commit messages, we do not wrap the body text to ensure links are clickable in Git user interfaces."

.github/workflows/commit.yml
MattiSG: I believe this happens because, by default, the `checkout` action checks out only the current commit, for performance reasons :) so the `main` reference does not exist.

I can't try that right now as I'm on my phone, but if I recall correctly there is an option to pass to `checkout` (maybe `depth: 0`?) to check out the whole tree.

I believe it is rather important to unify CI and local tests and avoid an Action dependency if we can :)

MattiSG (approved):
Looking forward to trying that out! As discussed synchronously, this might be useful or just as likely very frustrating for contributors. I am curious to see if this proves useful or if we remove it 😊 Thank you for suggesting that experiment! I already like that we could update our rules to reflect reality 👍

.github/workflows/commit.yml
```diff
@@ -0,0 +1,27 @@
...
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout "main" branch
```
MattiSG, line 14:
```suggestion
      - name: Checkout main branch history
```

MattiSG (approved):
👍

MattiSG:
@Ndpnt do you need anything from me here?

Ndpnt:
> @Ndpnt do you need anything from me here?

No, it's ok now, thanks!

PR #1063, opened 2024-03-18
Enhance error handling to minimize tracking interruptions from plugin errors

Ndpnt:
This errors handling improvement will significantly reduce tracking process interruptions caused by errors occurring in plugins. However, errors that occur in non _awaited_ asynchronous functions inside the plugin code will not be handled.
To have a completely safe environment where plugin errors cannot interrupt the terms tracking process, we could create an improved plugin system where each plugin is executed in its own sub-process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [minor]
+
+> Development of this release was supported by the [French Ministry for Foreign Affairs](https://www.diplomatie.gouv.fr/fr/politique-etrangere-de-la-france/diplomatie-numerique/) through its ministerial [State Startups incubator](https://beta.gouv.fr/startups/open-terms-archive.html) under the aegis of the Ambassador for Digital Affairs.
+
+### Changed
+
+- Improve errors handling mechanism to significantly reduce tracking process interruptions caused by errors occurring in plugins
+
 ## 1.0.0 - 2024-03-14
 
 _Full changeset and discussions: [#1061](https://github.com/OpenTermsArchive/engine/pull/1061)._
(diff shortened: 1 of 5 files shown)


src/archivist/index.test.js
```diff
@@ -245,6 +245,76 @@ describe('Archivist', function () {
...
+        await resetGitRepositories();
+      });
+
+      it('handles it properly and do not crash the whole execution', done => {
+        app.track({ services: [services[0]] }).then(() => {
+          if (error) {
+            return done(error);
+          }
+          done();
+        });
```
MattiSG, line 313: > Do we really need this complex setup with `unhandledRejection`? Couldn't we just use `should.not.throw`?

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [minor]
+
+> Development of this release was supported by the [French Ministry for Foreign Affairs](https://www.diplomatie.gouv.fr/fr/politique-etrangere-de-la-france/diplomatie-numerique/) through its ministerial [State Startups incubator](https://beta.gouv.fr/startups/open-terms-archive.html) under the aegis of the Ambassador for Digital Affairs.
+
+### Changed
+
+- Improve errors handling mechanism to significantly reduce tracking process interruptions caused by errors occurring in plugins
```
MattiSG, line 11: I would consider this an Added, especially since no reuser code adaptation is needed.

```suggestion
- Introduce `pluginError` error type to isolate plugin errors from the engine and increase robustness
```

MattiSG (approved):
Nice!

src/archivist/index.js
```diff
@@ -91,7 +88,13 @@ export default class Archivist extends events.EventEmitter {
       const handlerName = `on${event[0].toUpperCase()}${event.substring(1)}`;
 
       if (listener[handlerName]) {
-        this.on(event, listener[handlerName].bind(listener));
+        this.on(event, async (...params) => {
+          try {
+            await listener[handlerName](...params); // Prefer try...catch over .catch() for handling errors to take account both synchronous and asynchronous functions, as .catch() cannot be applied to synchronous functions
```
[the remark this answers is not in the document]
Ndpnt: In our convention, single line comment starts with an uppercase.

PR #1065, opened 2024-03-20
Ensure to reopen associated issue if a terms becomes inaccessible again

Ndpnt:
Additionally, include missing tests for the `GitHub` class in the `Reporter` module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All changes that impact users of this module are documented in this file, in the
 
 ### Fixed
 
-- Ensure to reopen associated issue if a terms becomes inaccessible again
+- Ensure to reopen associated issue if terms become inaccessible again
 
 ## 1.1.0 - 2024-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [patch]
+
+> Development of this release was supported by the [French Ministry for Foreign Affairs](https://www.diplomatie.gouv.fr/fr/politique-etrangere-de-la-france/diplomatie-numerique/) through its ministerial [State Startups incubator](https://beta.gouv.fr/startups/open-terms-archive.html) under the aegis of the Ambassador for Digital Affairs.
+
+### Fixed
+
+- Ensure to reopen associated issue if a terms becomes inaccessible again
+
 ## 1.1.0 - 2024-03-18
 
 _Full changeset and discussions: [#1063](https://github.com/OpenTermsArchive/engine/pull/1063)._
--- a/src/reporter/github.test.js
+++ b/src/reporter/github.test.js
@@ -1,0 +1,519 @@
+import { createRequire } from 'module';
+
+import { expect } from 'chai';
+import nock from 'nock';
+
+import GitHub from './github.js';
+
+const require = createRequire(import.meta.url);
+
+describe('GitHub', function () {
+  this.timeout(5000);
+
+  let github;
+  const MANAGED_LABELS = require('./labels.json');
+
+  before(() => {
+    github = new GitHub('owner/repo');
+  });
+
+  describe('#initialize', () => {
+    const scopes = [];
+
+    before(async () => {
+      const existingLabels = MANAGED_LABELS.slice(0, -2);
+
+      nock('https://api.github.com')
+        .get('/repos/owner/repo/labels')
+        .reply(200, existingLabels);
+
+      const missingLabels = MANAGED_LABELS.slice(-2);
+
+      for (const label of missingLabels) {
+        scopes.push(nock('https://api.github.com')
+          .post('/repos/owner/repo/labels', body => body.name === label.name)
+          .reply(200, label));
+      }
+
+      await github.initialize();
+    });
+
+    after(nock.cleanAll);
+
+    it('should create missing labels', async () => {
+      scopes.forEach(scope => expect(scope.isDone()).to.be.true);
+    });
+  });
+
+  describe('#getRepositoryLabels', () => {
+    let scope;
+    let result;
+    let labels;
+
+    before(async () => {
+      labels = [{ name: 'bug' }, { name: 'enhancement' }];
+
+      scope = nock('https://api.github.com')
+        .get('/repos/owner/repo/labels')
+        .reply(200, labels);
+
+      result = await github.getRepositoryLabels();
+    });
+
+    after(nock.cleanAll);
+
+    it('fetches repository labels', async () => {
+      expect(scope.isDone()).to.be.true;
+    });
+
+    it('returns the repository labels', async () => {
+      expect(result).to.deep.equal(labels);
+    });
+  });
+
+  describe('#createLabel', () => {
+    let scope;
+
+    before(async () => {
+      const label = { name: 'new_label', color: 'ffffff' };
+
+      scope = nock('https://api.github.com')
+        .post('/repos/owner/repo/labels', body => body.name === label.name)
+        .reply(200, label);
+
+      await github.createLabel(label);
+    });
+
+    after(nock.cleanAll);
+
+    it('creates the new label', async () => {
+      expect(scope.isDone()).to.be.true;
+    });
+  });
+
+  describe('#createIssue', () => {
+    let scope;
+    let createdIssue;
+    let result;
+
+    before(async () => {
+      const newIssue = {
+        title: 'New Issue',
+        description: 'Description of the new issue',
+        labels: ['bug'],
+      };
+
+      createdIssue = { number: 123, ...newIssue };
+
+      scope = nock('https://api.github.com')
+        .post('/repos/owner/repo/issues', body => body.title === newIssue.title && body.body === newIssue.description && body.labels[0] === newIssue.labels[0])
+        .reply(200, createdIssue);
+
+      result = await github.createIssue(newIssue);
+    });
+
+    after(nock.cleanAll);
+
+    it('creates the new issue', async () => {
+      expect(scope.isDone()).to.be.true;
+    });
+
+    it('returns the created issue', async () => {
+      expect(result).to.deep.equal(createdIssue);
+    });
+  });
+
+  describe('#setIssueLabels', () => {
+    let scope;
+
+    before(async () => {
+      const issueNumber = 123;
+      const labels = [ 'bug', 'enhancement' ];
+
+      scope = nock('https://api.github.com')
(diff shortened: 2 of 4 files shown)


clementbiron (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [patch]
+
+> Development of this release was supported by the [French Ministry for Foreign Affairs](https://www.diplomatie.gouv.fr/fr/politique-etrangere-de-la-france/diplomatie-numerique/) through its ministerial [State Startups incubator](https://beta.gouv.fr/startups/open-terms-archive.html) under the aegis of the Ambassador for Digital Affairs.
+
+### Fixed
+
+- Ensure to reopen associated issue if a terms becomes inaccessible again
```
MattiSG, line 11:
```suggestion
- Ensure to reopen associated issue if terms become inaccessible again
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+          .post('/repos/owner/repo/labels', body => body.name === label.name)
+          .reply(200, label));
+      }
+
+      await github.initialize();
+    });
+
+    after(nock.cleanAll);
+
+    it('should create missing labels', async () => {
```
MattiSG, line 43: > Why is this `async`?

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+
+      const missingLabels = MANAGED_LABELS.slice(-2);
+
+      for (const label of missingLabels) {
+        scopes.push(nock('https://api.github.com')
+          .post('/repos/owner/repo/labels', body => body.name === label.name)
+          .reply(200, label));
+      }
+
+      await github.initialize();
```
MattiSG, line 38: Since this is the tested method, shouldn't it be called in the test code rather than in the `before`? 🙂

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+    });
+  });
+
+  describe('#getRepositoryLabels', () => {
+    let scope;
+    let result;
+    let labels;
+
+    before(async () => {
+      labels = [{ name: 'bug' }, { name: 'enhancement' }];
```
MattiSG, line 54: > Initialise this at declaration with a `const`, not in the `before`.

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+
+      scope = nock('https://api.github.com')
+        .get('/repos/owner/repo/labels')
+        .reply(200, labels);
+
+      result = await github.getRepositoryLabels();
```
MattiSG, line 60: Since this is the tested method, shouldn't it be called in the test code rather than in the `before`? 🙂

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+
+    after(nock.cleanAll);
+
+    it('fetches repository labels', async () => {
+      expect(scope.isDone()).to.be.true;
+    });
+
+    it('returns the repository labels', async () => {
+      expect(result).to.deep.equal(labels);
+    });
```
MattiSG, line 71: > I'm not sure there is much value in splitting this test in two, but it increases maintenance (and production) costs.

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+    let scope;
+
+    before(async () => {
+      const label = { name: 'new_label', color: 'ffffff' };
+
+      scope = nock('https://api.github.com')
+        .post('/repos/owner/repo/labels', body => body.name === label.name)
+        .reply(200, label);
+
+      await github.createLabel(label);
```
MattiSG, line 84: Since this is the tested method, shouldn't it be called in the test code rather than in the `before`? 🙂

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+    });
+  });
+
+  describe('#createIssue', () => {
+    let scope;
+    let createdIssue;
+    let result;
+
+    before(async () => {
+      const newIssue = {
```
MattiSG, line 100:
```suggestion
      const ISSUE = {
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+        title: 'New Issue',
+        description: 'Description of the new issue',
+        labels: ['bug'],
+      };
+
+      createdIssue = { number: 123, ...newIssue };
+
+      scope = nock('https://api.github.com')
+        .post('/repos/owner/repo/issues', body => body.title === newIssue.title && body.body === newIssue.description && body.labels[0] === newIssue.labels[0])
```
MattiSG, line 109: > `body.body` is not very readable 😉 

```suggestion
        .post('/repos/owner/repo/issues', request => request.title === newIssue.title && request.body === newIssue.description && request.labels[0] === newIssue.labels[0])
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+        .reply(200, createdIssue);
+
+      result = await github.createIssue(newIssue);
```
MattiSG, line 112: Since this is the tested method, shouldn't it be called in the test code rather than in the before? 🙂

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+    });
+
+    after(nock.cleanAll);
+
+    it('creates the new issue', async () => {
+      expect(scope.isDone()).to.be.true;
+    });
+
+    it('returns the created issue', async () => {
+      expect(result).to.deep.equal(createdIssue);
```
MattiSG, line 122: > Merge into one test.

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+          .patch(`/repos/owner/repo/issues/${issue.number}`, { state: GitHub.ISSUE_STATE_CLOSED })
+          .reply(200);
+
+        await github.closeIssueWithCommentIfExists({ title: issue.title, comment: 'Closing comment' });
+      });
+
+      it('searches for the issue', () => {
+        expect(getIssuesScope.isDone()).to.be.true;
+      });
+
```
MattiSG, line 269: This is white box testing. This shouldn't be tested.
```suggestion
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
```
MattiSG, line 307: Whitebox. 
```suggestion
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+      it('does not attempt to add comment', () => {
```
MattiSG, line 308:
```suggestion
      it('does not add comment', () => {
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+        expect(addCommentScope.isDone()).to.be.false;
+      });
+
+      it('does not attempt to close the issue', () => {
+        expect(closeIssueScope.isDone()).to.be.false;
+      });
```
MattiSG, line 314: > Not sure how much it makes sense to test that. If we tried to close a closed issue, would that be a problem from a user perspective?

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+      before(async () => {
+        issue = { number: 123, title: 'Non-existent Issue', state: GitHub.ISSUE_STATE_CLOSED };
+
+        getIssuesScope = nock('https://api.github.com')
+          .get('/repos/owner/repo/issues')
+          .query(true)
+          .reply(200, []);
+
+        addCommentScope = nock('https://api.github.com')
+          .post(`/repos/owner/repo/issues/${issue.number}/comments`)
```
MattiSG, line 332: I don't understand: if this issue does not exist, why does it have a number and why should we reply with `200` to get its comments? 🤔

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+          .patch(`/repos/owner/repo/issues/${issue.number}`, { state: GitHub.ISSUE_STATE_CLOSED })
+          .reply(200);
+
+        await github.closeIssueWithCommentIfExists({ title: issue.title, comment: 'Closing comment' });
+      });
+
+      it('searches for the issue', () => {
+        expect(getIssuesScope.isDone()).to.be.true;
+      });
+
```
MattiSG, line 345: Whitebox.
```suggestion
```

src/reporter/github.test.js
```diff
@@ -0,0 +1,519 @@
...
+      it('does not attempt to add comment', () => {
+        expect(addCommentScope.isDone()).to.be.false;
+      });
+
+      it('does not attempt to close the issue', () => {
+        expect(closeIssueScope.isDone()).to.be.false;
+      });
+    });
+
+    context('when the issue does not exist', () => {
```
MattiSG, line 317: While I am convinced of the relevance of testing this case, I am not sure of the relevance of the suggested tests. What does it mean to test that we do not try to add a comment to nor close a non-existent issue? If anything, we should check that our code gracefully handles non-200 replies, but checking that we do not send back further requests does not seem relevant to me.

MattiSG (changes requested):
Congratulations for creating a sorely missing test suite for what has grown into an important part of the workflow! 👏 

Here are some suggestions that are valid for the whole suite, aiming at decreasing the code size:

1. Remove all white box testing (typically spy functions checking that a search was made on the list of issues, which is an implementation detail we shouldn't care about).
2. Move all tested functions from `before` to the `it` body.
3. It seems all tested cases have only 1 to 2 test functions (once suggestion 1 above is applied), mostly checking that an API call was made and sometimes checking for return value. I suggest merging these functions into 1 single test function, enabling replacing the repeated `after(nock.cleanall)` by a single `afterEach` call at the beginning of the test suite.
4. Use a shared fixture for issue data to avoid repeating the data setup.

src/reporter/github.test.js
Ndpnt: Not necessary. For example if we want to have multiple test on the result of a function, it makes sens for me to have this in the `before` and each `it` test something in particular.

src/reporter/github.test.js
Ndpnt: In fact, if the issue does not exist we still receive a 200 HTTP code with an empty result `[]`. If we try to add a comment or close the issue, we probably have an error in our code.

src/reporter/github.test.js
Ndpnt: It changes nothing from a user perspective, but for us it means that there is a mistake in our code.

src/reporter/github.test.js
Ndpnt: For me whitebox testing is not necessarily a bad practice event if I also prefer blackbox testing. And in this situation, it seems to me that this test is valuable. Let's discuss this IRL.

MattiSG (approved):

PR #1066, opened 2024-03-29
Fix validation of modified declarations

Ndpnt:
Rewrite the method used to identify modified services term types. It address two key issues:
- Fixed results when a new service is added: Previously, the method failed to detect newly added services, leading to CI tests consistently validating pull requests that introduced new services.
- Fixed results when filters are modified: Previously, the method failed to detect services related to updated filters, leading to CI tests consistently validating pull requests that only modified services filters.

Others improvements:
- Added tests
- Removed [`deep-diff`](https://github.com/flitbit/diff) dependency


scripts/declarations/utils/index.test.js
```diff
@@ -145,6 +145,8 @@ async function setupRepository() {
 
   declarationUtils = new DeclarationUtils(INSTANCE_PATH, 'main');
   await declarationUtils.git.init();
+  await declarationUtils.git.addConfig('user.name', 'Open Terms Archive Testing Bot');
```
clementbiron, line 148: It's not blocking but I find it a bit strange to name it as if it were a specific bot for the test.

clementbiron (approved):

scripts/declarations/utils/index.test.js
Ndpnt: I have to specify a user name and email otherwise it does not work, do you have a better suggestion?

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
+        const declaration = await this.getJSONFromFile(this.defaultBranch, `declarations/${serviceId}.json`);
 
-      if (!modifiedFilePath.endsWith('.json')) {
-        // Here we should compare AST of both files to detect on which function
-        // change has been made, and then find which terms type depends on this
-        // function.
-        // As this is a complicated process, we will just send back all terms types
-        const declaration = await this.getJSONFile(`declarations/${serviceId}.json`, this.defaultBranch);
+        // TODO: Implement AST comparison between original and modified "filters" files to detect changes in functions, enabling identification of terms whose types depend on modified functions.
+        // Due to the complexity involved, temporarily returning all term types.
```
MattiSG, line 50: Good rephrasing of the comment!
I am not sure of the relevance of the content however, no matter the form. Indeed:

1. This is as much a TODO as any other improvement in the codebase. We typically don't use this TODO / FIXME syntax in the code to keep track of potential improvements but use project tracking tools.
2. I am not sure we need “AST comparison”. Every JS function can be `toString`-ed and one possible implementation could thus be to compare the stringified content of every `filter` entry in the declaration and compare it to the previous version.

Since this comment seems to be “programmer’s side notes” rather than providing a relevant explanation for a design choice at line level, I am in favour of rephrasing further.
```suggestion
         // Considering how rarely filters are used, simply return all term types that could potentially be impacted to spare implementing a function change check
```

scripts/declarations/utils/index.test.js
clementbiron: Either `Open Terms Archive Bot` associated with `bot@opentermsarchive.org`

Or `Open Terms Archive Testing Bot` associated with `testing-bot@opentermsarchive.org`

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
 
-  async getModifiedServiceTermsTypes() {
-    const { modifiedFilePaths, modifiedServicesIds } = await this.getModifiedData();
+  async getModifiedServicesTermsTypes() {
+    const { modifiedFilePaths } = await this.getModifiedData();
     const servicesTermsTypes = {};
 
     await Promise.all(modifiedFilePaths.map(async modifiedFilePath => {
-      const serviceId = DeclarationUtils.filePathToServiceId(modifiedFilePath);
+      const serviceId = DeclarationUtils.getServiceIdFromFilePath(modifiedFilePath);
```
MattiSG, line 44: Extract the block in `map` to new function, it is too long to be declared this way in an iteration 🙂

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
 
-      if (!modifiedFilePath.endsWith('.json')) {
-        // Here we should compare AST of both files to detect on which function
-        // change has been made, and then find which terms type depends on this
-        // function.
-        // As this is a complicated process, we will just send back all terms types
-        const declaration = await this.getJSONFile(`declarations/${serviceId}.json`, this.defaultBranch);
+        // TODO: Implement AST comparison between original and modified "filters" files to detect changes in functions, enabling identification of terms whose types depend on modified functions.
+        // Due to the complexity involved, temporarily returning all term types.
+        servicesTermsTypes[serviceId] = Object.keys(declaration.documents);
```
MattiSG, line 51: > Since we `map`, why use an outside variable instead of using the return value of `map`?

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
-        } else if (path[0] == 'documents') {
-          acc.add(path[1]);
-        }
+      const originalServiceTermsTypes = Object.keys(originalService.documents);
+      const addedTermsTypes = modifiedServiceTermsTypes.filter(termsType => !originalServiceTermsTypes.includes(termsType));
 
-        return acc;
-      }, new Set());
+      if (addedTermsTypes.length) {
+        servicesTermsTypes[serviceId] = addedTermsTypes;
```
MattiSG, line 70: Wait, why do we assign instead of inserting? Couldn't there be both additions and changes?

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
+      }
+
+      const potentiallyModifiedTermsTypes = modifiedServiceTermsTypes.filter(termsType => originalServiceTermsTypes.includes(termsType));
```
MattiSG, line 73: We use two filters in a row to partition a space in two, this is inefficient and unclear.
Suggested alternative: something along the lines of

```
servicesTermsTypes[serviceId] = modifiedServiceTermsTypes.filter(termsType => {
   !originalServiceTermsTypes.includes(termsType)) ||
   JSON.stringify(originalService.documents[termsType]) != JSON.stringify(modifiedService.documents[termsType]);  // that comparison could quite well be extracted to function to encapsulate the implementation, be more understandable, and hint at future performance optimisation
}):
```

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
...
+
+      potentiallyModifiedTermsTypes.forEach(termsType => {
+        const areTermsModified = JSON.stringify(originalService.documents[termsType]) != JSON.stringify(modifiedService.documents[termsType]);
+
+        if (!areTermsModified) {
+          return;
+        }
 
-      servicesTermsTypes[serviceId] = Array.from(new Set([ ...servicesTermsTypes[serviceId] || [], ...modifiedTermsTypes ]));
+        servicesTermsTypes[serviceId] = [termsType].concat(servicesTermsTypes[serviceId] || []);
```
MattiSG, line 82: This is incredibly complex because there are too many data structure changes and variables with same or too similar names that shadow each other. I strongly recommend using the `return` values and extracting to functions to reduce the complexity.

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+import { expect } from 'chai';
+
+import DeclarationUtils from './index.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+let declarationUtils;
+
+const INSTANCE_PATH = path.resolve(__dirname, './test');
```
MattiSG, line 13:
```suggestion
const SUBJECT_INSTANCE_PATH = path.resolve(__dirname, './test');
```

or

```suggestion
const SUBJECT_PATH = path.resolve(__dirname, './test');
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+describe('DeclarationUtils', () => {
+  describe('#getModifiedServicesTermsTypes', () => {
+    before(async () => {
+      await loadFixtures();
+      await setupRepository();
+    });
+
+    after(() => fs.rm(INSTANCE_PATH, { recursive: true }));
+
+    context('when terms has been modified in an existing service', () => {
```
MattiSG, line 48:
```suggestion
    context('when an existing declaration has been modified', () => {
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+      before(() => commitChanges(COMMIT_PATHS.serviceA, FIXTURES.serviceATermsUpdated.content));
+      after(() => removeLatestCommit());
```
MattiSG, line 50:
```suggestion
      after(removeLatestCommit);
```

scripts/declarations/utils/index.js
```diff
@@ -33,48 +36,55 @@ export default class DeclarationUtils {
     return modifiedServicesIds;
   }
 
-  async getModifiedServiceTermsTypes() {
-    const { modifiedFilePaths, modifiedServicesIds } = await this.getModifiedData();
+  async getModifiedServicesTermsTypes() {
```
MattiSG, line 39: If, as I understand from the tests, this actually returns `{ services: ['serviceId' ], servicesTermsTypes: { serviceId: [ termsTypes] } }`, it should be named 
```suggestion
  async getModifiedServicesAndTermsTypes() {
```

(but also, why should it duplicate information that can be obtained with `Object.keys`??!)

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+      it('returns the service ID and the updated terms type', async () => {
+        expect(await declarationUtils.getModifiedServicesTermsTypes()).to.deep.equal({
+          services: ['ServiceA'],
+          servicesTermsTypes: { ServiceA: ['Terms of Service'] },
+        });
+      });
+    });
+
+    context('when terms has been added to an existing service', () => {
```
MattiSG, line 60:
```suggestion
    context('when new terms are declared in an existing declaration', () => {
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+      it('returns the service ID and the added terms type', async () => {
+        expect(await declarationUtils.getModifiedServicesTermsTypes()).to.deep.equal({
+          services: ['ServiceA'],
+          servicesTermsTypes: { ServiceA: ['Imprint'] },
+        });
+      });
+    });
+
+    context('when a whole service has been added', () => {
```
MattiSG, line 72:
```suggestion
    context('when a new declaration has been added', () => {
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+      it('returns the added service ID along with the associated terms type', async () => {
+        expect(await declarationUtils.getModifiedServicesTermsTypes()).to.deep.equal({
+          services: ['ServiceB'],
+          servicesTermsTypes: { ServiceB: ['Terms of Service'] },
+        });
+      });
+    });
+
+    context('when a service has been removed', () => {
```
MattiSG, line 84:
```suggestion
    context('when a declaration has been removed', () => {
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+      it('returns no services and no terms types', async () => {
+        expect(await declarationUtils.getModifiedServicesTermsTypes()).to.deep.equal({
+          services: [],
+          servicesTermsTypes: {},
+        });
+      });
+    });
+
+    context('when a terms has been removed from an existing service', () => {
```
MattiSG, line 99:
```suggestion
    context('when terms are removed from an existing declaration', () => {
```

scripts/declarations/utils/index.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+      it('returns no services and no terms types', async () => {
+        expect(await declarationUtils.getModifiedServicesTermsTypes()).to.deep.equal({
+          services: [],
+          servicesTermsTypes: {},
+        });
+      });
+    });
+
+    context('when there is a combination of a service updated and a service added', () => {
```
MattiSG, line 111:
```suggestion
    context('when there is a combination of an updated declaration and an added declaration', () => {
```

MattiSG (commented):
Congrats for the improvements, the fixes, the tests, and dropping a dependency! 👏 

I find the code unusually hard to understand, with very long functions, shadowing variables and inconsistent data structures.

scripts/declarations/utils/index.js
Ndpnt: I not changed the previous API

scripts/declarations/utils/index.js
Ndpnt: Nope it's not possible, you can't both declare a new service and have terms that changed for this service. If it's a new service you can only have all terms added. 
This cannot be an override of a previous value as there is an early exit inside the previous if statement.

scripts/declarations/utils/index.js
Ndpnt: I do not understand what you mean here

MattiSG: The whole anonymous function that is defined and called as the `map` parameter is huge. I believe it would help with readability if it was extracted into its own named function.

Ndpnt: With the code simplification, I no longer find it useful.

PR #1068, opened 2024-05-01
Fix validation of modified declarations when declaration history is modified

Ndpnt:


clementbiron (approved):

scripts/declarations/utils/index.js
```diff
@@ -43,6 +43,10 @@ export default class DeclarationUtils {
     await Promise.all(modifiedFilePaths.map(async modifiedFilePath => {
       const serviceId = DeclarationUtils.getServiceIdFromFilePath(modifiedFilePath);
 
+      if (modifiedFilePath.endsWith('.history.json')) {
+        return; // Assuming history modifications imply corresponding changes in the service declaration and that the analyse of which terms types of this service have changed will be done when analyzing the related declaration, no further action is required here
```
MattiSG, line 47:
```suggestion
        return; // Assuming history modifications imply corresponding changes in the service declaration and that the analysis of which terms types of this service have changed will be done when analysing the related declaration, no further action is required here
```

MattiSG (approved):
How could we let a `.only` pass review 🤦‍♀️ thanks for adding the ESlint check!

PR #1069, opened 2024-05-02
Consider extraction errors as expected operational errors without crashing the engine

Ndpnt:
### Changed

- Consider extraction errors as expected operational errors without crashing the engine

### Added

- Added label `empty content` to reported issues on GitHub when server returns empty content or when PDF content is unextractable
- Added label `invalid selector` to reported issues on GitHub when CSS selector is invalid

### Fixed

- Prevent engine crashes caused by invalid URLs in source document links
- Ensure the CLI option `--types` is not ignored
- Fix counts of terms when they are specified via the CLI option `--types`
- Fixed display issues for service ID and terms type associated with errors

---

Other technical improvements:
- Make `logger` silent in `test` env
- Also allow passing no params to the `track` function
- Ensure all errors from `fetch` function will bubble as `FetchDocumentError`
- Ensure all errors from `extract` function will bubble as `ExtractDocumentError`
- Minor improvements in logged messages


clementbiron (approved):
It sounds great and, well done!

src/archivist/extract/index.js
```diff
@@ -63,7 +69,7 @@ export async function extractFromHTML(sourceDocument) {
       });
       /* eslint-enable no-await-in-loop */
     } catch (error) {
-      throw new InaccessibleContentError(`The filter function "${filterFunction.name}" failed: ${error}`);
+      throw new Error(`The filter function "${filterFunction.name}" failed: ${error}`);
```
MattiSG, line 72: > I see the point of a higher-order try / catch to wrap the error, but I am not sure I understand why we decrease the specificity of the error here. What do we gain from changing to the generic `Error` type?

src/archivist/fetcher/index.js
```diff
@@ -28,9 +29,13 @@ export default async function fetch({
     waitForElementsTimeout = config.get('fetcher.waitForElementsTimeout'),
   } = {},
 }) {
-  if (executeClientScripts) {
-    return fetchFullDom(url, cssSelectors, { navigationTimeout, language, waitForElementsTimeout });
-  }
+  try {
+    if (executeClientScripts) {
+      return await fetchFullDom(url, cssSelectors, { navigationTimeout, language, waitForElementsTimeout });
```
MattiSG, line 34: With this change, the documentation is wrong. We are not returning a promise that will be fulfilled anymore, but the fulfilled promise. I am not sure of the performance impact. Is the higher-order try / catch worth that change? Shouldn't we instead chain a `.error` handler to the promise before returning it?

src/archivist/fetcher/index.test.js
```diff
@@ -148,18 +151,26 @@ describe('Fetcher', function () {
...
           expect(mimeType).to.equal('application/pdf');
         });
 
-        it('returns a blob with the file content', async () => {
+        it('returns a blob with the file content', () => {
           expect(content.equals(expectedPDFContent)).to.be.true;
         });
       });
+
+      context('when server responds with an emtpy content', () => {
```
MattiSG, line 167:
```suggestion
      context('when server responds with empty content', () => {
```

src/archivist/services/service.test.js
```diff
@@ -131,29 +132,49 @@ describe('Service', () => {
...
+    });
+
+    context('without any filter', () => {
+      it('returns the number of all terms types', () => {
+        expect(subject.getNumberOfTerms()).to.equal(2);
+      });
+    });
+
+    context('with a filter', () => {
+      it('returns the number of filtered terms types', () => {
```
MattiSG, line 175:
```suggestion
      it('returns the number of terms matching the provided terms types', () => {
```

src/reporter/labels.json
```diff
@@ -73,5 +78,10 @@
     "name": "location",
     "color": "FBCA04",
     "description": "Fetch location is outdated"
+  },
+  {
+    "name": "invalid selector",
+    "color": "FBCA04",
+    "description": "Selector is invalid"
```
MattiSG, line 85: It should be plural, as there are several selectors in a declaration.

> I propose a rephrasing that uses different words than “invalid” to increase the amount of information provided to the user in case “invalid” is unclear.

```suggestion
    "description": "Some selectors cannot be understood by the engine"
```

test/fixtures/termsNoText.pdf
MattiSG, line 1: > Suggested rename: `termsWithNoText` or `termsWithoutText`.

src/archivist/index.js
```diff
@@ -208,7 +232,7 @@ export default class Archivist extends events.EventEmitter {
     return record;
   }
 
-  async recordSnapshots(terms) {
```
MattiSG, line 211: How can this not be `async` when it returns a `Promise`?

src/archivist/index.js
```diff
@@ -182,12 +186,32 @@ export default class Archivist extends events.EventEmitter {
...
+          throw error;
+        }
+
+        extractDocumentErrors.push(error.message);
+      }
+    }));
+
+    if (extractDocumentErrors.length) {
+      throw new InaccessibleContentError(extractDocumentErrors);
+    }
```
MattiSG, line 205: This change is part of a commit named “Consider extraction errors as operational errors”, but I fail to understand exactly what we are adding here. As I understand it, we are now making `extract` potentially fail if any single document fails at extraction 😱 
We should probably have a sync call on that one 🙂

MattiSG: I had misread the variable names: this will not fail _tracking_ if any _terms_ fail to extract, but fail the _extraction_ if any _document_ fails to extract, which is right 🙂

src/archivist/index.js
MattiSG: Because we don't use `await` but write a Promise ourselves.

src/archivist/fetcher/fullDomFetcher.js
```diff
@@ -26,13 +24,13 @@ export default async function fetch(url, cssSelectors, config) {
     response = await page.goto(url, { waitUntil: 'networkidle0' });
 
     if (!response) {
-      throw new FetchDocumentError(`Response is empty when trying to fetch '${url}'`);
+      throw new Error(`Response is empty when trying to fetch '${url}'`);
```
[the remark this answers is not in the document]
MattiSG: Following sync conversation: we gain a slight decrease in programmer's necessary attention. I personally don't believe this is worth the churn nor the loss in precision, but I'm fine with it 😉

src/archivist/fetcher/index.js
MattiSG: I was wrong in my understanding of `return await` behaviour. This is still not blocking and won't change anything. Thanks @Ndpnt for having worked together on a test case, I learned something new today 😃

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+
+### Added
+
+- Added label `empty content` to reported issues on GitHub when server returns empty content or when PDF content is unextractable
+- Added label `invalid selector` to reported issues on GitHub when CSS selector is invalid
+
+### Fixed
+
+- Prevent engine crashes caused by invalid URLs in source document links
+- Ensure the CLI option `--types` is not ignored
```
MattiSG, line 21:
```suggestion
- Take into account the `--types` CLI option 
```

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+- Consider extraction errors as expected operational errors without crashing the engine
+
+### Added
+
+- Added label `empty content` to reported issues on GitHub when server returns empty content or when PDF content is unextractable
+- Added label `invalid selector` to reported issues on GitHub when CSS selector is invalid
+
+### Fixed
+
+- Prevent engine crashes caused by invalid URLs in source document links
```
MattiSG, line 20:
```suggestion
- Do not crash when source documents links point to invalid URLs
```

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+- Ensure the CLI option `--types` is not ignored
+- Fix counts of terms when they are specified via the CLI option `--types`
+- Fixed display issues for service ID and terms type associated with errors
```
MattiSG, line 23: Unify tense.
```suggestion
- Fix display issues for service ID and terms type associated with errors
```

> This is really not clear, I have to say 😅

src/archivist/extract/index.js
MattiSG: Following sync conversation: we gain a slight decrease in programmer's necessary attention. I personally don't believe this is worth the churn nor the loss in precision, but I'm fine with it 😉

MattiSG (approved):
Assuming typos and copywriting changes are taken into account, LGTM! 👏 
So many fixes and improvements, nice! There were a few riders though, such as the terms types filtering and count correction; it's always better to ship them separately 😉

PR #1070, opened 2024-05-08
Fix `ota lint` CLI command

Ndpnt:


MattiSG:
@Ndpnt why do we need to move the `no-only` plugin to the standard requirements? Are we exposing this to reusers? 🤔

Ndpnt:
> @Ndpnt why do we need to move the `no-only` plugin to the standard requirements? Are we exposing this to reusers? 🤔

Yes, through the `eslint` config used by the `ota lint` command

PR #1102, opened 2024-09-10
Update types to terms: types

madoleary:
We found that Archivist expects a variable by the name of 'terms' and not 'types' for the track function.


michielbdejong:
Right, this was already fixed in https://github.com/OpenTermsArchive/engine/commit/8c3525c779ad738c9415c0f3c9cc362a4f6b8a60

PR #1104, opened 2024-09-13
More skip options, fixes #1103

### LINKED ISSUE #1103: More skip options
IIUC the current default behaviour is:
* load from snapshots (for 'pre run')
* record version ('pre run')
* fetch source documents
* record snapshots
* load ('read back') from snapshots
* record version

And the `extractOnly` flag changes this to:
* load from snapshots
* record version

I would like to have the following behaviour in our instance:
* skip the pre run (IIUC this is only needed if the declaration or the software version changed, and we can do an explicit pre run in those situations)
* skip the read back
* maybe even skip the recoding of snapshots altogether.

So the behaviours I would like to have access to would be:
1) skipPreRun + skipReadBack:
* fetch source documents
* record snapshots
* record version

2) skipPreRun + skipSnapshots
* fetch source documents
* record version

I'll create a PR that adds the `skipPreRun`, `skipReadBack` and `skipSnapshots` options, OK?

michielbdejong:
This would require some tests before merging I guess, but let me know what you think!

### DIFF
--- a/bin/ota-track.js
+++ b/bin/ota-track.js
@@ -16,6 +16,9 @@ program
   .option('-s, --services [serviceId...]', 'service IDs of services to track')
   .option('-t, --types [termsType...]', 'terms types to track')
   .option('-e, --extract-only', 'extract versions from existing snapshots with latest declarations and engine, without recording new snapshots')
-  .option('--schedule', 'track automatically at a regular interval');
+  .option('--schedule', 'track automatically at a regular interval')
+  .option('--skipPreRun', 'skip the pre run')
+  .option('--skipReadBack', 'skip the read-back of snapshots')
+  .option('--skipSnapshots', 'skip both the writing and the read-back of the snapshots');
 
 track(program.parse(process.argv).opts());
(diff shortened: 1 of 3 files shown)


Ndpnt:
Hi @michiel,

Thanks for your suggestion, and apologies for the delayed response.

The options you suggest for skipping snapshots conflict with our second design principle, ["Do not require trust in maintainers"](https://docs.opentermsarchive.org/concepts/design-principles/#examples-of-consequential-choices-1), as snapshots are fundamental to ensuring trust in our system. For this reason, we do not plan to enable them.

Regarding the "prerun" step, rather than allowing it to be skipped, which might be tempting due to its time cost, we see an intermediate solution in optimizing it. Instead of running it at every engine start, we could trigger it only when necessary, such as when the engine, its dependencies, or the service declaration change, to minimize its impact on runtime.

This step is crucial to ensure that technical or declaration-related updates, which may affect extracted versions without being meaningful, are not mixed with actual content changes. Our current process already differentiates these in version history, and, for instance, the engine does not send notifications for purely technical updates.

While we don’t have a timeline due to other priorities, we believe improving the step’s efficiency is a better approach than making it optional.

I’ll close this PR for now. Thanks again for your suggestions and for contributing to improve the engine! 🙂

PR #1109, opened 2024-10-15
Prevent the engine from exiting on non-critical SMTP errors from the logger

Ndpnt:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [patch]
+
+> Development of this release was supported by the [French Ministry for Foreign Affairs](https://www.diplomatie.gouv.fr/fr/politique-etrangere-de-la-france/diplomatie-numerique/) through its ministerial [State Startups incubator](https://beta.gouv.fr/startups/open-terms-archive.html) under the aegis of the Ambassador for Digital Affairs.
+
+### Fixed
+
+- Prevent the engine from exiting on non-critical SMTP errors from the logger
```
MattiSG, line 11: Aren't the errors “critical” from the standpoint of the SMTP server? :)

```suggestion
- Prevent the engine from exiting on SMTP errors from the logger
```

src/logger/index.js
```diff
@@ -51,7 +62,6 @@ if (config.get('@opentermsarchive/engine.logger.sendMailOnError')) {
       ssl: true,
       timeout: 30 * 1000,
       formatter: args => args[Object.getOwnPropertySymbols(args)[1]], // Returns the full error message, the same visible in the console. It is referenced in the argument object with a Symbol of which we do not have the reference but we know it is the second one.
-      exitOnError: true,
```
MattiSG, line 54: Why didn't we simply implement the whole thing by removing this line? 🤔 😅

MattiSG (approved):
I am fine with the implementation, but I am concerned that we change the behavior significantly: we now exit on every logging error except SMTP ones, whereas before we exited only one SMTP logging errors. Why not simply never `exitOnError`? 😅

src/logger/index.js
Ndpnt: If we remove this line while keeping `rejectionHandlers` defined, Winston will catch all `uncaughtException` events without exiting the process, which is not the intended behavior. By default, both Node.js and Winston terminate the process when an `uncaughtException` occurs, as these are considered programmer errors. And this is the correct approach. As explained [here](https://github.com/OpenTermsArchive/engine/blob/main/CONTRIBUTING.md#handling-programmer-errors), crashing immediately is the best way to handle programmer errors.

Ndpnt:
> I am fine with the implementation, but I am concerned that we change the behavior significantly: we now exit on every logging error except SMTP ones, whereas before we exited only one SMTP logging errors. Why not simply never `exitOnError`? 😅

I agree that the previous code may have given the impression that we were exiting only on SMTP logging errors, but that wasn’t actually the case. Once exitOnError is set to true, the process will exit for any type of error, not just those from a specific transport. Also, I’m not sure this option is available at the transport level—it seems to be only available at the logger level. I believe the previous configuration at the transport level was a mistake.
So the behavior is kept the same!

PR #1113, opened 2024-10-23
Reduce GitHub API calls in issues reporter

Ndpnt:


src/reporter/github.js
```diff
@@ -53,6 +56,49 @@ export default class GitHub {
...
+        state: GitHub.ISSUE_STATE_ALL,
+        per_page: 100,
+      });
+
+      const onlyIssues = issues.filter(issue => !issue.pull_request); // Filter out pull requests since GitHub treats them as a special type of issue
+
+      onlyIssues.forEach(issue => {
+        const cachedIssue = this.issuesCache.get(issue.title);
+
+        if (!cachedIssue || new Date(issue.created_at) < new Date(cachedIssue.created_at)) { // Only cache the oldest issue if there are duplicates
```
MattiSG, line 72:
```suggestion
        if (!cachedIssue || new Date(issue.created_at) < new Date(cachedIssue.created_at)) { // Only work on the oldest issue if there are duplicates, in order to consolidate the longest history possible
```

src/reporter/github.js
```diff
@@ -53,6 +56,49 @@ export default class GitHub {
...
+      this.issuesCache.clear();
+      this.cacheLoadedPromise = this.loadAllIssues();
+      await this.cacheLoadedPromise;
+      logger.info('Issues cache refreshed successfully');
+    } catch (error) {
+      logger.error(`Failed to refresh issues cache: ${error.message}`);
+    }
+  }
+
+  async ensureCacheLoaded() {
```
MattiSG, line 95: Why not simply `await this.cacheLoadedPromise` when we call `ensureCacheLoaded()`? 🙂 Then we don't need to `await` it in `refreshIssuesCache()` and we don't need this method.

src/reporter/github.js
```diff
@@ -69,40 +115,42 @@ export default class GitHub {
...
-      ...this.commonParams,
-      per_page: 100,
-      ...searchParams,
-    }, response => response.data);
-
-    const [issue] = issues.filter(item => item.title === title); // Since only one is expected, use the first one
-
-    return issue;
+  getIssue(title) {
+    return this.issuesCache.get(title) || null;
```
MattiSG, line 148: Why isn't `undefined` acceptable as a return value if the issue is not found in the cache? 🙂 
```suggestion
    return this.issuesCache.get(title);
```

src/reporter/github.js
```diff
@@ -114,25 +162,30 @@ export default class GitHub {
...
 
-      await this.addCommentToIssue({ issue: openedIssue, comment });
-      logger.info(`Added comment to issue #${openedIssue.number}: ${openedIssue.html_url}`);
+      await this.addCommentToIssue({ issue, comment });
+
+      const updatedIssue = await this.updateIssue(issue, { state: GitHub.ISSUE_STATE_CLOSED });
 
-      await this.updateIssue(openedIssue, { state: GitHub.ISSUE_STATE_CLOSED });
-      logger.info(`Closed issue #${openedIssue.number}: ${openedIssue.html_url}`);
+      this.issuesCache.set(updatedIssue.title, updatedIssue);
```
MattiSG, line 177: Shouldn't the cache update be done in `updateIssue`? I find it very surprising that we have to handle that in the caller.

src/reporter/github.js
```diff
@@ -114,25 +162,30 @@ export default class GitHub {
 
   async closeIssueWithCommentIfExists({ title, comment }) {
     try {
```
MattiSG, line 164: I feel that only the calls to GitHub should be wrapped in a `try/catch`, as they are the only predictable points of failure.

src/reporter/github.js
```diff
@@ -53,6 +56,49 @@ export default class GitHub {
...
+        }
+      });
+
+      logger.info(`Cached ${onlyIssues.length} issues from the repository`);
+    } catch (error) {
+      logger.error(`Failed to load issues: ${error.message}`);
+    }
+  }
+
+  async refreshIssuesCache() {
```
MattiSG, line 83: I would find it more straightforward if there was some `issues` getter that memoizes the promise and populates the cache if needed, and a `clearCache` method that does only that.

src/reporter/github.js
```diff
@@ -69,40 +115,42 @@ export default class GitHub {
...
-    const issues = await this.octokit.paginate('GET /repos/{owner}/{repo}/issues', {
-      ...this.commonParams,
-      per_page: 100,
-      ...searchParams,
-    }, response => response.data);
-
-    const [issue] = issues.filter(item => item.title === title); // Since only one is expected, use the first one
-
-    return issue;
+  getIssue(title) {
```
MattiSG, line 147: What are the benefits of wrapping `this.issues.get(title)` into `this.getIssue(title)`? 🙂

src/reporter/github.js
```diff
@@ -141,23 +194,22 @@ export default class GitHub {
...
       if (issue.state !== GitHub.ISSUE_STATE_CLOSED && managedLabel?.name === label) {
         return;
       }
 
-      await this.updateIssue(issue, {
-        state: GitHub.ISSUE_STATE_OPEN,
-        labels: [ label, ...labelsNotManagedToKeep ],
-      });
-      logger.info(`Updated issue #${issue.number}: ${issue.html_url}`);
+      const updatedIssue = await this.updateIssue(issue, { state: GitHub.ISSUE_STATE_OPEN, labels: [ label, ...labelsNotManagedToKeep ].filter(label => label) });
```
MattiSG, line 205: Why do we add the `filter` here? Do we have cases where some labels are empty strings?

> I found it more readable over multiple lines.

src/reporter/github.js
```diff
@@ -141,23 +194,22 @@ export default class GitHub {
...
+
       await this.addCommentToIssue({ issue, comment: description });
 
-      logger.info(`Added comment to issue #${issue.number}: ${issue.html_url}`);
+      this.issuesCache.set(updatedIssue.title, updatedIssue);
```
MattiSG, line 209: I feel that the cache update should take place in `updateIssue`, not in the caller.

src/reporter/github.test.js
```diff
@@ -71,77 +78,81 @@ describe('GitHub', function () {
...
+      const scope = nock('https://api.github.com')
         .post('/repos/owner/repo/labels', body => body.name === LABEL.name)
         .reply(200, LABEL);
 
       await github.createLabel(LABEL);
+      expect(scope.isDone()).to.be.true;
     });
 
-    after(nock.cleanAll);
+    it('throws error when creating label fails', async () => {
```
MattiSG, line 94:
```suggestion
    it('throws an error when creating a label fails', async () => {
```

src/reporter/github.test.js
```diff
@@ -163,35 +174,33 @@ describe('GitHub', function () {
     it('adds the comment to the issue', () => {
       expect(scope.isDone()).to.be.true;
     });
+
+    it('throws error when adding comment fails', async () => {
```
MattiSG, line 178:
```suggestion
    it('throws an error when adding a comment fails', async () => {
```

src/reporter/github.test.js
```diff
@@ -303,48 +288,38 @@ describe('GitHub', function () {
...
             .reply(200);
 
-          await github.createOrUpdateIssue(ISSUE);
+          await github.createOrUpdateIssue({ title: EXISTING_CLOSED_ISSUE.title, description: EXISTING_CLOSED_ISSUE.description, label: 'location' });
         });
 
-        it('reopens the issue and its labels', () => {
+        after(() => {
+          github.issuesCache.delete(EXISTING_CLOSED_ISSUE);
+          github.issuesCache.set(EXISTING_CLOSED_ISSUE.title, EXISTING_CLOSED_ISSUE);
```
MattiSG, line 319: I don't understand these: we delete an entry from the cache and then set it back in? 🤔

src/reporter/github.test.js
```diff
@@ -354,40 +329,63 @@ describe('GitHub', function () {
...
+          after(() => {
+            github.issuesCache.delete(EXISTING_OPEN_ISSUE.title);
+            github.issuesCache.set(EXISTING_OPEN_ISSUE.title, EXISTING_OPEN_ISSUE);
+          });
+
+          it("updates the issue's labels", () => {
+            expect(updateIssueScope.isDone()).to.be.true;
+          });
+
+          it('adds comment to the issue', () => {
```
MattiSG, line 357:
```suggestion
          it('adds a comment to the issue', () => {
```

src/reporter/github.test.js
```diff
@@ -354,40 +329,63 @@ describe('GitHub', function () {
...
+            expect(addCommentScope.isDone()).to.be.true;
+          });
         });
-
-        it('adds comment to the issue', () => {
-          expect(addCommentScope.isDone()).to.be.true;
+        context('when the reason unchanged', () => {
```
MattiSG, line 361:
```suggestion
        context('when the reason did not change', () => {
```

src/reporter/github.test.js
```diff
@@ -354,40 +329,63 @@ describe('GitHub', function () {
...
+
+            await github.createOrUpdateIssue({ title: EXISTING_OPEN_ISSUE.title, description: EXISTING_OPEN_ISSUE.description, label: EXISTING_OPEN_ISSUE.labels[0].name });
+          });
+
+          after(() => {
+            github.issuesCache.delete(EXISTING_OPEN_ISSUE.title);
+            github.issuesCache.set(EXISTING_OPEN_ISSUE.title, EXISTING_OPEN_ISSUE);
+          });
+
+          it("does not attempt to updates the issue's labels", () => {
```
MattiSG, line 382: Inconsistent quotes 😉 

```suggestion
          it('does not attempt to updates the issue’s labels', () => {
```

src/reporter/github.test.js
```diff
@@ -354,40 +329,63 @@ describe('GitHub', function () {
...
+            expect(updateIssueScope.isDone()).to.be.false;
+          });
+
+          it('does not attempt to add comment to the issue', () => {
```
MattiSG, line 386:
```suggestion
          it('does not attempt to add any comment to the issue', () => {
```

src/reporter/index.js
```diff
@@ -49,6 +49,10 @@ export default class Reporter {
     return this.github.initialize();
   }
 
+  onTrackingStarted() {
+    return this.github.refreshIssuesCache();
```
MattiSG, line 53: Why be eager? Let's just wait until we actually need the cache 🙂

MattiSG (changes requested):
Thanks for this work!
The caching mechanism is a bit too eager from my perspective: I believe we would make it significantly simpler by simply having a getter that populates when it is requested for the first time, and better encapsulating its updates 🙂

src/reporter/github.js
Ndpnt: You're absolutely right!

src/reporter/github.js
Ndpnt: I had it in tests but not sure if it can happen in real life. I will double check.

src/reporter/github.js
Ndpnt: Yes, of course. Fatigue...

src/reporter/github.test.js
Ndpnt: Yes because this entry in the cache was modified by the test

src/reporter/github.js
Ndpnt: And it was already done

src/reporter/github.js
Ndpnt: The behavior will almost be the same (as plugin errors won't stop the process) and if we do that we need three try catch in this function. So I'm not in favor of doing this

MattiSG (approved):
Looks really good, thanks!!

PR #1139, opened 2025-02-19
Update metadata specs

Ndpnt:


MattiSG (approved):

src/collection-api/routes/metadata.js
```diff
@@ -20,107 +20,146 @@ const PACKAGE_JSON_PATH = '../../../package.json';
...
+ *             This data is maintained and analysed by the [Platform Governance Archive](https://www.platformgovernancearchive.org/) at the Universität Bremen's [Center for Media Communication and Information Research (ZeMKI)](https://www.uni-bremen.de/zemki).
+ *
+ *             This initiative offers researchers, journalists and citizens the tools to analyze how platforms structure and regulate communication and interaction in our societies.
+ *
+ *             It also aims to promote greater transparency and accountability of these powerful digital services.
  *         totalTerms:
  *           type: integer
- *           description: Total number of terms tracked in the collection
+ *           description: Total number of terms tracked in the collection.
+ *           x-ota-generated: true
```
MattiSG, line 70: > So where is that tag used to filter out the view? 🤔

src/collection-api/routes/metadata.js
```diff
@@ -20,107 +20,146 @@ const PACKAGE_JSON_PATH = '../../../package.json';
...
+ *         - name
+ *         - tagline
+ *         - languages
+ *         - jurisdictions
  *       properties:
  *         id:
  *           type: string
- *           description: Unique identifier of the collection
+ *           description: Unique identifier derived from name (acronyms, dash-separated).
+ *           example: pga
```
MattiSG, line 33: Why do we take `pga` as an example, which is a third-party that we don't control, instead of `demo` that is here to that end?

Ndpnt: I've simply copied the examples from our current documentation, but you're right.

PR #1144, opened 2025-04-03
Move fullDomFetcher to Playwright

LVerneyEC:
Hi,

Here is a proposal for a rewriting of the full DOM fetcher, moving it to Playwright instead of Puppeteer.

This edited browser also has support for HTTP/HTTPS proxy (e.g. corporate proxy) and behavior can be adjusted by two environment variables:
* `PLAYWRIGHT_NO_SANDBOX` to disable all the sandboxing in Chrome (required for running in Docker, depending on the Docker setup).
* `PLAYWRIGHT_NO_HEADLESS` to run it in headful mode (sometimes useful for debugging purposes)

This is using [patchright](https://github.com/Kaliiiiiiiiii-Vinyzu/patchright) wrapper around Playwright, which adds several patches for obvious Playwright detection mechanisms. Similar to the previously used `puppeteer-extra-plugin-stealth`.

Best,


LVerneyEC:
I noticed the tests are failing due to linting and commit/changelog issues. I'll fix these, but happy to have a first high-level review first to ensure this is useful and worth merging and fix everything at once afterwards :)

MattiSG:
Thanks @LVerneyEC for this contribution! Fully agree with a first high-level overview before ironing out details  :)
The intervention seems minimal. Do you have examples of cases that were blocked with the previous implementation and are unblocked with that switch? 🙂

LVerneyEC:
> Do you have examples of cases that were blocked with the previous implementation and are unblocked with that switch? 🙂

Not so much. I have another PR to come for the htmlOnlyFetcher, for which this increases widely coverage.

Here, the main benefit is to move away from `puppeteer-extra-stealth` which is unmaintained for a couple years: https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth. 

Also, more high-level updates such as supporting corporate proxies and offering the ability to run headful for debugging purposes.

LVerneyEC:
Would you have more details the benchmark and the results? I am a bit surprised about the 403 and selector errors, since it does not really match my experience so far.

Ndpnt:
Hi @LVerneyEC,
Sorry for the late reply, I was on holiday and when I got back we had a seminar.

For the benchmark, I used the [PGA collection](https://github.com/OpenTermsArchive/pga-declarations) as it includes many VLOPs and for which we have many blocking issues.

I ran the engine five times consecutively using version 5.0.3 with Puppeteer, and another five times using the version from this PR with Playwright. All runs were executed on a server hosted on OVH Horizon.

You can find the output for each run [here](https://github.com/user-attachments/files/19955582/Benchmark.outputs.zip).

Since your experience seems different, could you share a bit more about your own results and setup? It would be great to compare and understand the differences.

Ndpnt:
Hi @LVerneyEC,

I've conducted additional testing across 5 collections, including yours, comparing Puppeteer, Playwright/Patchright, and [rebrowser](https://github.com/rebrowser/rebrowser-patches). 

Here are my results:
- While Playwright/Patchright successfully tracked some terms that Puppeteer failed on (like the Shein and Temu documents you mentioned), Puppeteer performed better overall across most terms.
- I identified that Puppeteer's failures were primarily due to configuration issues that caused navigation timeout. After fixing this, Puppeteer consistently outperformed both Playwright/Patchright and rebrowser.

In addition to the [fix of Puppeteer configuration](https://github.com/OpenTermsArchive/engine/blob/main/CHANGELOG.md#510---2025-05-12), I've implemented two improvements to enhance tracking reliability:
- [Automatic fallback to headless browser](https://github.com/OpenTermsArchive/engine/blob/main/CHANGELOG.md#530---2025-05-14) when bot detection is encountered (eliminating the need for `executeClientScripts` workaround to bypass bot protection)
- [Automatic retry mechanism](https://github.com/OpenTermsArchive/engine/blob/main/CHANGELOG.md#520---2025-05-13) for transient errors (network/server issues)

Can you test these improvements with your collection and give me a feedback?

LVerneyEC:
Hi,

Many thanks for the extra details. I did some extensive testing and comparison on my end, comparing with the latest baseline in `main` as of today.

First, **the current Puppeteer implementation cannot use an authenticating proxy**. I think it correctly grabs `http_proxy` environment variable, but it cannot handle the credentials (env variable of the form `http_proxy=http://username:password@host:port`). This would require some tuning (see https://github.com/Decodo/Puppeteer/blob/master/puppeteer.js#L8-L15), similar to the proxy parsing in this PR. If sticking to Puppeteer, this should be backported.


Then, I did a full comparison on our dataset of declarations (https://code.europa.eu/dsa/terms-and-conditions-database/vlops-and-vloses/vlop-vlose-declarations). Apart from a few transient and blinking terms (which are quite similar for both browsers), the striking difference is:
* `Amazon Store - Global Store terms and conditions` was only working with your Puppeteer implementation (HTTP code 403 otherwise, systematically working in one case / systematically failing in the other, over a few tries).
* `TikTok - Commercial Terms` was only working with our Playwright implementation (no match of the selector - getting a captcha page, systematically working in one case / systematically failing in the other, over a few tries).
* `Shein - Seller Agreement` is never working, no matter the browser.
* A few other platforms are blinking, but this is not statistically relevant and this is similar under both situations.


In light of this, **I would propose to strengthen and expand your latest retry strategy as such**:
* If scraping without client scripts and getting a failure, then retry with fullDomFetcher (this is your latest addition)
* If scraping with client scripts and getting a failure, then retry once (or a configurable number of times) with fullDomFetcher (just a basic retry for transient issues)

Beyond the fact that there might be transient issues, the reason for this is that depending on your infrastructure/proxy solution you might end up doing the second scraping try with a different IP and therefore augmenting your success rate.



Given the results on my testing set (no clear winner, 1 vs 1 failures), I'm wondering whether it would make sense to keep both browsers and either have it configurable in the `executeClientScripts` (e.g. `true` == Puppeteer, `false` == `htmlOnlyFetcher`, "playwright" == playwright codebase) or use it as an escalation when retrying a failed term?

Best

Ndpnt:
Hi @LVerneyEC,

Thanks for the testing and feedback.

Regarding the issue you mentioned with `TikTok - Commercial Terms` not working under Puppeteer: I dug into it and found that the problem can be fixed by changing the `waitUntil` option to consider page loaded to either `networkidle0` or `networkidle2`. However, reverting to that approach will led to navigation timeouts on other terms.
To avoid that, I [implemented a solution](https://github.com/OpenTermsArchive/engine/pull/1159) to explicitly waiting for expected elements to be present and non-empty on the page. 
With this update, the `TikTok - Commercial Terms` page now loads successfully with Puppeteer both on my local machine and on our OVH experimentation server.

While adding a fallback to Playwright when Puppeteer fails could be a practical solution, it would significantly increase code complexity and maintenance overhead. For this reason, I prefer to avoid it unless it's absolutely necessary and there's no reliable workaround within the existing Puppeteer setup.

Regarding support for authenticating proxies, would you be open to proposing an implementation in a separate pull request?

Ndpnt:
And I forgot, but indeed, it also seems like a good idea to expand the retry strategy to include failures that occurred with client script enabled since we’ve both seen transient blocking errors in that scenario.

Ndpnt:
Hi @LVerneyEC ,

Just following up as we haven’t received a reply, and the issues you were facing with Puppeteer have been addressed.
Thanks again for contributing to improve the tracking success rate. All improvements are available since engine [version 5.6.0](https://github.com/OpenTermsArchive/engine/blob/main/CHANGELOG.md#560---2025-06-19).
If you're interested in adding support for authenticated proxies, we'd be happy to review and merge a PR for that.

Closing this for now, but feel free to reopen if you run into other limitations that could be addressed with Playwright.

PR #1153, opened 2025-05-16
Support archiving plaintext files/markdown files

LVerneyEC:
Hi,

This adds the possibility to archive raw content serve over HTTP (plaintext or markdown). This would typically cover archiving of content served through https://raw.githubusercontent.com.

Best,

### DIFF
--- a/src/archivist/extract/index.js
+++ b/src/archivist/extract/index.js
@@ -35,6 +35,9 @@ export default async function extract(sourceDocument) {
     if (sourceDocument.mimeType == mime.getType('pdf')) {
       return await extractFromPDF(sourceDocument);
     }
+    if (sourceDocument.mimeType == mime.getType('txt') || sourceDocument.mimeType == mime.getType('md')) {
+      return await extractFromMd(sourceDocument);
+    }
 
     return await extractFromHTML(sourceDocument);
   } catch (error) {
@@ -121,6 +124,14 @@ export async function extractFromPDF({ location, content: pdfBuffer }) {
 
   if (!markdownContent) {
     throw new Error(`The PDF file at '${location}' contains no text, it might contain scanned images of text instead of actual text`);
+  }
+
+  return markdownContent;
+}
+
+export async function extractFromMd({ location, content: markdownContent }) {
+  if (!markdownContent) {
+    throw new Error(`The markdown file at '${location}' contains no text`);
   }
 
   return markdownContent;


MattiSG:
Hi @LVerneyEC!

Thank you for this suggestion 🙂 
Can you please provide a few examples of contractual documents that currently cannot be tracked and that would become trackable with this changeset?

Thank you!

MattiSG:
Thanks for clarifying the intention @LVerneyEC and for this suggestion!

Considering that adding a feature to the engine means guaranteeing its maintenance, we want to make sure that every additional feature aligns with Open Terms Archive’s [Design Principles](https://docs.opentermsarchive.org/concepts/design-principles/), that there are clear use cases associated with each of them, and that software quality is ensured 🙂 

In the current case, we do appreciate the upcoming relevance of prompts, and would add software licenses as potential cases as well. We will request the following elements before proceeding with merging:

- [ ] At least five examples of contractual documents that are provided to end users in the newly supported format. This could be obtained by a simple [online search](https://www.google.com/search?rls=en&q=terms+of+use+filetype%3Atxt&ie=UTF-8&oe=UTF-8) (ensuring that the TXT files are indeed the ones that are shown to the end users, in accordance with [principle 3](https://docs.opentermsarchive.org/concepts/design-principles/#3-obtain-documents-like-a-user-would)) for existing terms types.
  - If no such cases currently exist, seeing the provided example, [new terms types](https://github.com/OpenTermsArchive/terms-types/blob/main/CONTRIBUTING.md#add-new-terms-types) such as “Source Code License” or “System Prompt” could be added, that would then support listing examples.
- [ ] Automated tests are added for each new supported format, to ensure the stability of the feature over time. You can liaise with @Ndpnt for test design.

For the provided example (grok prompts stored on GitHub), while the topic is definitely exciting, the relevance of Open Terms Archive for tracking seems a bit remote, as one could simply clone the repository for history preservation, and directly [subscribe to RSS](https://docs.opentermsarchive.org/terms/how-to/be-notified/) to be notified of changes to that specific file 🙂

LVerneyEC:
Quick update on this one because we just got another use case with a TXT file served by X: https://code.europa.eu/dsa/terms-and-conditions-database/vlops-and-vloses/vlop-vlose-declarations/-/merge_requests/82.

The current workaround is to have a dummy selector (e.g. `select: body` even though there are no body). The downside is that this loses all the newlines characters.

Ndpnt:
Hi @LVerneyEC,

Thanks for keeping this moving and for the additional use case. Quick recap on where things stand:

1. The terms type needs to be discussed in `terms-types` first. "LLM Documentation Index" is not a recognized terms type in https://github.com/OpenTermsArchive/terms-types. That discussion should happen there before the engine grows support to track such content, including the [principle 3](https://docs.opentermsarchive.org/concepts/design-principles/#3-obtain-documents-like-a-user-would) question. `llms.txt` is a convention aimed at LLM crawlers rather than end users, so the same concern as the initial Grok example applies. Could you open an issue on `terms-types`?

2. The corpus of five examples is still needed, per @MattiSG's earlier comment.

3. On the implementation side, a few points would need to be addressed before merging:
- The MIME-type check is too strict. `sourceDocument.mimeType == mime.getType('txt')` is strict equality against `'text/plain'`, but both `raw.githubusercontent.com` and `docs.x.com` return `text/plain; charset=utf-8`. The check fails and content falls back to `extractFromHTML`, reproducing the bug the PR aims to fix. The PDF branch in the code handles this via `mime.getExtension(...)`.
- The declaration schema is not updated (it still requires `select` for non-PDF, hence the lingering `select: body` workaround).
- Automated tests are still missing.

Happy to take a closer look at the PR once (1) and (2) are addressed.

PR #1176, opened 2025-07-18
Fetch multiple source documents sequentially to prevent bot detection

Ndpnt:
Tested with the Platform Governance Archive collection and it should fix https://github.com/OpenTermsArchive/pga-declarations/issues/296, https://github.com/OpenTermsArchive/pga-declarations/issues/260 and https://github.com/OpenTermsArchive/pga-declarations/issues/205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All changes that impact users of this module are documented in this file, in the
 
 ## Unreleased [minor]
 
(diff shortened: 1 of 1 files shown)


MattiSG (approved):
Yes, I agree, that's also what came to my mind: in large combines, there are just too many requests to do them concurrently.
I believe we could optimize this much further by:
1. Doing domain grouping (this only works for combines, if we track many terms for the same service won't we have the same issue across terms?)
2. Using a map-reduce rather than a for-await
3. Control the number of concurrent calls to an arbitrary amount (3? 5?) rather than going down to 1

But that will already likely be a great improvement as is, thanks!

PR #1178, opened 2025-07-23
Add support for parameterized filters in service declarations and add `removeQueryParams` built-in filter

Ndpnt:


src/archivist/services/index.test.js
```diff
@@ -1,14 +1,506 @@
...
+      const paramFilter = (dom, param) => param;
+      const serviceFilters = { paramFilter };
+      const result = getServiceFilters(serviceFilters, [{ paramFilter: 'foo' }]);
+
+      expect(result[0]).to.be.a('function');
+      expect(result[0].name).to.equal('paramFilter');
+      expect(result[0](null, 'context')).to.equal('foo');
+    });
+
+    context('parameters passed to filters', () => {
```
MattiSG, line 302:
```suggestion
    describe('parameters passed to filters', () => {
```

src/archivist/services/index.test.js
```diff
@@ -1,14 +1,506 @@
...
+          testParameterPassing({ param1: 'param1', param2: 'param2' });
+        });
+      });
+    });
+  });
+
+  describe('#createSourceDocuments', () => {
+    const realFilterNames = Object.keys(exposedFilters);
+    let result;
+    const serviceId = 'service_with_filters_history';
```
MattiSG, line 346:
```suggestion
    const SERVICE_ID = 'service_with_filters_history';
    let result;
```

src/archivist/services/index.test.js
```diff
@@ -1,14 +1,506 @@
...
+        combine: [
+          {
+            fetch: 'https://example.com/doc1',
+            select: '.content',
+            filter: ['removeShareButton'],
+          },
+          {
+            executeClientScripts: true,
+            remove: '.doc2-ads',
+          },
```
MattiSG, line 409: In [our doc](https://docs.opentermsarchive.org/terms/reference/declaration/), we say:

> Common properties (can be a combination of “select”, “remove”, “filter” and “executeClientScripts”) that are shared across all source documents can be factorized

…yet what I see here is a source document with no `fetch` property. Is that expected to be valid? If so, the doc needs to be corrected.

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
-  const rawServiceDeclaration = await fs.readFile(jsonDeclarationFilePath);
+export async function loadServiceDeclaration(serviceId) {
+  const filePath = path.join(declarationsPath, `${serviceId}${JSON_EXT}`);
 
   try {
-    return JSON.parse(rawServiceDeclaration);
+    return JSON.parse(await fs.readFile(filePath));
   } catch (error) {
-    throw new Error(`The "${serviceId}" service declaration is malformed and cannot be parsed in ${jsonDeclarationFilePath}`);
+    throw new Error(`The "${serviceId}" service declaration is malformed and cannot be parsed in ${filePath}`);
```
MattiSG, line 52: This is not true anymore: with the change above now including file read in the `try` block, there are other potential failure reasons than a malformed file.

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
-        location: sourceDocumentLocation || location,
-        executeClientScripts: (sourceDocumentExecuteClientScripts === undefined || sourceDocumentExecuteClientScripts === null ? executeClientScripts : sourceDocumentExecuteClientScripts),
-        contentSelectors: sourceDocumentContentSelectors || contentSelectors,
-        insignificantContentSelectors: sourceDocumentInsignificantContentSelectors || insignificantContentSelectors,
-        filters: sourceDocumentFilters || filters,
-      }));
-    }
+export function getServiceFilters(serviceFilters, filterNames) {
+  if (!filterNames) {
+    return undefined;
```
MattiSG, line 91:
```suggestion
    return;
```

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
   }
 
-  service.addTerms(new Terms({ service, type: termsType, sourceDocuments }));
+  const filters = filterNames.reduce((filters, filterItem) => {
```
MattiSG, line 94: Then `filterNames` is misnamed, if its entries are not filter names 😉

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
+    let filterFunction;
+
+    if (typeof filterItem === 'string') {
+      filterFunction = exposedFilters[filterItem] || serviceFilters[filterItem];
+    } else if (typeof filterItem === 'object' && filterItem !== null) {
+      const [ filterName, params ] = Object.entries(filterItem)[0];
+      const baseFunction = exposedFilters[filterName] || serviceFilters[filterName];
+
+      if (baseFunction) {
```
MattiSG, line 103: So we silently fail if the filter name has no match? Shouldn't we throw a validation error if the user tries to apply a filter that does not exist? Most likely it is a typo and we will save them time; if not, it is likely a misunderstanding of the usage of filters, such as copy-pasting another declaration without having the appropriate filter file, and we should inform them as well.

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
+      }
+    }
+    if (filterFunction) {
+      filters.push(filterFunction);
+    }
+
+    return filters;
+  }, []);
+
+  return filters.length ? filters : undefined;
```
MattiSG, line 117: Why don't we return an empty array, so that the value type is consistent?

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
-  Object.keys(filters).forEach(filterName => {
-    if (filterNames.includes(filterName)) {
-      filters[filterName].forEach(({ validUntil }) => allHistoryDates.push(validUntil));
-    }
-  });
+  await Promise.all(Object.entries(declarations).map(([ termsType, declarationEntries ]) => addTermsHistory(service, service.id, termsType, declarationEntries, filters)));
+}
+
+async function addTermsHistory(service, serviceId, termsType, declarationEntries, filters) {
+  const filterNames = [...new Set(declarationEntries.flatMap(d => d.filter))].filter(Boolean);
```
MattiSG, line 143: `d`? What is `d`?
```suggestion
  const filterNames = [...new Set(declarationEntries.flatMap(termsDeclaration => termsDeclaration.filter))].filter(Boolean);
```

src/archivist/services/index.js
```diff
@@ -4,261 +4,284 @@ import { pathToFileURL } from 'url';
...
+  return filtersHistory;
 }
 
 async function fileExists(filePath) {
   try {
     await fs.access(filePath);
 
     return true;
-  } catch (error) {
-    if (error.code === 'ENOENT') {
```
MattiSG, line 260: Why was this removed? It reads like a safety mechanism to avoid being too broad about silencing errors, and I could not find a rationale for broadening in 2e69c40ccf5866654f9a555635d4562f9ac3fc5b.

MattiSG (changes requested):
Wow, that's a big changeset! 😆 I think it will be useful to work on the doc before shipping the new version.

In order to minimize the number of comments in the review, I have consolidated hopefully non-controversial suggestions in a dedicated PR https://github.com/OpenTermsArchive/engine/pull/1181.
 
- Let's talk about the usage of `before`/`beforeEach`/`it` in tests 🙂 
- I discover that the engine uses `filter` as a terminology for handling the whole extraction process, including `remove` and… filters. This is very confusing. What are our plans for aligning the vocabulary better?
- As a related note, I like the idea that the standard treatment we apply systematically (converting relative URLs, removing scripts and styles…) could just be filters like any other. This would simplify the architecture. Let's talk about it.

src/archivist/extract/filter.js
```diff
@@ -0,0 +1,61 @@
+export const LINKS_TO_CONVERT_SELECTOR = 'a[href]:not([href^="#"]):not([href=""])';
+
+export default async function filter(webPageDOM, sourceDocument) {
+  await applyCustomFilters(webPageDOM, sourceDocument);
+  convertRelativeURLsToAbsolute(webPageDOM, sourceDocument.location);
+  removeUnwantedElements(webPageDOM);
+  updateProtectedLinks(webPageDOM);
+
+  webPageDOM.remove(sourceDocument.insignificantContentSelectors);
```
Ndpnt, line 9: Move back in `index.js`

MattiSG:
> should we change the signature of the filter function?

Following sync discussion with @Ndpnt, we decide to go with the suggested implementation, leading to a breaking change, but with no adaptation costs as no filter from the federated collections uses that second parameter.

In the future, our favorite implementation is to provide `document` and `declaration` through `this`, as it is part of what is provided by the engine; and the parameters that are passed are the ones in the declaration, so that the user aligns declaring and passing parameters;

src/archivist/services/index.test.js
Ndpnt: fetch property is factorized and here is `fetch: 'https://example.com/base',`

PR #1197, opened 2025-09-30
Reduce memory footprint when tracking terms with multiple source documents

Ndpnt:
On the VLOPSEs IE server, memory usage is reduced by a factor of 10 when processing combined terms from approximately 50 source documents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All changes that impact users of this module are documented in this file, in the
 ### Changed
 
 - Improve memory management when tracking terms with multiple source documents
+- All available snapshots are now stored in multi-documents terms, even if some are unavailable. 
 
 ## 9.0.0 - 2025-09-25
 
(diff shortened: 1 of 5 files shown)


src/archivist/services/sourceDocument.js
```diff
@@ -35,6 +35,11 @@ export default class SourceDocument {
     return result.filter(selector => selector);
   }
 
+  clearContent() {
+    this.content = null;
```
MattiSG, line 39: > You know better, but just sharing that it might be useful to set some sort of indicator in the object that content has been cleared, so that empty content is not confused with an issue in the content loading itself… ?

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [minor]
```
MattiSG, line 5: Why is this not just a patch?
```suggestion
## Unreleased [patch]
```

Ndpnt: See in sync, behavior changed.

PR #1199, opened 2025-10-07
Update supported Node versions

MattiSG:
Done in #1189, missed metadata update

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,9 @@ jobs:
           - ubuntu-22.04
           - windows-latest
           - macos-latest
+        node_version:
+          - 20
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - uses: ankane/setup-mongodb@v1
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
```
Ndpnt, line 27: I prefer using the oldest supported Node.js version in CI. It’s easy to verify compatibility with the latest version, but I never test with the oldest one locally. This approach ensures the version we claim to support actually works.

PR #1200, opened 2025-10-10
Fix: Prevent crash by skipping snapshots from deleted services

GundlaPreethika:
What this Pull request does:
- Fixes the issue where the Git repository crashes when loading snapshots from deleted services.
- Ensures that snapshots from deleted services are safely skipped.

Changes made:
- Updated  src/archivist/recorder/snapshot.js to handle deleted services.
- Added a new test `test/snapshots.test.js` to verify this behavior.

Why this is important:
Previously, loading snapshots from services that were deleted caused the application to crash. This fix improves stability and ensures smooth snapshot processing.

### DIFF
--- a/src/archivist/recorder/snapshot.js
+++ b/src/archivist/recorder/snapshot.js
@@ -1,5 +1,29 @@
-import Record from './record.js';
+import fs from 'fs/promises';
+import path from 'path';
 
-export default class Snapshot extends Record {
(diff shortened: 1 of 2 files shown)


Ndpnt:
Hi @GundlaPreethika,

Thanks for the contribution. Before diving into the code, could you clarify how you reproduced the crash? I’m not sure this issue exists in the current version of the codebase.

Looking at the implementation, the changes in `snapshot.js` don’t seem connected to how snapshots are actually loaded or validated, so it’s unlikely they address the described problem. Could you explain what led you to this solution and how you tested this?

PR #1242, opened 2026-04-29
Add feed endpoints

clementbiron:
This pull request adds Atom feed endpoints to the collection API so consumers can subscribe to recent terms updates.  A new `/feed` endpoint exposes a collection feed, with variants at `/feed/:serviceId` and `/feed/:serviceId/:termsType`.

~~Contrary to what was discussed in [#1238](https://github.com/OpenTermsArchive/engine/issues/1238), each entry's primary alternate link points to the GitHub commit rather than to the version API. While testing the feed in real readers, a link resolving to a JSON document felt unhelpful for human consumers, who expect to land on a readable page. The version API is still exposed alongside as a related link, so programmatic consumers can discover it from the feed.~~

**Update:** each entry now exposes a single `alternate` link. By default it points to the version API, which works for any storage backend. Operators who want feed readers to land on a human-readable page can configure `collection-api.feed.versionUrlTemplate` (e.g. `https://github.com/openTermsArchive/demo-versions/commit/%VERSION_ID`); the placeholder is interpolated with the version's identifier.


src/archivist/recorder/repositories/git/index.js
```diff
@@ -92,6 +92,26 @@ export default class GitRepository extends RepositoryInterface {
     return Promise.all((await this.#getCommits()).map(commit => this.#toDomain(commit, { deferContentLoading: true })));
   }
 
+  async findRecent(limit, { serviceId, termsType } = {}) {
```
Ndpnt: I think it is better and more idiomatic to not add this method and to update findAll with `limit` and `offset` params

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+
+import { getCollection } from '../../archivist/collection/index.js';
+import { COMMIT_MESSAGE_PREFIXES } from '../../archivist/recorder/repositories/git/dataMapper.js';
+import { toISODateWithoutMilliseconds } from '../../archivist/utils/date.js';
+
+import { findServiceCaseInsensitive } from './utils.js';
+import versionsRepository, { storageConfig } from './versionsRepository.js';
+
+const TAG_AUTHORITY = 'opentermsarchive.org,2026';
+const FEED_AUTHOR_NAME = 'OTA-Bot';
```
Ndpnt: Should use commit author

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+const DEFAULT_LIMIT = 100;
```
Ndpnt: Should be moved in the default config file

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+
+function getFeedLimit() {
+  if (config.has('@opentermsarchive/engine.collection-api.feed.limit')) {
```
Ndpnt: Avoid accessing config from this file directly. Follow dependency injection pattern instead.

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+  termsType: `tag:${TAG_AUTHORITY}:scheme:terms-type`,
+  recordType: `tag:${TAG_AUTHORITY}:scheme:record-type`,
+};
+
+function buildAbsoluteBaseUrl(req) {
+  return `${req.protocol}://${req.get('host')}${req.baseUrl}`;
+}
+
+function classifyRecordType(version) {
+  if (version.isFirstRecord) { return RECORD_TYPES.firstRecord; }
```
Ndpnt: I think I would have use a switch here. I find it more meaningful

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+  if (version.isTechnicalUpgrade) { return RECORD_TYPES.technicalUpgrade; }
+
+  return RECORD_TYPES.change;
+}
+
+function buildEntryTitle(version) {
+  let prefix = COMMIT_MESSAGE_PREFIXES.update;
+
+  if (version.isFirstRecord) { prefix = COMMIT_MESSAGE_PREFIXES.startTracking; } else if (version.isTechnicalUpgrade) { prefix = COMMIT_MESSAGE_PREFIXES.technicalUpgrade; }
```
Ndpnt: Multiple lines. Update Eslint if needed

src/collection-api/routes/feed.js
```diff
@@ -0,0 +1,262 @@
...
+   *               type: string
+   */
+  router.get('/feed', async (req, res) => {
+    const collection = await getCollection();
+    const baseUrl = buildAbsoluteBaseUrl(req);
+    const selfHref = `${baseUrl}/feed`;
+    const feedId = `tag:${TAG_AUTHORITY}:feed:${collection.metadata?.id}`;
+
+    const versions = await versionsRepository.findRecent(getFeedLimit());
+    const document = buildFeedDocument({ collection, selfHref, feedId, versions, baseUrl });
```
Ndpnt: As this is called in all exposed endpoints, factorize it. Maybe into `sendAtom` if it makes sense

src/collection-api/routes/feed.js
clementbiron: This refers to the author of the feed, not an entry within the feed ;)

Ndpnt (approved):

PR #1247, opened 2026-05-26
Restore stealth plugin and fix fetcher language configuration

clementbiron:
This PR reactivates `puppeteer-extra-plugin-stealth` in the full DOM fetcher. Since v10.3.1 (commit `5e4945d2`), `puppeteer.use(stealthPlugin(...))` was called inside `fetch()` instead of before `puppeteer.launch()`, so `puppeteer-extra` never bound its `onPageCreated` hooks and the plugin was silently inert. `navigator.webdriver` and `HeadlessChrome` were exposed to every tracked service for around four months.

It also applies the `@opentermsarchive/engine.fetcher.language` configuration to `navigator.language` and `navigator.languages` through the stealth sub-evasions, not just the `Accept-Language` HTTP header that was already affected. The stealth wrapper's `locale` option was being silently ignored. The redundant `setExtraHTTPHeaders` and CDP `Network.setUserAgentOverride` calls in `configurePage` are removed: they duplicated stealth's path with subtly different results.

**Breaking:** quality factors (`;q=...`) are no longer accepted in `@opentermsarchive/engine.fetcher.language`. Previously accepted values such as `en-IE,en-GB;q=0.9,en;q=0.8` now throw at `launchHeadlessBrowser`. Combined with the now-removed `setExtraHTTPHeaders` call, those values used to produce malformed headers like `q=0.9;q=0.9`, and the surface area was not worth preserving. Provide a plain comma-separated priority list instead (e.g. `en-IE,en-GB,en`); the browser derives the `Accept-Language` quality factors from tag order.

As a side effect of applying the configuration to `navigator.languages`, `language: "en"` now exposes `navigator.languages` as `["en"]` instead of the previous default `["en-US", "en"]`. Setting `language: "en-US,en"` restores the previous default.


src/archivist/fetcher/fullDomFetcher.js
```diff
@@ -1,13 +1,20 @@
+import config from 'config';
```
Ndpnt: Don't import the config module directly here, rely on dependency injection for the language parameter. All callers already pass it explicitly, so the fallback L.116 can go and the parameter becomes required.

src/archivist/fetcher/fullDomFetcher.test.js
```diff
@@ -118,4 +141,93 @@ describe('Full DOM Fetcher', function () {
...
+      });
+
+      it('sends every configured tag as Accept-Language header', async () => {
+        const result = await fetch(`http://127.0.0.1:${SERVER_PORT}/lang-header`, [], config);
+
+        expect(result.content).to.match(/data-accept-language="fr-FR,fr[^"]*"/);
+      });
+    });
+
+    context('with quality factors such as en-IE,en-GB;q=0.9,en;q=0.8', () => {
```
Ndpnt: The current implementation strips q-factors from the input and lets Chrome regenerate them from the order, so configured q-factors are silently overridden. Either preserve them end-to-end (override Accept-Language explicitly after the page is created), or drop q-factor support from the documented API.

I'd suggest the latter (option B): simpler, matches `navigator.languages` semantics, and rejects ambiguity at the boundary. Concretely:
- Update the JSDoc to describe the input as a comma-separated list of BCP 47 tags in priority order, with no q-factors, and drop the `en-IE,en-GB;q=0.9,en;q=0.8` example.
- Have `parseLanguage` reject inputs containing `;q=` rather than silently stripping them, so existing configurations don't keep "working" with different semantics.
- Rename the test to reflect that Chrome derives the q-factors from the order, instead of claiming they're preserved.

If a real need for explicit q-factors emerges later, the other option remains addable in a backward-compatible way.

src/archivist/fetcher/fullDomFetcher.test.js
```diff
@@ -118,4 +141,108 @@ describe('Full DOM Fetcher', function () {
...
+        expect(result.content).to.match(/data-language="fr-FR"/);
+      });
+
+      it('exposes every tag through navigator.languages', async () => {
+        const result = await fetch(`http://127.0.0.1:${SERVER_PORT}/lang-echo`, [], config);
+
+        expect(result.content).to.match(/data-languages="fr-FR,fr"/);
+      });
+
+      it('drives client-side language detection on the fetched document', async () => {
```
Ndpnt: This test does not had value. It covers the same as the previous ones.

src/archivist/fetcher/fullDomFetcher.test.js
```diff
@@ -118,4 +141,108 @@ describe('Full DOM Fetcher', function () {
...
+      it('rejects language values containing quality factors', async () => {
+        await stopHeadlessBrowser();
+        await expect(launchHeadlessBrowser('en-IE,en-GB;q=0.9,en;q=0.8'))
+          .to.be.rejectedWith('Quality factors are not supported');
+      });
+    });
+  });
+
+  describe('Stealth evasions', () => {
+    // These assertions guard against a class of regressions where the
```
Ndpnt: Oneline

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 All changes that impact users of this module are documented in this file, in the [Common Changelog](https://common-changelog.org) format with some additional specifications defined in the CONTRIBUTING file. This codebase adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased [major]
+
+> Development of this release was supported by [User Rights](https://www.user-rights.org).
+
+### Changed
+
+- **Breaking:** Apply `@opentermsarchive/engine.fetcher.language` to `navigator.language` and `navigator.languages` in addition to the `Accept-Language` header; `language: "en"` now exposes `navigator.languages` as `["en"]` instead of `["en-US", "en"]`, set `language: "en-US,en"` to restore the previous default
```
Ndpnt: Small factual correction for the PR description: in the post-5e4945d2 state, `navigator.language` and `navigator.languages` were already affected by the `Network.setUserAgentOverride.acceptLanguage` call in `configurePage` (CDP derives both from `acceptLanguage`, not just the header). The real change is that the new path uses the stealth `navigator.languages` evasion's JS-injected getter, which exposes the configured list strictly without Chrome's regional-variant expansion. That's where the `["en"] vs ["en-US", "en"]` divergence comes from.

Ndpnt: So I suggest:

```suggestion
- **Breaking:** `@opentermsarchive/engine.fetcher.language` now exposes `navigator.languages` strictly as configured, without Chrome's automatic regional-variant expansion. `language: "en"` exposes `navigator.languages` as `["en"]` instead of the previous `["en-US", "en"]` derived by Chrome; set `language: "en-US,en"` to restore the previous value
```

Ndpnt (approved):

PR #1250, opened 2026-06-21
Reject empty documents object in service schema

dharma3306:
## Summary

This change rejects service declarations containing an empty `documents` object.

## Changes

- Added validation to prevent empty documents objects.
- Added an invalid fixture: `empty-documents.json`.
- Added a schema validation test to ensure empty documents objects are rejected.

## Testing

- Ran schema validation tests successfully.
- All tests passed.

### DIFF
--- a/scripts/declarations/validate/fixtures/invalid/empty-documents.json
+++ b/scripts/declarations/validate/fixtures/invalid/empty-documents.json
@@ -1,0 +1,4 @@
+{
+    "name": "Example",
+    "documents": {}
+}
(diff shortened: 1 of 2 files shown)


MattiSG:
This is clearly LLM-generated, messes up indentation, adds an irrelevant test, uses a file format deprecated since v4 of the engine. We don't invest time in supporting LLM-based, low-effort pull requests that cannot qualify as showing an interest from the contributor in actually contributing as opposed to farming reputation. Closing. Reopening a PR showing a similar low level of investment will result in banning.

Issue #1105, opened 2024-09-24
Fetch repeated pages only once


MattiSG:
If several terms come from the same `fetch` origin, the engine should cache the source document.

Issue #166, opened 2020-10-09
Bypassing bot detectors?


LucasVerneyDGE:
Hi,

Rakuten and Leboncoin have very strong bot detectors, hence preventing from automatically fetching their CGUs (at least on a regular OVH machine). See https://fr.shopping.rakuten.com/newhelp/conditions-generales/ or https://www.leboncoin.fr/dc/cgu. It is possible that #138 and having JS enabled will help here, but I think this won't be enough.

Best,

Ndpnt:
Hi,

I hope using a headless browser will fix this. So I suggest to wait for #138 to be implemented and see if there is still this issue. 
Unless you have a quicker to implement idea to fix it?

LucasVerneyDGE:
Using a headless browser is not enough to fix this. You have to disguise it (https://github.com/berstend/puppeteer-extra/tree/master/packages/puppeteer-extra-plugin-stealth for instance) and you are still identified by your IP address (DataDome used on Leboncoin for instance does this), if you are connecting from a server infrastructure (not residential).

MattiSG:
As discussed with @LucasVerneyDGE and @TomHouriezDGE, this option will be needed for some sources, even after #138 is fixed. However, it also raises legal questions. @LucasVerneyDGE will investigate which entities might have power to legally bypass access control systems, and we will design the most appropriate software architecture (opt-in, opt-out, plugin) based on the legal assessment 🙂

MattiSG:
Legal analysis by PEReN was still pending on 08/03/2022.

Imperva and Cloudflare answers are still pending.

In order to help with prioritisation, instead of listing issues in this repository, they are now labeled in each affected instance [with dedicated tags](https://github.com/OpenTermsArchive/france-declarations/issues?q=is%3Aopen+is%3Aissue+label%3A403) (403, timeout…).

MattiSG:
@LVerneyPEReN did the PEReN finish its legal analysis? 🙂 

On our side, I believe we never got a reply from Imperva nor Cloudflare (please correct me if I'm wrong @martinratinaud).

martinratinaud:
> @LVerneyPEReN did the PEReN finish its legal analysis? 🙂
> 
> On our side, I believe we never got a reply from Imperva nor Cloudflare (please correct me if I'm wrong @martinratinaud).

Indeed, we did not 😔

MattiSG:
[...]

PR #1253, opened 2026-06-23
Randomize tracking

LVerneyEC:
Hey,

Here is a small proposal for an addition to shuffle the tasks to be processed for scraping/tracking terms in order to increase stealth and not always hit the same service's pages in the exact same order.

Otherwise, the exact same pattern will happen again and again when scraping a service's terms. This means:
1/ Hitting rate limiting because we are sending a batch (up to MAX_CONCURRENCY) calls more or less the same time to the service.
2/ Easier detection of automated scraping, even more when coupled with a cron-based scheduler, as order of scraping is almost entirely predictible. 

Randomizing the order of tracking of the terms is a cheap solution for this. The only downside might be a slight decrease in readability when ran interactively, but with concurrency enabled this was already the case. If the feature is worth merging, happy to have a better integration with config option or so.

Best,

### DIFF
--- a/src/archivist/index.js
+++ b/src/archivist/index.js
@@ -156,6 +156,7 @@ export default class Archivist extends events.EventEmitter {
     this.trackingQueue.concurrency = concurrency;
 
     const tasksToQueue = [];
+
     servicesIds.forEach(serviceId => {
       this.services[serviceId].getTermsTypes().forEach(termsType => {
         if (termsTypes.length && !termsTypes.includes(termsType)) {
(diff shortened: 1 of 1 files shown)


MattiSG:
Thank you @LVerneyEC for your ongoing work on #166!

We have considered randomising the tracking order, however this raises very strong concerns with regards to debugging. Indeed, true randomisation means that there is no way to reproduce a failed tracking run. As such, we will not merge such a proposal without a reproducible benchmark demonstrating tangible benefits.

**We would instead prefer to introduce an algorithmic**, deterministic sorting system that manipulates the queue to maximise the space between fetches to the same server. In this way, the same set of declarations would always yield the same fetching order. It would also guarantee that the space is maximised, while randomisation will also yield groupings with high rates of requests to the same server.
In the same vein, you might also want to look into implementing #1105 as a way to decrease bot detection.

Admittedly, this would only solve your point 1 (rate limit). While point 2 (deterministic order) is theoretically possible, we are skeptical of its actual role in triggering bot detection.

One way for us to reconsider the cost-to-benefit ratio (as in, “cost to reproducibility and maintainability to benefit to tracking”) would be to **provide reproducible data on tracking success** comparing the current naive implementation to a space-maximising algorithm to randomisation.

I will close this PR now, as it will not be considered without such a benchmark. We will happily reopen it if such data is provided and is conclusive, as we would love to increase the tracking success rate!

PR #1254, opened 2026-06-23
Validate unique combine fetch

clementbiron:
A combine can currently repeat the same fetch URL within a terms type, which causes redundant fetches. `npm run declarations:validate` now rejects it and reports the duplicated URL(s). The same fetch stays allowed across different terms types.


Ndpnt (approved):

MattiSG:
This change removes the ability to insert a subset of page A, then page B, then another subset of page A.
There is no clear use case for this.
There is however a clear cost, and a recurring problem, with duplicating documents in a combine.
This is worthwhile; if in the future we find a clear use case for inserting pages in-between one page, we can reconsider this decision.

PR #1259, opened 2026-06-30
Ci speedup

MattiSG:
Depends on #1258.

### Speedup and failure rate decrease by stopping to run integration tests cross-platform

This changeset stops running metadata and declaration validation on macOS and Windows. This is based on the idea that engine tests definitely should be cross-platform, but smoke-testing downstream dependencies does not need to be cross-platform: cross-platform behavioural consistency is guaranteed by the engine tests, integration testing is not necessary cross-platform. If this hypothesis does not hold in production and we get defect reports, we can revert this changeset.

**Expected speedup: at least 1 minute per run (18% speedup), up to 2m30 (40%) per run.**

Indeed, Windows runs around ~1m15 for both metadata and declarations validation, while macOS and Ubuntu run in around ~42s.
With perfect parallelism, that's 2 x 30s saved; with totally failing parallelism, that's 2 x 75s saved. The reality will probably be somewhere in between.

**Preliminary run result shows no evident speed gain** (5m17s, the lowest in the range of previous runs but still in the same range).

### Test cross-platform on Node latest instead of oldest

For some reason, we were testing on macOS and Windows only on the oldest supported version of Node instead of the most recent. This changeset inverts that behaviour.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         operating_system: 
-          - ubuntu-22.04
+          - ubuntu-22.04 # Using Ubuntu 22.04 due to AppArmor restrictions in newer versions (23.10+) that interfere with Puppeteer's sandbox functionality. See https://pptr.dev/troubleshooting#issues-with-apparmor-on-ubuntu
           - windows-latest
           - macos-latest
         node_version:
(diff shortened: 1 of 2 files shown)


Ndpnt (approved):

MattiSG:
I'm going to update the required checks to allow merging this PR. This means that existing open PRs will need to be rebased before being merged 🙂 
cc @OpenTermsArchive/core

PR #1260, opened 2026-07-02
Add normalize whitespace built-in filter

### LINKED ISSUE #147: Harmonize whitespaces
https://github.com/ambanum/CGUs-versions/commit/3af7bf439 results in a notification when the only modifications happening are changes between non-breakable spaces and regular ones.

clementbiron:
Some documents use Unicode space separators (non-breaking space U+00A0, narrow no-break space U+202F, ideographic space U+3000, ...) instead of regular spaces. These characters are visually identical to a normal space but differ byte-for-byte, which produces blink when tracking terms.

This changeset adds a `normalizeWhitespace` built-in filter that replaces every Unicode space separator (`\p{Zs}`) with a regular space in the text nodes of the page.

### DIFF
--- a/src/archivist/extract/exposedFilters.js
+++ b/src/archivist/extract/exposedFilters.js
@@ -26,3 +26,18 @@ export function removeQueryParams(webPageDOM, paramsToRemove = []) {
     }
   }
 }
+
+const SPACE_SEPARATORS = /\p{Zs}/gu;
+
+export function normalizeWhitespace(webPageDOM) {
+  const walker = webPageDOM.createTreeWalker(webPageDOM.body, webPageDOM.defaultView.NodeFilter.SHOW_TEXT);
+
+  for (let node = walker.nextNode(); node; node = walker.nextNode()) {
+    const original = node.nodeValue;
+    const normalized = original.replace(SPACE_SEPARATORS, ' ');
+
+    if (normalized !== original) {
+      node.nodeValue = normalized;
+    }
+  }
+}
--- a/src/archivist/extract/exposedFilters.js
+++ b/src/archivist/extract/exposedFilters.js
@@ -29,7 +29,7 @@ export function removeQueryParams(webPageDOM, paramsToRemove = []) {
 
 const SPACE_SEPARATORS = /\p{Zs}/gu;
 
-export function normalizeWhitespace(webPageDOM) {
+export function convertSpacesToStandard(webPageDOM) {
   const walker = webPageDOM.createTreeWalker(webPageDOM.body, webPageDOM.defaultView.NodeFilter.SHOW_TEXT);
 
   for (let node = walker.nextNode(); node; node = walker.nextNode()) {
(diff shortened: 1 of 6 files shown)


src/archivist/extract/exposedFilters.test.js
```diff
@@ -265,4 +265,46 @@ fetch(trackingUrl);
...
+      expect(element.textContent).to.equal('first partsecond part');
+    });
+
+    it('does not alter element attributes or URLs, only text content', () => {
+      const href = 'https://example.com/page?label=gen173__nr-1';
+
+      element = webPageDOM.createElement('a');
+      element.setAttribute('href', href);
+      element.textContent = 'read\u00A0the\u00A0policy';
+      webPageDOM.body.appendChild(element);
```
MattiSG: This should be in a `before` block.

src/archivist/extract/exposedFilters.test.js
```diff
@@ -265,4 +265,46 @@ fetch(trackingUrl);
...
+      element.innerHTML = '<span>first\u00A0part</span><span>second\u00A0part</span>';
+      webPageDOM.body.appendChild(element);
+
+      normalizeWhitespace(webPageDOM);
+
+      expect(element.textContent).to.equal('first partsecond part');
+    });
+
+    it('does not alter element attributes or URLs, only text content', () => {
+      const href = 'https://example.com/page?label=gen173__nr-1';
```
MattiSG: How is that content testing that whitespace in an attribute is not edited? There is no nbsp-equivalent in this bit.
I doubt that `href` is a good attribute as it should be URL-encoded. Maybe use `class` or `title` instead?

src/archivist/extract/exposedFilters.test.js
```diff
@@ -265,4 +265,46 @@ fetch(trackingUrl);
...
+      element = webPageDOM.createElement('p');
+      element.textContent = 'a\u00A0b\u202Fc\u2009d\u3000e';
+      webPageDOM.body.appendChild(element);
+
+      normalizeWhitespace(webPageDOM);
+
+      expect(element.textContent).to.equal('a b c d e');
+    });
+
+    it('normalizes text across nested elements', () => {
```
MattiSG: I don't understand the added value of this test.

src/archivist/extract/exposedFilters.js
```diff
@@ -26,3 +26,18 @@ export function removeQueryParams(webPageDOM, paramsToRemove = []) {
     }
   }
 }
+
+const SPACE_SEPARATORS = /\p{Zs}/gu;
+
+export function normalizeWhitespace(webPageDOM) {
```
MattiSG: The function name is a bit too broad from my perspective: it does not normalise _whitespace_, which would also include newlines and tabs (a recurring source of noise when multiple newlines are inserted, for example), but only spaces.

I also recommend to use British English spelling:

```suggestion
export function normaliseSpaces(webPageDOM) {
```

I wonder the following version might not be more understandable by non-technical users: 

```suggestion
export function convertSpacesToStandard(webPageDOM) {
```

MattiSG (changes requested):
Exciting to see built-in filters expanding!

MattiSG (approved):

PR #1266, opened 2026-09-06
Run duplicate checks during schema-only validation

### LINKED ISSUE #1265: Run the duplicate source documents check in schema-only validation
The check `does not declare the same source document more than once` ([`index.mocha.js#L121-L127`](https://github.com/OpenTermsArchive/engine/blob/0669bc3994b45a57bc6925fd1dbb4b5eb6b0e035/scripts/declarations/validate/index.mocha.js#L121-L127)) is gated behind `if (!schemaOnly && service)`. It therefore only runs in the full validation, which collections CI restricts to the terms types modified by a pull request. Duplicates predating v14.1.0, or living in terms types that are not modified again, are never reported.

In `user-rights-declarations`, TikTok "Advertising Content Policy" has declared the same URL twice since 2026-05-19 (removed in https://github.com/OpenTermsArchive/user-rights-declarations/pull/62), and Facebook and Instagram "Advertising Content Policy" still do on `main`. `ota validate declarations --schema-only` passes on all three.

Since the check needs neither network nor browser, move it out of the `!schemaOnly` branch so it runs on every declaration, keeping only the source documents loop gated.

be-student:
## Summary

- run duplicate source-document checks for every current terms type during schema-only validation
- keep browser and network source-document checks behind full validation
- add a CLI regression fixture and document the patch-level change

Closes #1265

## Validation

- `npm run test:only -- scripts/declarations/validate/index.test.js scripts/declarations/validate/schemas.test.js` (20 passing)
- `NODE_ENV=test npm run declarations:validate:schema -- --services service_with_multiple_source_documents_terms` (4 passing, including duplicate validation)
- `npx eslint scripts/declarations/validate/index.test.js scripts/declarations/validate/index.mocha.js`
- `git diff --check`

The full `npm test` run reached 1,250 passing tests. Three Mongo setup hooks timed out because no local Mongo service was available, and the existing dataset export test found a generated README mismatch. The new CLI regression initially hit Mocha's two-second suite timeout and now has an explicit ten-second limit; it passes in 0.7-1.1 seconds.


Ndpnt:
Thanks @be-student for taking this on!

I pushed a few commits on top of yours, here is why:

- I kept the duplicate check in the existing terms types loop. The second loop bypassed the `--types` and `--modified` options, which collections rely on in their CI, and duplicated the filter on current terms. Skipping only the source document checks in schema-only mode fixes the gap without changing how the options apply.
- I also removed CLI test. The rule is already covered by the `Terms` unit tests, and the CI job validating `demo-declarations` exercises the schema-only path end to end. Spawning the CLI from the test suite would set a new pattern in this repository, which deserves its own discussion rather than a patch release.

Thanks again, this closes a gap that was silently affecting real collections.

Ndpnt (approved):

PR #1267, opened 2026-09-06
Remove tracking query parameters by default

### LINKED ISSUE #1263: Remove tracking query parameters by default
Provide default configuration for the built-in `removeQueryParameters` filter to remove the well-known [tracking query parameters](https://privacytests.org).

be-student:
## Summary

- use the 23 current PrivacyTests tracking query parameters when `removeQueryParams` is enabled without configuration
- preserve explicit string, array, and empty-array configurations
- cover both direct filter behavior and the runtime filter-context path

Closes #1263

The default list is pinned to [`privacytests/privacytests@dda473a`](https://github.com/privacytests/privacytests/blob/dda473a462e5c37f9f2c8b2367fbc5f834796be4/live/results.js#L22-L68), so its provenance and update boundary are explicit.

## Validation

- Baseline: the new default-list regression failed because all 23 tracking parameters remained in the URL
- `npm run test:only -- src/archivist/fetcher/index.test.js src/archivist/extract/filter.test.js src/archivist/extract/exposedFilters.test.js` (65 passing)
- `npx eslint src/archivist/extract/exposedFilters.js src/archivist/extract/exposedFilters.test.js src/archivist/extract/filter.test.js`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All changes that impact users of this module are documented in this file, in the
 
 ### Changed
 
-- Remove well-known tracking query parameters (`utm_*`, `fbclid`, `gclid`, ...) by default when the built-in `removeQueryParams` filter is declared without parameters; see the full list in the [built-in filters documentation](https://docs.opentermsarchive.org/terms/reference/built-in-filters/)
+- Remove well-known tracking query parameters (`utm_*`, `fbclid`, `gclid`, ...) by default when the built-in `removeQueryParams` filter is declared without parameters; see the [built-in filters documentation](https://docs.opentermsarchive.org/terms/reference/built-in-filters/)
 
 ## 15.1.0 - 2026-07-13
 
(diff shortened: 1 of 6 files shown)


Ndpnt:
Thank you @be-student for this contribution.

I pushed two adjustments on top of your commit:
- The loader now wraps built-in filters even when they are declared without parameters, so they always receive `(webPageDOM, params, context)`. This lets `removeQueryParams` keep a plain default parameter instead of guessing from the type of its second argument, and future built-in filters with parameters get the same guarantee for free.

- The default list now also includes the five `utm_*` campaign parameters. PrivacyTests inherits Brave's list, which deliberately excludes them because they don't identify individual users. Our goal is different as we want stable versions of documents, and `utm_*` is a common source of noise in terms links.

I also bumped the release to `[minor]` since declarations that already list `removeQueryParams` without parameters will start producing new versions.

Ndpnt:
Related PR for documentation https://github.com/OpenTermsArchive/docs/pull/223

Ndpnt (approved):

MattiSG:
Thank you @be-student for these contributions and thank you @Ndpnt for handling and finalising them!

@be-student we'll be happy to mention you as a contributor to the project. Please make sure to review and add details such as name as you'd like in the associated PR 🙂 → @allcontributors add @be-student as a contributor for code.

If you'd like to contribute further to Open Terms Archive, we'd love to have you join our [next community call](https://docs.opentermsarchive.org/community/reference/community-call/) and our [chat](https://docs.opentermsarchive.org/community/how-to/join/)!

Issue #1262, opened 2026-07-08
Self-hosted dataset exposure through the collection API


clementbiron:
## Context

The engine periodically generates a dataset of a collection: a ZIP archive of all versions of all tracked terms. In production it runs on a weekly schedule (dataset.publishingSchedule) that generates and publishes the archive in one step.

Distribution is push-based: depending on the credentials configured on the tracking machine, `ota dataset --publish` uploads the archive to a Git forge release (GitHub or GitLab) and to data.gouv.fr. These are the dataset's **sole distribution channel**: the local archive is deleted after upload.

## Problem statement

A collection's dataset cannot be retrieved from the instance that produces it; its distribution depends entirely on external platforms. This weakens the decentralised architecture: each instance is the authoritative source of its versions and snapshots, but not of its own dataset, which exists only as a copy hosted by a third party. And with no platform credentials configured, the dataset cannot be distributed at all.

This RFC proposes to reverse the flow, from push to pull: each instance stores its latest dataset locally and exposes it, with its metadata, through the existing collection API, becoming the authoritative source of its own dataset.

The existing publications to GitHub, GitLab and data.gouv.fr are kept for backward compatibility and catalogue visibility, but become downstream consumers of this API.

Out of scope:

- Dataset history: only the latest dataset is exposed. History remains available through the third-party releases for collections that publish there.
- On-the-fly generation: the API serves the latest pre-generated archive and never triggers a generation itself.

### Local storage

The generated dataset is kept on the instance in a configurable directory (`dataset.storagePath`, defaulting to `./data/datasets`, alongside the existing `./data/versions` and `./data/snapshots`). The directory contains the archive under its dated name and a `metadata.json` sidecar file written at generation time:

```
data/datasets/
├── demo-2026-07-06.zip
└── metadata.json
```

```json
{
  "filename": "demo-2026-07-06.zip",
  "title": "demo",
  "license": "ODbL-1.0",
  "releaseDate": "2026-07-06T08:30:12Z",
  "firstVersionDate": "2022-01-01T12:00:00Z",
  "lastVersionDate": "2026-07-05T23:12:00Z",
  "servicesCount": 262,
  "termsCount": 641,
  "versionsCount": 38294,
  "size": 104857600,
  "sha256": "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
}
```

| Field | Type | Description |
|---|---|---|
[...]

PR #1270, opened 2026-09-09
Expose dataset through the Collection API

clementbiron:
Implements the [RFC #1262](https://github.com/OpenTermsArchive/engine/issues/1262): each instance stores its latest dataset locally and exposes it through the Collection API, becoming the authoritative source of its own dataset instead of relying solely on third-party publications.

- `ota dataset` writes the archive into the directory defined by the new `dataset.storagePath` configuration option, atomically, alongside a `metadata.json` sidecar describing it (title, license, dates, counts, size, checksum).
- `GET /dataset/latest` returns the metadata with an absolute `downloadURL`; `GET /dataset/latest/download` streams the archive with HTTP caching (`ETag`/`Last-Modified`, conditional requests) and resumable downloads (`Range`).
- The local copy is now the reference dataset served by the API, so `--remove-local-copy` is removed and the archive is kept after publishing; existing publications to GitHub, GitLab and data.gouv.fr are unaffected and become downstream consumers of this local store.


config/default.json
```diff
@@ -46,7 +46,8 @@
       "timestampPrefix": true
     },
     "dataset": {
-      "publishingSchedule": "30 8 * * MON"
+      "publishingSchedule": "30 8 * * MON",
+      "storagePath": "./data/datasets"
```
Ndpnt: 👍

Ndpnt (approved):

Issue #1241, opened 2026-04-15
Tracking results storage and exposure


Ndpnt:
# Tracking results storage and exposure

## Context

The engine periodically tracks whether online documents are accessible and extractable. For each service/terms pair, the outcome is either success (content fetched and version recorded) or failure (content inaccessible or extraction broken). When tracking fails, the `Reporter` module creates an issue on a third-party software forge (GitHub or GitLab). When tracking resumes, it closes the issue. These forge issues are the **sole persistent record** of tracking results.

## Problem statement

Tracking results have no local persistence. Even when everything works, the persistence of tracking data depends entirely on an external forge. If the forge is unreachable or no API token is configured, failures are logged but not persisted, and the information is lost once the process ends. There is no way to query the current tracking status of a service without calling an external forge API.

This RFC proposes to introduce a local storage layer that persists tracking results independently of any external service, and to expose them through the collection API. This storage becomes the single source of truth for tracking status. External integrations (such as the current GitHub and GitLab issue managers) would then be extracted into separate modules that consume tracking results through this API. The active consumption of this data (creating issues, sending notifications) is out of scope.

## Alternatives considered

Two existing standards and four alternative storage strategies were evaluated before arriving at the proposed design. None was retained. A dedicated Git repository with a custom JSON format was chosen instead.

<details>
<summary>Existing standards considered</summary>

### TAP — Test Anything Protocol

[TAP](https://testanything.org/) is a text-based protocol for reporting test results. The analogy with tracking is genuine: for each service/terms pair, the engine verifies that content is accessible and extractable, producing a pass or fail result.

TAP is not retained for two reasons. First, the standard only structures a small fraction of the data: TAP provides a binary `ok`/`not ok` outcome, but the engine needs to persist richer information (error reasons, source document metadata, snapshot references, timing data, transient errors). All of this must go into free-form YAML diagnostic blocks, outside the standard's structure:

```tap
TAP version 14
1..1
not ok 1 - Facebook · Terms of Service
  ---
  date: "2026-03-15T10:30:00Z"
  runId: "f47ac10b-58cc-4372-a567-0e02b2c3d479"
  reasons:
    - "CSS selector \".content\" has no match in the document"
  sourceDocuments:
    - id: legal-terms
      fetch: "https://www.facebook.com/legal/terms"
      select: ".content"
      snapshotId: abc123
      mimeType: text/html
  ...
```

[...]

PR #1275, opened 2026-09-23
Record tracking results in a dedicated Git repository

Ndpnt:
Implements the first part of [RFC #1241](https://github.com/OpenTermsArchive/engine/issues/1241#issuecomment-4499539428): a `tracking-results` Git repository recording, for each terms, its latest tracking status, and for each run, its lifecycle (start, per-terms outcomes, completion). Its history is tamper-evident, so it shows which terms were actually tracked, when, and why they failed.


clementbiron:
I made these changes after our pairing session with @Ndpnt:

- `b7f7f5e1` Clarify the skip reason of crash recovery: terms without a committed outcome are skipped as `no outcome recorded before the engine crashed` instead of `engine crashed`, since unchanged terms leave no trace and may well have been processed.
- `58ca9987` Omit snapshot links for unrecorded documents: tracking issues no longer link a nonexistent `.null` file for source documents that were never recorded as snapshots.
- `e62b7b8a` Unquote file names listed from the git log: crash recovery no longer loses terms whose service ID contains a double quote, which git C-quotes in its log output.
- `568c4125` Reuse the previous content read on save: the content read to detect changes doubles as the rollback backup, sparing a second read per recorded outcome.
- `4887ceeb` Point operators to the repository on failed setup: the recurring warning now names the repository path to inspect or delete.

clementbiron (approved):

PR #1277, opened 2026-09-23
Prevent concurrent writers on the same Git repository

Ndpnt:
Two writers on the same Git repository, typically a second `ota track` started by hand while the scheduled tracker runs, race the working tree and the index: one fails, or its clean up discards the uncommitted changes of the other. #1273 made the readers safe, this handles the writers.

A writer now takes a lock file in `.git/ota-writer.lock`, holding its PID, and keeps it until the process exits. A second writer refuses to start with an error naming the holder. A lock left by a killed process is detected as stale and taken over atomically.

The `index.lock` left behind by a killed Git operation is also removed at each run, as it made the tracker fail at every restart.


clementbiron (approved):

clementbiron:
Seen in sync with @Ndpnt

